PR #4, opened 2020-08-29
Permutation

therealyingtong:
Implement copy constraints between advice wires in constraint system.


src/plonk/srs.rs
```diff
@@ -42,31 +43,155 @@ impl<C: CurveAffine> SRS<C> {
...
+        // advice wires involved.
+        let mut largest_permutation_length = 0;
+        for permutation in &meta.permutations {
+            largest_permutation_length =
+                std::cmp::max(permutation.len(), largest_permutation_length);
+        }
+
+        // The permutation argument will serve alongside the gates, so must be
+        // accounted for.
+        let mut degree = largest_permutation_length;
```
ebfull, line 97: This needs to be `std::cmp::max(largest_permutation_length + 1, 1)`

src/plonk/srs.rs
```diff
@@ -42,31 +43,155 @@ impl<C: CurveAffine> SRS<C> {
...
+                let right: (usize, usize) = *self.copy[permutation]
+                    .get_mut(right_wire)
+                    .and_then(|wire| wire.get_mut(right_row))
+                    .ok_or(Error::BoundsFailure)?;
+
+                if left == (left_wire, left_row) || right == (right_wire, right_row) {
+                    // Don't perform the copy constraint because it will undo
+                    // the effect of the permutation.
+                } else {
+                    *self.copy[permutation]
```
ebfull, line 69: We can just index into these vectors at this point since we've bounds checked these indicies previously.

src/plonk/verifier.rs
```diff
@@ -63,17 +78,62 @@ impl<C: CurveAffine> Proof<C> {
...
+            for (advice_eval, permutation_eval) in queries
+                .iter()
+                .map(|&query_index| self.advice_evals[query_index])
+                .zip(self.permutation_evals[permutation_index].iter())
+            {
+                left *= &(advice_eval + &(x_0 * permutation_eval) + &x_1);
+            }
+
+            let mut right = self.permutation_product_inv_evals[permutation_index];
+            let mut current_delta = x_0;
```
therealyingtong, line 115:
```suggestion
            let mut current_delta = x_0 * &x_3;
```

src/plonk/prover.rs
```diff
@@ -98,13 +120,127 @@ impl<C: CurveAffine> Proof<C> {
...
+                    *tmp_advice_value += &x_1;
+                }
+                modified_advice.push(tmp_advice_values);
+            }
+
+            // Batch invert to obtain the denominators for the permutation product
+            // polynomial
+            for v in &mut modified_advice {
+                C::Scalar::batch_invert(v);
+            }
```
daira, line 169: Can we batch invert all of the `v`s together?

daira (approved):
utACK

src/plonk/prover.rs
ebfull: Yes, there are a few places scattered around the code where such inversions can be folded together. We could try to write a generic Montgomery's trick API so that we can write

```rust
modified_advice.iter_mut().flat_map(|v| v.iter_mut()).batch_invert();
```

or something like that.

src/plonk/prover.rs
```diff
@@ -53,11 +53,32 @@ impl<C: CurveAffine> Proof<C> {
...
+                Ok(())
+            }
         }
 
         let mut meta = MetaCircuit::default();
         let config = ConcreteCircuit::configure(&mut meta);
 
+        // Get the largest permutation argument length in terms of the number of
+        // advice wires involved.
+        let mut largest_permutation_length = 0;
```
ebfull, line 76: We can get rid of this as we don't actually need it in the prover.

src/plonk/prover.rs
therealyingtong: `modified_advice` is a vector of vectors, so i think we have to break it up into vectors to pass into `batch_invert()`

ebfull: Done!

src/plonk/prover.rs
ebfull: Done.

PR #5, opened 2020-09-01
Fix typos

rex4539:


ebfull:
Hi! Thanks for the contribution.

Could you please sign the [contributor agreement](https://github.com/zcash/halo2/blob/main/Contributor_Agreement) and send it to us (sean@electriccoin.co works)? Signing the agreement this allows us to relicense the code later if we need to, without having to get your permission. This is similar to how contributor agreements work in Apache software projects as well.

ebfull:
Thanks!

PR #13, opened 2020-09-09
Support batching and accumulation in polynomial opening argument

### LINKED ISSUE #10: Support batching and accumulation in polynomial opening argument
Currently, the verifier performs a complete verification of the `OpeningProof`. We want to support both accumulation and batching in the opening argument, and here's a sketch of what that might look like:

We'll have an `MSM` (multiscalar multiplication) structure parallel to `Params`. This structure will let you add arbitrary terms (the scalar and the point) or add to terms that already exist (such as `g` and `h`). It'll also let you scale the entire thing by a random blinding factor. Finally, it'll let you check `is_zero()` to perform the multiexp and check that it results in zero.

The `OpeningProof::verify()` method will take an `MSM` as input for it to append the necessary values. It will _not_ compute the `G` element (expensive computation) and instead return (1) a structure containing the challenges, and (2) a guard structure which lets the prover either supply the challenges and obtain an MSM, or supply the purported G value and obtain an MSM. The caller will either keep the challenges and the G value and place it in an accumulator object, or it will discard the challenges and check the MSM either immediately or later on.

therealyingtong:
Closes #10

### DIFF
--- a/src/arithmetic.rs
+++ b/src/arithmetic.rs
@@ -70,10 +70,6 @@ where
 }
 
 /// This is a 128-bit verifier challenge.
-///
-/// The verifier samples its challenge here as u^2, i.e. the square of the
-/// actual challenge. This is an optimisation that is documented in Section 6.3
-/// of the [Halo](https://eprint.iacr.org/2019/1021) paper.
 #[derive(Copy, Clone, Debug)]
 pub struct Challenge(pub(crate) u128);
 
--- a/src/arithmetic.rs
+++ b/src/arithmetic.rs
@@ -70,6 +70,10 @@ where
 }
 
 /// This is a 128-bit verifier challenge.
+///
+/// The verifier samples its challenge here as u^2, i.e. the square of the
+/// actual challenge. This is an optimisation that is documented in Section 6.3
+/// of the [Halo](https://eprint.iacr.org/2019/1021) paper.
 #[derive(Copy, Clone, Debug)]
 pub struct Challenge(pub(crate) u128);
 
--- a/src/plonk/verifier.rs
+++ b/src/plonk/verifier.rs
@@ -1,4 +1,4 @@
-use super::{hash_point, Proof, SRS};
+use super::{hash_point, Error, Proof, SRS};
 use crate::arithmetic::{get_challenge_scalar, Challenge, Curve, CurveAffine, Field};
 use crate::poly::{
     commitment::{Params, MSM},
@@ -6,14 +6,14 @@ use crate::poly::{
 };
 use crate::transcript::Hasher;
 
-impl<C: CurveAffine> Proof<C> {
+impl<'a, C: CurveAffine> Proof<C> {
     /// Returns a boolean indicating whether or not the proof is valid
     pub fn verify<HBase: Hasher<C::Base>, HScalar: Hasher<C::Scalar>>(
         &self,
-        params: &Params<C>,
+        params: &'a Params<C>,
         srs: &SRS<C>,
-        msm: MSM<C>,
-    ) -> bool {
+        msm: MSM<'a, C>,
+    ) -> Result<MSM<'a, C>, Error> {
         // Create a transcript for obtaining Fiat-Shamir challenges.
         let mut transcript = HBase::init(C::Base::one());
 
@@ -137,7 +137,7 @@ impl<C: CurveAffine> Proof<C> {
         }
 
         if h_eval != (expected_h_eval * &(x_3n - &C::Scalar::one())) {
-            return false;
+            return Err(Error::ConstraintSystemFailure);
         }
 
         // We are now convinced the circuit is satisfied so long as the
@@ -279,6 +279,6 @@ impl<C: CurveAffine> Proof<C> {
 
         let msm_challenges = guard.use_challenges();
 
-        msm_challenges.is_zero()
+        Ok(msm_challenges)
     }
 }
--- a/src/plonk/verifier.rs
+++ b/src/plonk/verifier.rs
@@ -265,7 +265,7 @@ impl<C: CurveAffine> Proof<C> {
 
         // Verify the opening proof
         let default_msm = MSM::default(&params);
-        let (challenges, guard) = self
+        let guard = self
             .opening
             .verify(
                 params,
@@ -277,7 +277,7 @@ impl<C: CurveAffine> Proof<C> {
             )
             .unwrap();
 
-        let msm: &MSM<C> = &guard.use_challenges(params, challenges).unwrap();
+        let msm: &MSM<C> = &guard.use_challenges(params).unwrap();
 
         msm.is_zero(params)
     }
--- a/src/plonk/verifier.rs
+++ b/src/plonk/verifier.rs
@@ -12,6 +12,7 @@ impl<C: CurveAffine> Proof<C> {
         &self,
         params: &Params<C>,
         srs: &SRS<C>,
+        msm: MSM<C>,
     ) -> bool {
         // Create a transcript for obtaining Fiat-Shamir challenges.
         let mut transcript = HBase::init(C::Base::one());
@@ -264,12 +265,11 @@ impl<C: CurveAffine> Proof<C> {
         }
 
         // Verify the opening proof
-        let default_msm = params.msm();
         let guard = self
             .opening
             .verify(
                 params,
-                default_msm,
+                msm,
                 &mut transcript,
                 x_6,
                 &f_commitment.to_affine(),
@@ -277,8 +277,8 @@ impl<C: CurveAffine> Proof<C> {
             )
             .unwrap();
 
-        let msm: &MSM<C> = &guard.use_challenges();
+        let msm_challenges = guard.use_challenges();
 
-        msm.is_zero()
+        msm_challenges.is_zero()
     }
 }
--- a/src/plonk/verifier.rs
+++ b/src/plonk/verifier.rs
@@ -1,6 +1,9 @@
 use super::{hash_point, Proof, SRS};
 use crate::arithmetic::{get_challenge_scalar, Challenge, Curve, CurveAffine, Field};
-use crate::poly::{commitment::Params, Rotation};
+use crate::poly::{
+    commitment::{Params, MSM},
+    Rotation,
+};
 use crate::transcript::Hasher;
 
 impl<C: CurveAffine> Proof<C> {
@@ -261,12 +264,20 @@ impl<C: CurveAffine> Proof<C> {
         }
 
         // Verify the opening proof
-        self.opening.verify(
-            params,
-            &mut transcript,
-            x_6,
-            &f_commitment.to_affine(),
-            f_eval,
-        )
+        let (challenges, mut guard) = self
+            .opening
+            .verify(
+                params,
+                &mut MSM::default(&params),
+                &mut transcript,
+                x_6,
+                &f_commitment.to_affine(),
+                f_eval,
+            )
+            .unwrap();
+
+        let msm: MSM<C> = guard.use_challenges(challenges).unwrap();
+
+        msm.is_zero()
     }
 }
--- a/src/plonk/verifier.rs
+++ b/src/plonk/verifier.rs
@@ -264,11 +264,12 @@ impl<C: CurveAffine> Proof<C> {
         }
 
         // Verify the opening proof
-        let (challenges, mut guard) = self
+        let default_msm = MSM::default(&params);
+        let (challenges, guard) = self
             .opening
             .verify(
                 params,
-                &mut MSM::default(&params),
+                default_msm,
                 &mut transcript,
                 x_6,
                 &f_commitment.to_affine(),
@@ -276,8 +277,8 @@ impl<C: CurveAffine> Proof<C> {
             )
             .unwrap();
 
-        let msm: MSM<C> = guard.use_challenges(challenges).unwrap();
+        let msm: &MSM<C> = &guard.use_challenges(params, challenges).unwrap();
 
-        msm.is_zero()
+        msm.is_zero(params)
     }
 }
--- a/src/plonk/verifier.rs
+++ b/src/plonk/verifier.rs
@@ -264,7 +264,7 @@ impl<C: CurveAffine> Proof<C> {
         }
 
         // Verify the opening proof
-        let default_msm = MSM::default(&params);
+        let default_msm = params.msm();
         let guard = self
             .opening
             .verify(
@@ -277,8 +277,8 @@ impl<C: CurveAffine> Proof<C> {
             )
             .unwrap();
 
-        let msm: &MSM<C> = &guard.use_challenges(params).unwrap();
+        let msm: &MSM<C> = &guard.use_challenges();
 
-        msm.is_zero(params)
+        msm.is_zero()
     }
 }
--- a/src/poly/commitment.rs
+++ b/src/poly/commitment.rs
@@ -5,8 +5,7 @@
 
 use super::{Coeff, Error, LagrangeCoeff, Polynomial};
 use crate::arithmetic::{
-    best_fft, best_multiexp, get_challenge_scalar, parallelize, Challenge, Curve, CurveAffine,
-    Field,
+    best_fft, best_multiexp, parallelize, Challenge, Curve, CurveAffine, Field,
 };
 use crate::transcript::Hasher;
 use std::ops::{Add, AddAssign, Mul, MulAssign};
@@ -249,25 +248,17 @@ impl<C: CurveAffine> Params<C> {
 /// A guard returned by the verifier
 #[derive(Debug)]
 pub struct Guard<C: CurveAffine> {
-    /// MSM
     msm: MSM<C>,
-
-    /// Negation of z1 value in the OpeningProof
     neg_z1: C::Scalar,
-
     allinv: C::Scalar,
-
     challenges_sq: Vec<C::Scalar>,
+    challenges_sq_packed: Vec<Challenge>,
 }
 
 impl<C: CurveAffine> Guard<C> {
     /// Lets caller supply the challenges and obtain an MSM with updated
     /// scalars and points.
-    pub fn use_challenges(
-        mut self,
-        params: &Params<C>,
-        challenges_sq_packed: Vec<Challenge>,
-    ) -> Result<MSM<C>, Error> {
+    pub fn use_challenges(mut self, params: &Params<C>) -> Result<MSM<C>, Error> {
         let mut scalars: Vec<C::Scalar> = vec![];
         let mut bases: Vec<C> = vec![];
 
@@ -281,22 +272,7 @@ impl<C: CurveAffine> Guard<C> {
         }
 
         // - [z1] G
-        let mut allinv = C::Scalar::one();
-        let mut challenges_sq = Vec::with_capacity(params.k as usize);
-
-        for challenge_sq_packed in challenges_sq_packed.iter() {
-            let challenge_sq: C::Scalar = get_challenge_scalar(*challenge_sq_packed);
-            challenges_sq.push(challenge_sq);
-
-            let challenge = challenge_sq.deterministic_sqrt();
-            let challenge = challenge.unwrap();
-
-            let challenge_inv = challenge.invert();
-            let challenge_inv = challenge_inv.unwrap();
-            allinv *= &challenge_inv;
-        }
-
-        let s = compute_s(&challenges_sq, allinv * &self.neg_z1);
+        let s = compute_s(&self.challenges_sq, self.allinv * &self.neg_z1);
         scalars.extend(&s);
         bases.extend(&params.g);
 
@@ -308,7 +284,7 @@ impl<C: CurveAffine> Guard<C> {
     /// Lets caller supply the purported G point and simply appends it to
     /// return an updated MSM.
     pub fn use_g(mut self, g: C) -> Result<MSM<C>, Error> {
-        &self.msm.other_scalars.push(self.allinv * &self.neg_z1);
+        &self.msm.other_scalars.push(self.neg_z1);
         &self.msm.other_bases.push(g);
 
         Ok(self.msm)
@@ -436,11 +412,11 @@ fn test_opening_proof() {
             let opening_proof = opening_proof.unwrap();
             // Verify the opening proof
             let msm = MSM::default(&params);
-            let (challenges, guard) = opening_proof
+            let guard = opening_proof
                 .verify(&params, msm, &mut transcript_dup, x, &p, v)
                 .unwrap();
 
-            let msm = guard.use_challenges(&params, challenges).unwrap();
+            let msm = guard.use_challenges(&params).unwrap();
 
             assert!(msm.is_zero(&params));
             break;
--- a/src/poly/commitment.rs
+++ b/src/poly/commitment.rs
@@ -23,13 +23,14 @@ pub struct OpeningProof<C: CurveAffine> {
     z2: C::Scalar,
 }
 
-/// TODO: documentation
+/// An accumulator instance consisting of an evaluation claim and a proof.
 #[derive(Debug, Clone)]
 pub struct Accumulator<C: CurveAffine> {
-    /// TODO: documentation
+    /// The claimed output of the linear-time polycommit opening protocol
     pub g: C,
 
-    /// TODO: documentation
+    /// A vector of 128-bit challenges sampled by the verifier, to be used in
+    /// computing g.
     pub challenges_sq_packed: Vec<Challenge>,
 }
 
--- a/src/poly/commitment.rs
+++ b/src/poly/commitment.rs
@@ -3,8 +3,11 @@
 //!
 //! [halo]: https://eprint.iacr.org/2019/1021
 
-use super::{Coeff, LagrangeCoeff, Polynomial};
-use crate::arithmetic::{best_fft, best_multiexp, parallelize, Curve, CurveAffine, Field};
+use super::{Coeff, Error, LagrangeCoeff, Polynomial};
+use crate::arithmetic::{
+    best_fft, best_multiexp, get_challenge_scalar, parallelize, Challenge, Curve, CurveAffine,
+    Field,
+};
 use crate::transcript::Hasher;
 use std::ops::{Add, AddAssign, Mul, MulAssign};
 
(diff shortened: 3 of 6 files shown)


src/poly/commitment.rs
```diff
@@ -154,6 +217,95 @@ impl<C: CurveAffine> Params<C> {
...
+
+    /// Points produced by the verifier for multiscalar multiplication
+    pub bases: Vec<C>,
+}
+
+impl<'a, C: CurveAffine> Guard<'a, C> {
+    /// Lets caller supply the challenges and obtain an MSM with updated
+    /// scalars and points.
+    pub fn use_challenges(
+        &mut self,
```
ebfull, line 246: The `Guard` pattern typically involves consuming the `Guard` (taking it by `self`).

src/poly/commitment.rs
```diff
@@ -154,6 +217,95 @@ impl<C: CurveAffine> Params<C> {
...
+            challenges_sq.push(challenge_sq);
+
+            let challenge = challenge_sq.deterministic_sqrt();
+            if challenge.is_none() {
+                // We didn't sample a square.
+                return Err(Error::OpeningError);
+            }
+            let challenge = challenge.unwrap();
+
+            let challenge_inv = challenge.invert();
```
ebfull, line 264: Instead of inverting here, you could have a `challenges` vector outside of the loop that you push to. Then, you could do `let allinv = C::Scalar::batch_invert(&mut challenges);` to batch the inversions together.

src/poly/commitment.rs
```diff
@@ -154,6 +217,95 @@ impl<C: CurveAffine> Params<C> {
...
+}
+
+impl<'a, C: CurveAffine> Guard<'a, C> {
+    /// Lets caller supply the challenges and obtain an MSM with updated
+    /// scalars and points.
+    pub fn use_challenges(
+        &mut self,
+        challenges_sq_packed: Vec<Challenge>,
+    ) -> Result<MSM<C>, Error> {
+        // - [z1] G
```
ebfull, line 249: This comment is out of position/context.

src/poly/commitment.rs
```diff
@@ -154,6 +217,95 @@ impl<C: CurveAffine> Params<C> {
...
+            g: self.g.clone(),
+            h: self.h.clone(),
+            scalars: self.scalars.clone(),
+            bases: self.bases.clone(),
+        })
+    }
+
+    /// Lets caller supply the purported G point and simply appends it to
+    /// return an updated MSM.
+    pub fn use_s(&mut self, g: Vec<C>, mut s: Vec<C::Scalar>) -> Result<MSM<C>, Error> {
```
ebfull, line 292: This function should really be called `use_g` because it is the `g` point that is just appended to the MSM (with `neg_z` as its scalar) and there is no `s` vector. It's actually the intention that the `s` vector does not need to be computed if the `g` point is provided, since in that context it is part of an advice string.

src/poly/commitment.rs
```diff
@@ -154,6 +217,95 @@ impl<C: CurveAffine> Params<C> {
...
+
+        self.bases.extend(&self.g);
+        let mut s = compute_s(&challenges_sq, allinv);
+        // TODO: parallelize
+        for s in &mut s {
+            *s *= &self.neg_z1;
+        }
+        self.scalars.extend(s);
+
+        Ok(MSM {
```
ebfull, line 282: This returns a new `MSM`, but what we want is to operate on (and return) the MSM located in `self`.

src/poly/commitment.rs
```diff
@@ -154,6 +217,95 @@ impl<C: CurveAffine> Params<C> {
...
+    /// return an updated MSM.
+    pub fn use_s(&mut self, g: Vec<C>, mut s: Vec<C::Scalar>) -> Result<MSM<C>, Error> {
+        // - [z1] G
+        self.bases.extend(&g);
+        for s in &mut s {
+            *s *= &self.neg_z1;
+        }
+        self.scalars.extend(s);
+
+        Ok(MSM {
```
ebfull, line 300: Mirroring before, we'll have an `msm` field in this `Guard` structure that we operate on and return to the caller.

src/arithmetic.rs
```diff
@@ -70,6 +70,10 @@ where
 }
 
 /// This is a 128-bit verifier challenge.
+///
+/// The verifier samples its challenge here as u^2, i.e. the square of the
```
ebfull, line 74: These details are specific to the `Challenge` as it's used in the commitment scheme, but there are other challenges in PLONK that use this structure too. This comment can be moved in the commitment scheme implementation file.

ebfull (changes requested):
We need to make some adjustments to the `MSM`/`Guard` structures. `Params` could have a method called `new_msm()` which returns an `MSM` that holds a reference to the `Params`. Then, the MSM can be passed by value through to APIs and returned from them.

In fact, it might even be best for the `Guard` structure to store a copy of the challenges (and allinv/challenges_sq which the verifier probably can compute without additional inversions/sqrts) which it returns when you call `use_g` alongside the `MSM`. Otherwise, it uses it in `get_s` to operate on the MSM and returns just the MSM, since the challenges don't matter anymore.

src/poly/commitment.rs
```diff
@@ -154,6 +217,95 @@ impl<C: CurveAffine> Params<C> {
     }
 }
 
+/// A guard returned by the verifier
+#[derive(Debug)]
+pub struct Guard<'a, C: CurveAffine> {
```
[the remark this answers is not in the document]
ebfull: As I mentioned in the review comment above, it could actually also contain the `challenges_sq`/`allinv` values produced by the verifier so that they don't have to be recomputed in `use_s`.

src/poly/commitment.rs
ebfull: Alternatively, we don't need to compute this again. It can all be stored in the guard object.

src/poly/commitment.rs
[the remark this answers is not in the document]
therealyingtong: `Guard` now looks like: 

```
pub struct Guard<'a, C: CurveAffine> {
    msm: &'a mut MSM<C>,
    neg_z1: C::Scalar,
    allinv: C::Scalar,
    challenges_sq: Vec<C::Scalar>,
}
```

src/poly/commitment.rs
therealyingtong: i'm storing `allinv` in `Guard` to be used in `use_g()`. but i don't think it should be used in `use_challenges` because `allinv` is derived from `challenges` which should be provided by the caller here.

src/poly/commitment.rs
therealyingtong: `use_challenges` and `use_g` now take `mut self` instead of `&mut self`.

src/poly/commitment.rs
therealyingtong: (i've moved this comment down, not sure why the change isn't being reflected here)

src/plonk/verifier.rs
```diff
@@ -261,12 +264,21 @@ impl<C: CurveAffine> Proof<C> {
...
 
         // Verify the opening proof
-        self.opening.verify(
-            params,
-            &mut transcript,
-            x_6,
-            &f_commitment.to_affine(),
-            f_eval,
-        )
+        let default_msm = MSM::default(&params);
```
ebfull, line 267: I think we should have a method called `msm` on `Params` instead of doing `MSM::default()`, but in any case, we shouldn't call it `default()` because that conflicts with the `Default::default()` method signature, which is in the prelude and so in scope.

src/poly/commitment.rs
```diff
@@ -21,6 +23,95 @@ pub struct OpeningProof<C: CurveAffine> {
     z2: C::Scalar,
 }
 
+/// A multiscalar multiplication in the polynomial commitment scheme
+#[derive(Debug)]
+pub struct MSM<C: CurveAffine> {
```
ebfull, line 28: Sorry that I forgot to put this field in there before, but we should put a pointer to Params in MSM.

```rust
pub struct MSM<'a, C: CurveAffine> {
    params: &'a Params<C>,

    /// TODO: documentation
    g_scalars: Option<Vec<C::Scalar>>,

    /// TODO: documentation
    h_scalar: Option<C::Scalar>,

    /// TODO: documentation
    other_scalars: Vec<C::Scalar>,

    /// TODO: documentation
    other_bases: Vec<C>,
}
```

That way, whenever you call methods of `MSM` you don't have to provide the `Params`.

src/poly/commitment.rs
```diff
@@ -21,6 +23,95 @@ pub struct OpeningProof<C: CurveAffine> {
...
+    /// TODO: documentation
+    pub g_scalars: Option<Vec<C::Scalar>>,
```
ebfull, line 30: These fields should be private.

src/poly/commitment.rs
```diff
@@ -21,6 +23,95 @@ pub struct OpeningProof<C: CurveAffine> {
...
+    pub other_scalars: Vec<C::Scalar>,
+
+    /// TODO: documentation
+    pub other_bases: Vec<C>,
+}
+
+impl<'a, C: CurveAffine> MSM<C> {
+    /// Empty MSM
+    pub fn default(params: &Params<C>) -> Self {
+        let g_scalars = Some(vec![C::Scalar::one(); params.n as usize]);
```
ebfull, line 45: As mentioned, this should be a method of `Params`, but also we shouldn't be initializing `g_scalars` and `h_scalar` to anything. We can set them to both to `None` that way if we call the `MSM` and there are no scalars, it doesn't try to do a multiexp using the `g` and `h` bases from params, since it doesn't need to.

src/poly/commitment.rs
```diff
@@ -21,6 +23,95 @@ pub struct OpeningProof<C: CurveAffine> {
...
+    }
+
+    /// Add arbitrary term (the scalar and the point)
+    pub fn add_term(&mut self, scalar: C::Scalar, point: C) {
+        &self.other_scalars.push(scalar);
+        &self.other_bases.push(point);
+    }
+
+    /// Add a vector of scalars to `g_scalars`
+    pub fn add_to_g(&mut self, scalars: Vec<C::Scalar>) {
```
ebfull, line 65: Since we aren't mutating `scalars` it'd be better to take it as a slice: `pub fn add_to_g(&mut self, scalars: &[C::Scalar]) {`.

src/poly/commitment.rs
```diff
@@ -21,6 +23,95 @@ pub struct OpeningProof<C: CurveAffine> {
...
+        for (g_scalar, scalar) in self
```
ebfull, line 66: This code should check to see if `self.g_scalars` is `None`, and if so it should set it to `Some(scalars.to_vec())`. Otherwise in case it is `Some` then you can perform the loop to perform the addition as you've written it.

src/poly/commitment.rs
```diff
@@ -21,6 +23,95 @@ pub struct OpeningProof<C: CurveAffine> {
...
+            .iter_mut()
+            .zip(scalars.iter())
+        {
+            *g_scalar += &scalar;
+        }
+    }
+
+    /// Add term to h
+    pub fn add_to_h(&mut self, scalar: C::Scalar) {
+        self.h_scalar = Some(self.h_scalar.unwrap() + &scalar);
```
ebfull, line 79: Similar to above, check that h_scalar is `None` first and simply set it. This can be done using

`self.h_scalar = self.h_scalar.map(|a| a + &scalar).or_else(Some(scalar));`

or something like that.

src/poly/commitment.rs
```diff
@@ -21,6 +23,95 @@ pub struct OpeningProof<C: CurveAffine> {
...
+    }
+
+    /// Scale all scalars in the MSM by a random blinding factor
+    pub fn scale(&mut self, factor: C::Scalar) {
+        for g_scalar in self.g_scalars.as_mut().unwrap().iter_mut() {
```
ebfull, line 84: You'll need to check that `g_scalars` isn't `None` before calling this loop, or else this `unwrap()` will assert.

src/poly/commitment.rs
```diff
@@ -21,6 +23,95 @@ pub struct OpeningProof<C: CurveAffine> {
...
+            *g_scalar *= &factor;
+        }
+        for other_scalar in self.other_scalars.iter_mut() {
+            *other_scalar *= &factor;
+        }
+        self.h_scalar = Some(self.h_scalar.unwrap() * &factor);
```
ebfull, line 90: Similarly, this should be done like `self.h_scalar = self.h_scalar.map(|a| a * &factor);` which will leave a `None` if it's already `None`, which is what we want.

src/poly/commitment.rs
```diff
@@ -21,6 +23,95 @@ pub struct OpeningProof<C: CurveAffine> {
...
+    }
+
+    /// Perform multiexp and check that it results in zero
+    pub fn is_zero(&self, params: &'a Params<C>) -> bool {
```
ebfull, line 94: This structure will contain `params` so you'll not need to take it as an argument.

src/poly/commitment.rs
```diff
@@ -21,6 +23,95 @@ pub struct OpeningProof<C: CurveAffine> {
...
+        let mut scalars: Vec<C::Scalar> = vec![];
+        let mut bases: Vec<C> = vec![];
```
ebfull, line 96: nit: It's possible to do `Vec::with_capacity` on these since we know in advance how many scalars/bases we have.

```rust
let len = self.g_scalars.as_ref().map(|v| v.len()).unwrap_or(0)
          + self.h_scalar.map(|_| 1).unwrap_or(0)
          + self.other_scalars.len();
let mut scalars: Vec<C::Scalar> = Vec::with_capacity(len);
let mut bases: Vec<C> = Vec::with_capacity(len);
```

src/poly/commitment.rs
```diff
@@ -21,6 +23,95 @@ pub struct OpeningProof<C: CurveAffine> {
...
+    pub other_bases: Vec<C>,
+}
+
+impl<'a, C: CurveAffine> MSM<C> {
+    /// Empty MSM
+    pub fn default(params: &Params<C>) -> Self {
+        let g_scalars = Some(vec![C::Scalar::one(); params.n as usize]);
+        let h_scalar = Some(C::Scalar::one());
+        let other_scalars: Vec<C::Scalar> = Vec::with_capacity(params.k as usize * 2 + 3);
+        let other_bases: Vec<C> = Vec::with_capacity(params.k as usize * 2 + 3);
```
ebfull, line 48: I think that in practice because the MSM will be (eventually) passed into the plonk verifier, these initial capacities will be lower. This is typically a small vector anyway so we can just initialize these to `vec![]`.

src/poly/commitment.rs
```diff
@@ -154,6 +245,52 @@ impl<C: CurveAffine> Params<C> {
...
+    neg_z1: C::Scalar,
+    allinv: C::Scalar,
+    challenges_sq: Vec<C::Scalar>,
+    challenges_sq_packed: Vec<Challenge>,
+}
+
+impl<C: CurveAffine> Guard<C> {
+    /// Lets caller supply the challenges and obtain an MSM with updated
+    /// scalars and points.
+    pub fn use_challenges(mut self, params: &Params<C>) -> Result<MSM<C>, Error> {
```
ebfull, line 261: This function can just return `MSM<C>` since it cannot fail.

src/poly/commitment.rs
```diff
@@ -154,6 +245,52 @@ impl<C: CurveAffine> Params<C> {
...
+        let mut scalars: Vec<C::Scalar> = vec![];
```
ebfull, line 262: We shouldn't need to keep a `scalars`/`bases` temporaries around here. The MSM will do everything for us.

We just need to do

```rust
        // - [z1] G
        let s = compute_s(&self.challenges_sq, self.allinv * &self.neg_z1);
        self.msm.add_to_g(&s);

        self.msm
```

src/poly/commitment.rs
```diff
@@ -154,6 +245,52 @@ impl<C: CurveAffine> Params<C> {
...
+        let mut bases: Vec<C> = vec![];
+
+        scalars.extend(&self.msm.other_scalars);
+        bases.extend(&self.msm.other_bases);
+
+        // - [z2] H
+        if let Some(h_scalar) = self.msm.h_scalar {
```
ebfull, line 269: This operation is unnecessary because the commitment verifier will emplace the correct `h_scalar` into the `MSM` for us.

src/poly/commitment.rs
```diff
@@ -154,6 +245,52 @@ impl<C: CurveAffine> Params<C> {
...
+        bases.extend(&params.g);
+
+        self.msm.g_scalars = Some(s);
+
+        Ok(self.msm)
+    }
+
+    /// Lets caller supply the purported G point and simply appends it to
+    /// return an updated MSM.
+    pub fn use_g(mut self, g: C) -> Result<MSM<C>, Error> {
```
ebfull, line 286: Remove the `Error` from the return here also.

src/poly/commitment/verifier.rs
```diff
@@ -31,28 +31,26 @@ impl<C: CurveAffine> OpeningProof<C> {
...
             if u_y.is_none() {
-                return false;
+                return Err(Error::OpeningError);
             }
             let u_y = u_y.unwrap();
 
             C::from_xy(u_x, u_y).unwrap()
         };
 
-        let mut extra_scalars = Vec::with_capacity(self.rounds.len() * 2 + 4 + params.n as usize);
```
ebfull, line 41: We'll want to keep these working vectors around still, rather than trying to access the `MSM` and modify them. When we're ready later on in this function (after the challenge `c` is sampled), we'll just loop on `extra_scalars`/`extra_bases` and call `msm.add_term`.

src/poly/commitment/verifier.rs
```diff
@@ -109,7 +108,7 @@ impl<C: CurveAffine> OpeningProof<C> {
         // [c] P + [c * v] U + [c] sum(L_i * u_i^2) + [c] sum(R_i * u_i^-2) + delta - [z1] G - [z1 * b] U - [z2] H
```
ebfull, line 108: Before we do anything with the `msm` object we need to do

```rust
msm.scale(C::Scalar::random());
```

This will ensure that if the existing `MSM` has `is_zero() == true` then this argument won't be able to interfere with it. It's a way of keeping the msm's linearly independent.

src/poly/commitment/verifier.rs
```diff
@@ -118,31 +117,29 @@ impl<C: CurveAffine> OpeningProof<C> {
...
-
-        // - [z1] G
-        extra_bases.extend(&params.g);
-        let mut s = compute_s(&challenges_sq, allinv);
-        // TODO: parallelize
-        for s in &mut s {
-            *s *= &neg_z1;
-        }
-        extra_scalars.extend(s);
+        msm.other_bases.push(self.delta);
```
ebfull, line 128: Here we should be doing `msm.add_term` for all this.

src/poly/commitment/verifier.rs
```diff
@@ -118,31 +117,29 @@ impl<C: CurveAffine> OpeningProof<C> {
...
+        msm.other_scalars.push(Field::one());
+
+        // z2
+        msm.h_scalar = Some(-self.z2);
```
ebfull, line 132: This should be `msm.add_to_h(-self.z2)`.

ebfull (changes requested):

src/plonk/verifier.rs
```diff
@@ -261,12 +264,21 @@ impl<C: CurveAffine> Proof<C> {
...
+                params,
+                default_msm,
+                &mut transcript,
+                x_6,
+                &f_commitment.to_affine(),
+                f_eval,
+            )
+            .unwrap();
+
+        let msm: &MSM<C> = &guard.use_challenges();
```
ebfull, line 280: Reference and type annotation can be removed.

src/poly/commitment/verifier.rs
```diff
@@ -109,40 +111,45 @@ impl<C: CurveAffine> OpeningProof<C> {
...
-        extra_bases.extend(&params.g);
-        let mut s = compute_s(&challenges_sq, allinv);
-        // TODO: parallelize
-        for s in &mut s {
-            *s *= &neg_z1;
-        }
-        extra_scalars.extend(s);
+        msm.add_term(Field::one(), self.delta);
+
+        // z2
```
ebfull, line 141: Comment could be changed to `- [z2] H`

ebfull (changes requested):
Perfect! It'd be nice (for testing purposes) to do a couple things:

1. Let's make it so that `MSM` and `Guard` are `Clone`.
2. Let's break the `use_challenges` method of `Guard` up into another method `compute_g(&self, scalar: C::Scalar)` which computes the g value when given a potential scalar like `neg_z1` as input. Then, we can write tests that `use_g` works correctly by cloning the `MSM` and verifying both by calling `use_challenges()` and by calling `use_g()` with the output of `compute_g(C::Scalar::one())`.
3. Let's introduce a `pub struct Accumulator<C> { pub g: C, pub challenges_sq_packed: Vec<Challenge> }` and return it alongside the `MSM` in `use_g`.
4. Modify the signature of the PLONK verifier to take in an MSM which is passed to the commitment verifier and then returned from the PLONK verifier. This way we can exercise batch verification of PLONK proofs in tests.

src/plonk/verifier.rs
```diff
@@ -261,12 +265,20 @@ impl<C: CurveAffine> Proof<C> {
...
+                &mut transcript,
+                x_6,
+                &f_commitment.to_affine(),
+                f_eval,
+            )
+            .unwrap();
+
+        let msm_challenges = guard.use_challenges();
+
+        msm_challenges.is_zero()
```
ebfull, line 282: Instead, let's return the MSM so that the caller can use it for the next proof they verify.

src/poly/commitment.rs
```diff
@@ -21,6 +23,89 @@ pub struct OpeningProof<C: CurveAffine> {
...
+    /// Add arbitrary term (the scalar and the point)
+    pub fn add_term(&mut self, scalar: C::Scalar, point: C) {
+        &self.other_scalars.push(scalar);
+        &self.other_bases.push(point);
+    }
+
+    /// Add a vector of scalars to `g_scalars`
+    pub fn add_to_g(&mut self, scalars: &[C::Scalar]) {
+        if let Some(g_scalars) = &mut self.g_scalars {
+            for (g_scalar, scalar) in g_scalars.iter_mut().zip(scalars.iter()) {
```
ebfull, line 57: Add a `// TODO: parallelize` comment

src/poly/commitment.rs
```diff
@@ -21,6 +23,89 @@ pub struct OpeningProof<C: CurveAffine> {
...
+
+impl<'a, C: CurveAffine> MSM<'a, C> {
+    /// Add arbitrary term (the scalar and the point)
+    pub fn add_term(&mut self, scalar: C::Scalar, point: C) {
+        &self.other_scalars.push(scalar);
+        &self.other_bases.push(point);
+    }
+
+    /// Add a vector of scalars to `g_scalars`
+    pub fn add_to_g(&mut self, scalars: &[C::Scalar]) {
```
ebfull, line 55: We should add an assertion at the beginning of this function that `assert_eq!(scalars.len(), self.params.n)`.

src/poly/commitment.rs
```diff
@@ -152,6 +237,81 @@ impl<C: CurveAffine> Params<C> {
 
         best_multiexp::<C>(&tmp_scalars, &tmp_bases)
     }
+
+    /// Generates an empty multiscalar multiplication struct using the
+    /// appropriate params.
+    pub fn msm(&self) -> MSM<C> {
```
ebfull, line 243: Perhaps this can be renamed to `empty_msm` but it's not a big deal either way.

src/poly/commitment.rs
```diff
@@ -152,6 +237,81 @@ impl<C: CurveAffine> Params<C> {
...
+            challenges_sq_packed: self.challenges_sq_packed,
+        };
+
+        (self.msm, accumulator)
+    }
+
+    /// Computes the g value when given a potential scalar as input.
+    pub fn compute_g(&self, scalar: C::Scalar) -> C {
+        let s = compute_s(&self.challenges_sq, self.allinv * &scalar);
+        let mut g = C::Projective::zero();
```
ebfull, line 295: Instead, you can just return `best_multiexp(&s, &self.params.g).to_affine()`. The remaining logic in this function tries to interact with `g_scalars` but that is unrelated to the computation of the `G` element; in fact it will cause a bug if the MSM is reused after `use_challenges` is called.

src/poly/commitment.rs
```diff
@@ -273,8 +433,40 @@ fn test_opening_proof() {
             transcript.absorb(Field::one());
         } else {
             let opening_proof = opening_proof.unwrap();
-            assert!(opening_proof.verify(&params, &mut transcript_dup, x, &p, v));
+            // Verify the opening proof
+            let msm = params.msm();
+            let guard = opening_proof
```
ebfull, line 438: One way to improve this test so it can catch the bug currently in `MSM::compute_g` is to call `use_challenges` on this guard to get a new msm, and use that msm to verify the same `opening_proof` again and get a new guard. That guard can then proceed with the rest of the test as it's currently written. This way the `g_scalars` field of the `MSM` is populated already and you can see that your changes to `MSM::compute_g` fixed the issue.

src/poly/commitment.rs
ebfull: Also, the documentation for the function should say that it will panic if the caller provides a slice of scalars that is not of length `params.n`.

ebfull (changes requested):

src/poly/commitment.rs
```diff
@@ -21,6 +23,89 @@ pub struct OpeningProof<C: CurveAffine> {
...
+
+    /// Add term to h
+    pub fn add_to_h(&mut self, scalar: C::Scalar) {
+        self.h_scalar = self.h_scalar.map_or(Some(scalar), |a| Some(a + &scalar));
+    }
+
+    /// Scale all scalars in the MSM by a random blinding factor
+    pub fn scale(&mut self, factor: C::Scalar) {
+        if let Some(g_scalars) = &mut self.g_scalars {
+            for g_scalar in g_scalars.iter_mut() {
```
ebfull, line 73: Add a `// TODO: parallelize` comment

src/poly/commitment.rs
```diff
@@ -21,6 +23,89 @@ pub struct OpeningProof<C: CurveAffine> {
...
+                *g_scalar *= &factor;
+            }
+        }
+
+        for other_scalar in self.other_scalars.iter_mut() {
+            *other_scalar *= &factor;
```
ebfull, line 79: Add a `// TODO: parallelize` comment

src/poly/commitment.rs
```diff
@@ -152,6 +242,63 @@ impl<C: CurveAffine> Params<C> {
...
+    allinv: C::Scalar,
+    challenges_sq: Vec<C::Scalar>,
+    challenges_sq_packed: Vec<Challenge>,
+}
+
+impl<'a, C: CurveAffine> Guard<'a, C> {
+    /// Lets caller supply the challenges and obtain an MSM with updated
+    /// scalars and points.
+    pub fn use_challenges(mut self) -> MSM<'a, C> {
+        let g = self.compute_g(self.neg_z1);
```
ebfull, line 278: My bad, this is the wrong approach. (I think I made a mistake in my previous review comment.)

Here, we should do

```rust
pub fn use_challenges(mut self) -> MSM<'a, C> {
    let s = compute_s(&self.challenges_sq, self.allinv * &self.neg_z1);
    self.msm.add_to_g(&s);

    self.msm
}
```

src/poly/commitment.rs
```diff
@@ -152,6 +242,63 @@ impl<C: CurveAffine> Params<C> {
...
+        let accumulator = Accumulator {
+            g,
+            challenges_sq_packed: self.challenges_sq_packed,
+        };
+
+        (self.msm, accumulator)
+    }
+
+    /// Computes the g value when given a potential scalar as input.
+    pub fn compute_g(&self, scalar: C::Scalar) -> C {
```
ebfull, line 298: This function can be changed to not take in a `scalar` now, in light of the mistake I made in my previous review comment.

src/poly/commitment.rs
```diff
@@ -21,6 +23,94 @@ pub struct OpeningProof<C: CurveAffine> {
...
+
+        // TODO: parallelize
+        for other_scalar in self.other_scalars.iter_mut() {
+            *other_scalar *= &factor;
+        }
+        self.h_scalar = self.h_scalar.map(|a| a * &factor);
+    }
+
+    /// Perform multiexp and check that it results in zero
+    pub fn is_zero(&self) -> bool {
```
ebfull, line 90: This should take `self` by value because it doesn't make sense to do anything with an `MSM` object after you've obtained the result of this method.

src/poly/commitment.rs
```diff
@@ -21,6 +23,94 @@ pub struct OpeningProof<C: CurveAffine> {
...
+        if let Some(h_scalar) = self.h_scalar {
+            scalars.push(h_scalar);
+            bases.push(self.params.h);
+        }
+
+        if let Some(g_scalars) = &self.g_scalars {
+            scalars.extend(g_scalars);
+            bases.extend(self.params.g.iter());
+        }
+
```
ebfull, line 109: We should add an assertion here that the length of `scalars` matches the capacity we specified before, since it's something we can locally guarantee.

```rust
assert_eq!(scalars.len(), len);
```

ebfull (changes requested):

ebfull (approved):

PR #20, opened 2020-09-15
Small multiexp

### LINKED ISSUE #11: Use multiexp in commitment opening prover
Right now, in `parallel_generator_collapse` we do

```rust
for (g_lo, g_hi) in g_lo.iter().zip(g_hi.iter()) {
    // TODO: could use multiexp
    tmp.push(((*g_lo) * challenge_inv) + &((*g_hi) * challenge));
}
```

This could be done instead with a small multiexp. Right now we only have Pippinger's multiexp implemented, but this would be more efficient using the standard "share doublings" approach, maybe with some w-NAF added in.

Furthermore, it may be possible to "defer" these individual multiexps to later rounds and just work over the original basis temporarily until the number of individual multiexps is reduced.

therealyingtong:
Closes #11


therealyingtong:
This gives a performance improvement of ~25%:

```
     Running target/release/deps/plonk-8cbd2efa918afe22
Gnuplot not found, using plotters backend
Benchmarking double-and-add: Collecting 100 samples in estimated                                                                  double-and-add          time:   [3.8046 ms 3.8290 ms 3.8546 ms]
                        change: [-23.181% -22.535% -21.930%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild
```

Cargo.toml
```diff
@@ -20,6 +20,13 @@ publish = false
 [package.metadata.docs.rs]
 rustdoc-args = [ "--html-in-header", "katex-header.html" ]
 
+[dev-dependencies]
+criterion = "0.3"
+
+[[bench]]
+name = "plonk"
```
ebfull, line 27: Probably should rename this to `arithmetic` or something since it's not actually testing plonk itself.

src/arithmetic.rs
```diff
@@ -180,6 +180,39 @@ fn multiexp_serial<C: CurveAffine>(coeffs: &[C::Scalar], bases: &[C], acc: &mut
     }
 }
 
+/// Performs a small multi-exponentiation operation.
+/// Uses the double-and-add algorithm with doublings shared across points.
+
+pub fn small_multiexp<C: CurveAffine>(coeffs: &[C::Scalar], bases: &[C]) -> C::Projective {
+    // Gets the bit at position `i`. Bits are numbered from 0 (least significant) to 7 (most significant).
+    fn get_bit_at(byte: u8, i: usize) -> bool {
```
ebfull, line 188: I think it'd be better to just inline this function as `((byte >> bit_idx) & 1) == 1` at the callsite rather than introducing a new function for it.

src/arithmetic.rs
```diff
@@ -180,6 +180,39 @@ fn multiexp_serial<C: CurveAffine>(coeffs: &[C::Scalar], bases: &[C], acc: &mut
...
+
+    // for byte idx
+    for byte_idx in (0..32).rev() {
+        // for bit idx
+        for bit_idx in (0..8).rev() {
+            acc = acc.double();
+            // for each coeff
+            for coeff_idx in 0..coeffs.len() {
+                if get_bit_at(coeffs[coeff_idx][byte_idx], bit_idx) {
+                    acc = acc + &bases[coeff_idx].to_projective();
```
ebfull, line 207: It turns out that because `acc` is projective but `bases[coeff_idx]` is affine, we can use mixed point addition which is a tiny bit faster. I think the following will work:

```rust
acc += bases[coeff_idx];
```

Hopefully, anyway. The rust compiler has a hard time projecting some of the operator overloading traits through associated types. :shrug:

src/arithmetic.rs
```diff
@@ -180,6 +180,39 @@ fn multiexp_serial<C: CurveAffine>(coeffs: &[C::Scalar], bases: &[C], acc: &mut
     }
 }
 
+/// Performs a small multi-exponentiation operation.
+/// Uses the double-and-add algorithm with doublings shared across points.
+
```
ebfull, line 185: nit: remove newline

src/poly/commitment/prover.rs
```diff
@@ -220,15 +220,11 @@ fn parallel_generator_collapse<C: CurveAffine>(
     challenge_inv: C::Scalar,
 ) {
     let len = g.len() / 2;
-    let (mut g_lo, g_hi) = g.split_at_mut(len);
-
-    parallelize(&mut g_lo, |g_lo, start| {
```
ebfull, line 225: We still want to do the parallelization exactly as we've done before, it's just that we want to replace the line

```rust
tmp.push(((*g_lo) * challenge_inv) + &((*g_hi) * challenge));
```

with

```rust
tmp.push(small_multiexp(&[challenge_inv, challenge], &[*g_lo, *g_hi]));
```

ebfull (changes requested):

src/arithmetic.rs
```diff
@@ -180,6 +180,30 @@ fn multiexp_serial<C: CurveAffine>(coeffs: &[C::Scalar], bases: &[C], acc: &mut
...
+    // for byte idx
+    for byte_idx in (0..32).rev() {
+        // for bit idx
+        for bit_idx in (0..8).rev() {
+            acc = acc.double();
+            // for each coeff
+            for coeff_idx in 0..coeffs.len() {
+                let byte = coeffs[coeff_idx][byte_idx];
+                if (byte >> bit_idx & 1) != 0 {
```
ebfull, line 197: Can you add parenthesis around `byte >> bit_idx`? Apparently the order of operations is right already, but it's nice to be explicit.

ebfull (changes requested):

PR #21, opened 2020-09-15
Incorporate MSM/Guard into PLONK verifier API and arithmetic

### LINKED ISSUE #18: Incorporate MSM into PLONK verifier
Currently the PLONK verifier will manually compute the `f_commitment` for input into the commitment opening proof verifier. Instead, it should just incorporate this computation into the MSM that it feeds into the commitment verifier. Then, the responsibility for scaling the MSM randomly falls to the plonk verifier instead.

It also may be best to return the `Guard` from the PLONK verifier.

ebfull:
Closes #18

### DIFF
--- a/src/plonk/verifier.rs
+++ b/src/plonk/verifier.rs
@@ -256,17 +256,18 @@ impl<'a, C: CurveAffine> Proof<C> {
         let x_7: C::Scalar = get_challenge_scalar(Challenge(transcript.squeeze().get_lower_128()));
 
         // Compute the final commitment that has to be opened
-        msm.add_term(C::Scalar::one(), self.f_commitment);
+        let mut commitment_msm = params.empty_msm();
+        commitment_msm.add_term(C::Scalar::one(), self.f_commitment);
         for (_, &point_index) in srs.cs.rotations.iter() {
-            msm.scale(x_7);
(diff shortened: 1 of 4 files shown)


src/plonk/verifier.rs
```diff
@@ -256,29 +256,18 @@ impl<'a, C: CurveAffine> Proof<C> {
...
-        let mut f_commitment: C::Projective = self.f_commitment.to_projective();
+        let mut commitment_msm = params.empty_msm();
+        commitment_msm.add_term(C::Scalar::one(), self.f_commitment);
         for (_, &point_index) in srs.cs.rotations.iter() {
-            f_commitment *= x_7;
-            f_commitment = f_commitment + &q_commitments[point_index.0].as_ref().unwrap();
+            commitment_msm.scale(x_7);
+            commitment_msm.add_msm(&q_commitments[point_index.0]);
             f_eval *= &x_7;
             f_eval += &self.q_evals[point_index.0];
```
therealyingtong, line 265: we could consider renaming `f_eval` -> `commitment_eval`, since we've renamed `f_commitment` -> `commitment_msm`

therealyingtong (approved):
ACK with nonblocking comment.

PR #22, opened 2020-09-16
Support for auxiliary advice wires

### LINKED ISSUE #12: Support for auxiliary advice wires
In order to actually use the accumulation trick, we need "auxiliary" advice wires that are brought in from outside. These can engage in the argument itself, though the only example I can think where that is useful is with public inputs. Somehow, the proof structure exposes these auxiliary evaluations so that the verifier can check them if needed, which is the case with the accumulation.

therealyingtong:
Closes #12

### DIFF
--- a/src/plonk/circuit.rs
+++ b/src/plonk/circuit.rs
@@ -18,17 +18,6 @@ pub struct AdviceWire(pub usize);
 #[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
 pub struct AuxWire(pub usize);
 
-/// An enum over all wire types, to be used in permutations
-#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
-pub enum Wire {
-    /// Fixed wire
-    Fixed(FixedWire),
-    /// Advice wire
-    Advice(AdviceWire),
-    /// Auxiliary wire
-    Aux(AuxWire),
-}
-
 /// This trait allows a [`Circuit`] to direct some backend to assign a witness
 /// for a constraint system.
 pub trait Assignment<F: Field> {
@@ -187,7 +176,7 @@ pub struct ConstraintSystem<F> {
     // enforced between advice wire values in A, B and C, and another
     // permutation between wires (B, C, D) which allows the same with D instead
     // of A.
-    pub(crate) permutations: Vec<Vec<(Wire, usize)>>,
+    pub(crate) permutations: Vec<Vec<(AdviceWire, usize)>>,
 }
 
 impl<F: Field> Default for ConstraintSystem<F> {
@@ -211,7 +200,7 @@ impl<F: Field> Default for ConstraintSystem<F> {
 
 impl<F: Field> ConstraintSystem<F> {
     /// Add a permutation argument for some advice wires
-    pub fn permutation(&mut self, wires: &[Wire]) -> usize {
+    pub fn permutation(&mut self, wires: &[AdviceWire]) -> usize {
         let index = self.permutations.len();
         if index == 0 {
             let at = Rotation(-1);
@@ -220,11 +209,7 @@ impl<F: Field> ConstraintSystem<F> {
         }
         let wires = wires
             .iter()
-            .map(|&wire| match wire {
-                Wire::Advice(wire) => (Wire::Advice(wire), self.query_advice_index(wire, 0)),
-                Wire::Aux(wire) => (Wire::Aux(wire), self.query_aux_index(wire, 0)),
-                Wire::Fixed(wire) => (Wire::Fixed(wire), self.query_fixed_index(wire, 0)),
-            })
+            .map(|&wire| (wire, self.query_advice_index(wire, 0)))
             .collect();
         self.permutations.push(wires);
 
--- a/src/plonk/circuit.rs
+++ b/src/plonk/circuit.rs
@@ -14,6 +14,10 @@ pub struct FixedWire(pub usize);
 #[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
 pub struct AdviceWire(pub usize);
 
+/// This represents a wire which has an externally assigned value
+#[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
+pub struct AuxWire(pub usize);
+
 /// This trait allows a [`Circuit`] to direct some backend to assign a witness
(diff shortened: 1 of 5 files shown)


src/plonk/circuit.rs
```diff
@@ -6,21 +6,27 @@ use super::Error;
 use crate::arithmetic::Field;
 
 use crate::poly::Rotation;
-/// This represents a wire which has a fixed (permanent) value
+
+/// Enum over wires with variable values
 #[derive(Copy, Clone, Debug, Eq, PartialEq, Hash)]
-pub struct FixedWire(pub usize);
+pub enum Wire {
```
ebfull, line 12: I don't think we should do this. I think we should have a separate type `AuxWire` and in any APIs that need to be able to take in `AuxWire`/`AdviceWire`/`FixedWire` we can have a separate enum type `Wire` which implements `From<AuxWire>/From<AdviceWire>/From<FixedWire>`. In other cases we should be specific about the type.

src/plonk/circuit.rs
```diff
@@ -6,21 +6,27 @@ use super::Error;
...
+pub struct FixedWire(pub usize);
 
 /// This trait allows a [`Circuit`] to direct some backend to assign a witness
 /// for a constraint system.
 pub trait Assignment<F: Field> {
     /// Assign an advice wire value (witness)
     fn assign_advice(
         &mut self,
-        wire: AdviceWire,
+        wire: Wire,
```
ebfull, line 29: This is an example of why we shouldn't consolidate these types. It would be possible to call `assign_advice` on an aux wire.

src/plonk/circuit.rs
```diff
@@ -33,6 +39,14 @@ pub trait Assignment<F: Field> {
         to: impl FnOnce() -> Result<F, Error>,
     ) -> Result<(), Error>;
 
+    /// Assign an auxiliary wire value (external)
+    fn assign_aux(
```
ebfull, line 43: We don't need a way to assign the aux wire values from the prover side; those are fixed outside of the prover. Their values should be supplied as part of the call to `Proof::create`, and the commitments should be supplied as part of the call to `Proof::verify`.

src/plonk/prover.rs
```diff
@@ -27,23 +27,28 @@ impl<C: CurveAffine> Proof<C> {
...
+                    Wire::AdviceWire(idx) => {
+                        *self
+                            .advice
+                            .get_mut(idx)
+                            .and_then(|v| v.get_mut(row))
+                            .ok_or(Error::BoundsFailure)? = to()?;
+
+                        Ok(())
+                    }
+                    _ => return Err(Error::WireTypeError),
```
ebfull, line 50: Once you've reverted the unification of aux and advice wires, this error condition is no longer possible and so can be removed.

src/plonk/prover.rs
```diff
@@ -119,9 +144,52 @@ impl<C: CurveAffine> Proof<C> {
...
+                Wire::AdviceWire(idx) => {
+                    let poly = advice_polys[idx].clone();
+                    domain.coeff_to_extended(poly, at)
+                }
+                _ => unreachable!(),
+            })
+            .collect();
+
+        // Compute commitments to auxiliary wire polynomials
+        let aux_blinds: Vec<_> = witness
```
ebfull, line 157: Aux wires won't have blinding factors.

ebfull (changes requested):

src/plonk/prover.rs
```diff
@@ -24,6 +24,7 @@ impl<C: CurveAffine> Proof<C> {
         params: &Params<C>,
         srs: &SRS<C>,
         circuit: &ConcreteCircuit,
+        aux_lagrange_polys: Vec<Polynomial<C::Scalar, LagrangeCoeff>>,
```
ebfull, line 27: This should take a slice, not a `Vec`, since it's not being mutated.

src/plonk/prover.rs
```diff
@@ -152,15 +182,34 @@ impl<C: CurveAffine> Proof<C> {
...
-                        .zip(permuted_wire_values[start..].iter())
-                    {
-                        *modified_advice *= &(x_0 * permuted_advice_value + &x_1 + advice_value);
+                match wire {
+                    Wire::Advice(wire) => {
+                        parallelize(&mut modified_advice, |modified_advice, start| {
+                            for ((modified_advice, advice_value), permuted_advice_value) in
+                                modified_advice
+                                    .iter_mut()
+                                    .zip(witness.advice[wire.0][start..].iter())
```
ebfull, line 191: You could put a `match` inside what you pass to `zip` here in order to consolidate all of these loops together.

src/plonk/prover.rs
```diff
@@ -184,17 +233,38 @@ impl<C: CurveAffine> Proof<C> {
...
-                        *modified_advice *= &(deltaomega * &x_0 + &x_1 + advice_value);
-                        deltaomega *= &omega;
+                match wire {
+                    Wire::Advice(wire) => {
+                        parallelize(&mut modified_advice, |modified_advice, start| {
+                            let mut deltaomega =
+                                deltaomega * &omega.pow_vartime(&[start as u64, 0, 0, 0]);
+                            for (modified_advice, advice_value) in modified_advice
+                                .iter_mut()
+                                .zip(witness.advice[wire.0][start..].iter())
```
ebfull, line 243: Similarly, consolidate the cases together by putting the match at this level instead.

src/plonk/prover.rs
```diff
@@ -451,6 +529,17 @@ impl<C: CurveAffine> Proof<C> {
                 );
             }
 
+            for (query_index, &(wire, ref at)) in meta.aux_queries.iter().enumerate() {
+                let point_index = (*meta.rotations.get(at).unwrap()).0;
+
+                accumulate(
+                    point_index,
+                    &aux_polys[wire.0],
+                    Blind::default(),
```
ebfull, line 538: This will return Blind(1), we really want Blind(0) I think because auxiliary commitments won't have blinding factors.

ebfull (changes requested):

src/plonk/prover.rs
ebfull: TODO: open issue to revisit this

ebfull: https://github.com/zcash/halo2/issues/25

PR #24, opened 2020-09-19
Remove fork hack from OpeningProof::create()

### LINKED ISSUE #8: Remove fork hack from commitment opening implementation
Currently, the commitment opening implementation stores a `fork` value in the opening proof which the prover can vary until the point that is sampled at the beginning of the argument is on the curve. Let's remove this and instead make the proving fallible; the caller is then responsible for forking the transcript some other way (changing a blinding factor in a commitment) and trying again until it succeeds.

therealyingtong:
Add loop in PLONK prover to repeatedly call `OpeningProof::create()`, each time incrementing blinding factor `f_blind`, until correct value is sampled.

Closes #8

### DIFF
--- a/src/plonk/prover.rs
+++ b/src/plonk/prover.rs
@@ -530,21 +530,19 @@ impl<C: CurveAffine> Proof<C> {
                 .or_else(|| Some(poly));
         }
 
-        let final_opening;
-        let final_f_commitment;
+        let f_poly = f_poly.unwrap();
+        let mut f_blind = Blind(C::Scalar::random());
+        let mut f_commitment = params.commit(&f_poly, f_blind).to_affine();
+
         let final_q_evals;
 
-        loop {
-            let mut transcript_dup = transcript.clone();
-            let mut transcript_scalar_dup = transcript_scalar.clone();
-            let mut f_poly_dup = f_poly.clone().unwrap();
-
-            let mut f_blind = Blind(C::Scalar::random());
-            let f_commitment = params.commit(&f_poly_dup, f_blind).to_affine();
-            hash_point(&mut transcript_dup, &f_commitment)?;
+        let opening = loop {
+            let mut transcript = transcript.clone();
+            let mut transcript_scalar = transcript_scalar.clone();
+            hash_point(&mut transcript, &f_commitment)?;
 
             let x_6: C::Scalar =
-                get_challenge_scalar(Challenge(transcript_dup.squeeze().get_lower_128()));
+                get_challenge_scalar(Challenge(transcript.squeeze().get_lower_128()));
 
             let mut q_evals = vec![C::Scalar::zero(); meta.rotations.len()];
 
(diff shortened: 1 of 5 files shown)


src/plonk/prover.rs
```diff
@@ -529,50 +529,70 @@ impl<C: CurveAffine> Proof<C> {
...
+                        .zip(q_polys[point_index.0].as_ref().unwrap()[start..].iter())
+                    {
+                        *f *= &x_7;
+                        *f += a;
+                    }
+                });
+            }
 
-        // Let's prove that the q_commitment opens at x to the expected value.
-        let opening = OpeningProof::create(&params, &mut transcript, &f_poly, f_blind, x_6)
```
ebfull, line 574: I would recommend instead that we have `OpeningProof` itself fail, and then recover from that case to resume the `loop { }`. Otherwise, we're leaking implementation details of `OpeningProof` into the unrelated PLONK prover code.

src/plonk/prover.rs
```diff
@@ -529,50 +529,70 @@ impl<C: CurveAffine> Proof<C> {
...
-        let mut f_blind = Blind(C::Scalar::random());
 
-        let f_commitment = params.commit(&f_poly, f_blind).to_affine();
+        let final_opening;
+        let final_f_commitment;
+        let final_q_evals;
 
-        hash_point(&mut transcript, &f_commitment)?;
+        loop {
+            let mut transcript_dup = transcript.clone();
```
ebfull, line 538: It would be better to do

```rust
let mut transcript = transcript.clone();
let mut transcript_scalar = transcript_scalar.clone();
```

effectively shadowing `transcript`/`transcript_scalar` so that we don't accidentally interact with them inside the loop.

src/plonk/prover.rs
```diff
@@ -529,50 +529,70 @@ impl<C: CurveAffine> Proof<C> {
...
+            let mut transcript_scalar_dup = transcript_scalar.clone();
+            let mut f_poly_dup = f_poly.clone().unwrap();
 
-        let x_6: C::Scalar = get_challenge_scalar(Challenge(transcript.squeeze().get_lower_128()));
+            let mut f_blind = Blind(C::Scalar::random());
+            let f_commitment = params.commit(&f_poly_dup, f_blind).to_affine();
```
ebfull, line 543: We can compute this outside the loop and then within the loop just add `params.h` to it (and increment the corresponding `f_blind`) so that `params.commit` doesn't have to be called more than once.

ebfull (changes requested):

src/plonk/prover.rs
```diff
@@ -529,50 +529,65 @@ impl<C: CurveAffine> Proof<C> {
...
-        let f_commitment = params.commit(&f_poly, f_blind).to_affine();
+        let f_poly = f_poly.unwrap();
+        let mut f_blind = Blind(C::Scalar::random());
+        let mut f_commitment = params.commit(&f_poly, f_blind).to_affine();
 
-        hash_point(&mut transcript, &f_commitment)?;
+        let final_q_evals;
 
-        let x_6: C::Scalar = get_challenge_scalar(Challenge(transcript.squeeze().get_lower_128()));
+        let opening = loop {
```
ebfull, line 539: nit: this could be `let (opening, q_evals) = loop {` and you could get rid of `final_q_evals`. :)

PR #26, opened 2020-09-22
Optimise multi-point opening

### LINKED ISSUE #9: Optimize multi-point opening
Currently, the implementation randomly adds together commitments that are being opened at the same point, then performs a multi-point opening. However, this is inefficient when commitments are opened at multiple points, as then we have to do an unnecessary amount of arithmetic.

Instead, we'll make it so that the set of points at which every commitment is opened is queried separately, to minimize the amount of group arithmetic done by the verifier. It'll probably be nicer to break this abstraction out and separate it from the plonk prover; we could place it in the `poly` module somewhere.

therealyingtong:
Closes #9


src/plonk/prover.rs
```diff
@@ -454,209 +455,117 @@ impl<C: CurveAffine> Proof<C> {
...
-            advice_evals,
-            fixed_evals,
-            aux_evals,
-            h_evals,
-            f_commitment,
-            q_evals,
-            opening,
-        })
+        if let Ok(multiopening) =
+            multiopen::Proof::create(params, &mut transcript, &mut transcript_scalar, instances)
```
daira, line 552: I think using `.map_err(|_| Error::OpeningError)?` is clearer here.

src/poly/multiopen/prover.rs
```diff
@@ -0,0 +1,309 @@
...
+            let transcript_scalar_point =
+                C::Base::from_bytes(&(transcript_scalar.squeeze()).to_bytes()).unwrap();
+            transcript.absorb(transcript_scalar_point);
+
+            let x_7: C::Scalar =
+                get_challenge_scalar(Challenge(transcript.squeeze().get_lower_128()));
+
+            let mut f_blind_dup = f_blind;
+            let mut f_poly = f_poly.clone();
+            for (set_idx, _) in point_sets.iter().enumerate() {
```
ebfull, line 154: Probably can be replaced with

```rust
let (f_poly, f_blind) = q_polys.iter()
    .zip(q_blinds.iter())
    .fold((f_poly.clone(), f_blind), |(f_poly, f_blind), (poly, blind)| {
        (f_poly * x_7 + poly, f_blind * x_7 + blind)
    });
```

src/poly/multiopen/prover.rs
```diff
@@ -0,0 +1,309 @@
...
+            let mut transcript = transcript.clone();
+            let mut transcript_scalar = transcript_scalar.clone();
+            hash_point(&mut transcript, &f_commitment).unwrap();
+
+            let x_6: C::Scalar =
+                get_challenge_scalar(Challenge(transcript.squeeze().get_lower_128()));
+
+            let mut q_evals = vec![C::Scalar::zero(); point_sets.len()];
+
+            for (set_idx, _) in point_sets.iter().enumerate() {
```
ebfull, line 137: You're really just iterating over q_polys here, so it should be

```
let q_evals = q_polys.iter().map(|poly| eval_polynomial(poly.as_ref().unwrap(), x_6)).collect();
```

ebfull (changes requested):

src/poly/multiopen/verifier.rs
```diff
@@ -0,0 +1,92 @@
...
+
+        // Compute the final commitment that has to be opened
+        let mut commitment_msm = params.empty_msm();
+        commitment_msm.add_term(C::Scalar::one(), self.f_commitment);
+        for (point_index, _) in points.iter().enumerate() {
+            commitment_msm.scale(x_7);
+            commitment_msm.add_msm(&q_commitments[point_index]);
+            msm_eval *= &x_7;
+            msm_eval += &self.q_evals[point_index];
+        }
```
str4d: Do the points here line up with `q_commitments`? The pre-refactor version of this was:
```
        for (_, &point_index) in vk.cs.rotations.iter() {
            commitment_msm.scale(x_7);
            commitment_msm.add_msm(&q_commitments[point_index.0]);
            msm_eval *= &x_7;
            msm_eval += &self.q_evals[point_index.0];
        }
```
and `points` here is computed as:
```
        for (&row, &point_index) in vk.cs.rotations.iter() {
            points[point_index.0] = vk.get_domain().rotate_omega(x_3, row);
        }
```
IDK whether the indices in `vk.cs.rotations` are sequential and ordered, so it's possible that this refactor has altered the order in which the elements of `q_commitments` and `q_evals` are included in the MSM.

therealyingtong: `vk.cs.rotations` is a `BTreeMap<Rotation, PointIndex>`, so `vk.cs.rotations.iter()` does give us the entries sorted by key. https://doc.rust-lang.org/std/collections/struct.BTreeMap.html#examples-22

src/arithmetic/fields/fq.rs
```diff
@@ -62,6 +62,28 @@ impl PartialEq for Fq {
...
+        let right = other.to_bytes();
+        for (left_byte, right_byte) in left.iter().zip(right.iter()).rev() {
+            let tmp = left_byte.cmp(right_byte);
+            if let std::cmp::Ordering::Equal = tmp {
+                continue;
+            } else {
+                return tmp;
+            }
+        }
+        std::cmp::Ordering::Equal
```
therealyingtong, line 77:
```suggestion
        left.iter()
            .zip(right.iter())
            .rev()
            .find_map(|(left_byte, right_byte)| match left_byte.cmp(right_byte) {
                std::cmp::Ordering::Equal => None,
                res => Some(res),
            })
            .unwrap_or(std::cmp::Ordering::Equal)
```

src/poly/multiopen/verifier.rs
```diff
@@ -0,0 +1,246 @@
...
+        params: &'a Params<C>,
+        transcript: &mut HBase,
+        transcript_scalar: &mut HScalar,
+        queries: I,
+        msm: MSM<'a, C>,
+    ) -> Result<Guard<'a, C>, Error>
+    where
+        I: IntoIterator<Item = VerifierQuery<'a, C>> + Clone,
+    {
+        // Sample x_4 for compressing openings at the same points together
```
ebfull, line 33: The beginning of this function should have

```rust
        // Scale the MSM by a random factor to ensure that if the existing MSM
        // has is_zero() == false then this argument won't be able to interfere
        // with it to make it true, with high probability.
        msm.scale(C::Scalar::random());
```

and the existing code in `plonk/verifier.rs` should be removed because it is now no longer the PLONK implementation's responsibility to ensure these are linearly independent, and in any event usage of the multiopen API on its own would be unsafe.

src/poly/multiopen/verifier.rs
```diff
@@ -0,0 +1,246 @@
...
+        let x_4: C::Scalar = get_challenge_scalar(Challenge(transcript.squeeze().get_lower_128()));
+
+        let (commitment_map, point_sets) = construct_intermediate_sets::<'a, C, I>(queries.clone());
+
+        // Compress the commitments and expected evaluations at x_3 together.
+        // using the challenge x_4
+        let mut q_commitments: Vec<_> = vec![params.empty_msm(); point_sets.len()];
```
ebfull, line 40: I think we can do something here very similar to the `fold` stuff we did elsewhere.

src/poly/multiopen/verifier.rs
```diff
@@ -0,0 +1,246 @@
...
+        for eval in self.q_evals.iter() {
+            transcript_scalar.absorb(*eval);
+        }
+
+        let transcript_scalar_point =
+            C::Base::from_bytes(&(transcript_scalar.squeeze()).to_bytes()).unwrap();
+        transcript.absorb(transcript_scalar_point);
+
+        // lagrange_Interpolate polynomial for evaluations at each set
+        let mut r_evals = vec![C::Scalar::zero(); point_sets.len()];
```
ebfull, line 84: This implementation can look like the corresponding `fold` improvements made in the prover.

src/poly/multiopen/verifier.rs
```diff
@@ -0,0 +1,246 @@
...
+        for (points, evals) in point_sets.clone().iter().zip(q_eval_sets.clone().iter()) {
+            r_polys.push(lagrange_interpolate(&points, &evals));
+        }
+        for (r_eval, r_poly) in r_evals.iter_mut().zip(r_polys.iter()) {
+            *r_eval = eval_polynomial(r_poly, x_6);
+        }
+
+        // We can compute the expected msm_eval at x_6 using the q_evals provided
+        // by the prover and from x_5
+        let mut msm_eval = C::Scalar::zero();
```
ebfull, line 95: This can also be done with a `fold`.

src/poly/multiopen/verifier.rs
```diff
@@ -0,0 +1,246 @@
...
+            msm_eval *= &x_5;
+            msm_eval += &eval;
+        }
+
+        // Sample a challenge x_7 that we will use to collapse the openings of
+        // the various remaining polynomials at x_6 together.
+        let x_7: C::Scalar = get_challenge_scalar(Challenge(transcript.squeeze().get_lower_128()));
+
+        // Compute the final commitment that has to be opened
+        let mut commitment_msm = params.empty_msm();
```
ebfull, line 112: This can also be done with a `fold`.

src/poly/multiopen/verifier.rs
```diff
@@ -0,0 +1,246 @@
...
+        }
+
+        // Verify the opening proof
+        self.opening
+            .verify(params, msm, transcript, x_6, commitment_msm, msm_eval)
+    }
+}
+
+// For multiopen verifier: Construct intermediate representations relating commitments to sets of points by index
+fn construct_intermediate_sets<'a, C: CurveAffine, I>(
```
ebfull, line 128: It seems like you could potentially unify the logic in this function with the logic in the `prover`'s version of this function using generics, but maybe it's needlessly complicated.

ebfull (changes requested):

src/poly/multiopen/verifier.rs
therealyingtong: i do think we need to allocate `q_commitments` and `q_eval_sets` prior to accumulating their values. this is because each commitment is not being accumulated in order, but rather at an arbitrary `set_idx`.

src/arithmetic.rs
```diff
@@ -430,3 +430,44 @@ fn log2_floor(num: usize) -> u32 {
...
+                        .iter()
+                        .chain(std::iter::once(&F::zero()))
+                        .zip(std::iter::once(&F::zero()).chain(tmp.iter()))
+                        .enumerate()
+                    {
+                        product[i] = *a * (-denom * x_k) + *b * denom;
+                    }
+                    std::mem::swap(&mut tmp, &mut product);
+                }
+            }
```
ebfull, line 461: Add an assertion here that the capacity we initialized `product`/`tmp` to originally matches the actual length at the end of the loop.

src/poly/multiopen/prover.rs
```diff
@@ -0,0 +1,291 @@
...
+            transcript.absorb(transcript_scalar_point);
+
+            let x_7: C::Scalar =
+                get_challenge_scalar(Challenge(transcript.squeeze().get_lower_128()));
+
+            let (f_poly, f_blind_try) = q_polys.iter().zip(q_blinds.iter()).fold(
+                (f_poly.clone(), f_blind),
+                |(f_poly, f_blind), (poly, blind)| {
+                    (
+                        f_poly * x_7 + &poly.clone().unwrap(),
```
ebfull, line 145: Can be `f_poly * x_7 + poly.as_ref().unwrap(),`

src/arithmetic.rs
```diff
@@ -430,3 +430,46 @@ fn log2_floor(num: usize) -> u32 {
...
+                    let denom = (*x_j - x_k).invert().unwrap();
+                    product.resize(tmp.len() + 1, F::zero());
+                    for (i, (a, b)) in tmp
+                        .iter()
+                        .chain(std::iter::once(&F::zero()))
+                        .zip(std::iter::once(&F::zero()).chain(tmp.iter()))
+                        .enumerate()
+                    {
+                        product[i] = *a * (-denom * x_k) + *b * denom;
+                    }
```
ebfull, line 458: Instead of enumerating, just add a zip with `product.iter_mut()`.

ebfull (changes requested):

ebfull (changes requested):

src/poly/multiopen/verifier.rs
```diff
@@ -0,0 +1,154 @@
...
+            );
+
+        // Sample a challenge x_7 that we will use to collapse the openings of
+        // the various remaining polynomials at x_6 together.
+        let x_7: C::Scalar = get_challenge_scalar(Challenge(transcript.squeeze().get_lower_128()));
+
+        // Compute the final commitment that has to be opened
+        let mut commitment_msm = params.empty_msm();
+        commitment_msm.add_term(C::Scalar::one(), self.f_commitment);
+        let (commitment_msm, msm_eval) = q_commitments.iter().zip(self.q_evals.iter()).fold(
```
ebfull, line 117: We can do `q_commitments.into_iter()` here I think.

src/poly/multiopen/prover.rs
```diff
@@ -0,0 +1,191 @@
...
+        // x_4 challenge.
+        let mut q_polys: Vec<Option<Polynomial<C::Scalar, Coeff>>> = vec![None; point_sets.len()];
+        let mut q_blinds = vec![Blind(C::Scalar::zero()); point_sets.len()];
+
+        // A vec of vecs of evals. The outer vec corresponds to the point set,
+        // while the inner vec corresponds to the points in a particular set.
+        let mut q_eval_sets: Vec<Vec<_>> = vec![Vec::new(); point_sets.len()];
+        for (set_idx, point_set) in point_sets.iter().enumerate() {
+            q_eval_sets[set_idx] = vec![C::Scalar::zero(); point_set.len()];
+        }
```
ebfull, line 49: Simpler would be to `Vec::with_capacity(point_sets.len())` here and not do an enumeration in the loop below; just push to the vector.

src/poly/multiopen/prover.rs
```diff
@@ -0,0 +1,191 @@
...
+                );
+            }
+        }
+
+        let f_poly = point_sets
+            .iter()
+            .zip(q_eval_sets.iter())
+            .zip(q_polys.iter())
+            .fold(None, |f_poly, ((points, evals), poly)| {
+                let mut poly = poly.clone()?.values;
```
ebfull, line 86: Let's not `.clone()?` here because it's a bug in the multiopen abstraction for, at this point in the prover, any element in `q_polys` being `None`.

ebfull (changes requested):

src/poly/multiopen/prover.rs
therealyingtong: the difficulty i'm facing is that `accumulate` has to index into `q_eval_sets` at a certain `set_idx` for each new polynomial, and it's not guaranteed that these `set_idx`s will be in ascending order.

ebfull: In this particular code, we're just initializing the vectors in ascending order.

therealyingtong: oh! i misread the as a larger refactor than actually suggested.

ebfull (approved):

Issue #34, opened 2020-09-25
Check probability that Fp/Fq elements can be reinterpreting


ebfull:
Currently, we take field elements from a transcript instantiated over the "other curve" and interpret them as field elements in the "current curve." This is fine for one curve's scalar field but not the other one, because it's smaller. However, it's really close in size so perhaps if the output of the hash is random, there's no way for it to overflow. (Perhaps even when it overflows it doesn't matter?)

daira:
This is fine because the statistical distance between the [0, p) and [0, q) distributions is less than 2<sup>-128</sup>. Actually it is about 2<sup>-164.48</sup> (i.e. (p-q)/p or (p-q)/q, which are roughly the same).

PR #43, opened 2020-10-14
Make transcript generic over curve point

### LINKED ISSUE #19: Rename best_multiexp to best_msm

### LINKED ISSUE #40: Change the Hasher to be generic over curve points
Instead of setting up two transcripts individually, merge the transcripts together so that the implementation can choose how to optimize for the particular setting.

Also, probably rename Hasher to Transcript.

therealyingtong:
Closes #40.

### DIFF
--- a/src/plonk/prover.rs
+++ b/src/plonk/prover.rs
@@ -105,7 +105,9 @@ impl<C: CurveAffine> Proof<C> {
         drop(aux_commitments_projective);
 
         for commitment in &aux_commitments {
-            transcript.absorb_point(commitment).ok();
+            transcript
+                .absorb_point(commitment)
+                .map_err(|_| Error::TranscriptError)?;
         }
 
         let aux_polys: Vec<_> = aux
@@ -143,7 +145,9 @@ impl<C: CurveAffine> Proof<C> {
         drop(advice_commitments_projective);
 
         for commitment in &advice_commitments {
-            transcript.absorb_point(commitment).ok();
+            transcript
+                .absorb_point(commitment)
+                .map_err(|_| Error::TranscriptError)?;
         }
 
         let advice_polys: Vec<_> = witness
@@ -277,7 +281,9 @@ impl<C: CurveAffine> Proof<C> {
 
         // Hash each permutation product commitment
         for c in &permutation_product_commitments {
-            transcript.absorb_point(c).ok();
+            transcript
+                .absorb_point(c)
+                .map_err(|_| Error::TranscriptError)?;
         }
 
         // Obtain challenge for keeping all separate gates linearly independent
@@ -385,7 +391,9 @@ impl<C: CurveAffine> Proof<C> {
 
         // Hash each h(X) piece
         for c in h_commitments.iter() {
-            transcript.absorb_point(c).ok();
+            transcript
+                .absorb_point(c)
+                .map_err(|_| Error::TranscriptError)?;
         }
 
         let x_3: C::Scalar = get_challenge_scalar(Challenge(transcript.squeeze().get_lower_128()));
--- a/src/plonk/prover.rs
+++ b/src/plonk/prover.rs
@@ -1,6 +1,6 @@
 use super::{
     circuit::{Advice, Assignment, Circuit, Column, ConstraintSystem, Fixed},
-    hash_point, Error, Proof, ProvingKey,
+    Error, Proof, ProvingKey,
 };
 use crate::arithmetic::{
     eval_polynomial, get_challenge_scalar, parallelize, BatchInvert, Challenge, Curve, CurveAffine,
@@ -11,7 +11,7 @@ use crate::poly::{
     multiopen::{self, ProverQuery},
     LagrangeCoeff, Polynomial, Rotation,
 };
-use crate::transcript::Hasher;
+use crate::transcript::{Hasher, Transcript};
 
 impl<C: CurveAffine> Proof<C> {
     /// This creates a proof for the provided `circuit` when given the public
@@ -92,7 +92,7 @@ impl<C: CurveAffine> Proof<C> {
         let witness = witness;
 
         // Create a transcript for obtaining Fiat-Shamir challenges.
-        let mut transcript = HBase::init(C::Base::one());
+        let mut transcript = Transcript::<C, HBase, HScalar>::new();
 
         // Compute commitments to aux column polynomials
         let aux_commitments_projective: Vec<_> = aux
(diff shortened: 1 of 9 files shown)


src/transcript.rs
```diff
@@ -43,3 +44,85 @@ impl<F: Field> Hasher<F> for DummyHash<F> {
...
+            base_hasher,
+            scalar_hasher,
+            scalar_needs_squeezing: false,
+            _marker: PhantomData,
+        }
+    }
+
+    /// Initialise a new transcript with some given base_hasher and
+    /// scalar_hasher
+    pub fn init_with_hashers(base_hasher: &HBase, scalar_hasher: &HScalar) -> Self {
```
str4d, line 84: AFAICT this is only used by a test. This isn't an API we should be exposing without some thought as to what guarantees we require from it, so let's keep it internal for now:
```suggestion
    #[cfg(test)]
    pub(crate) fn init_with_hashers(base_hasher: &HBase, scalar_hasher: &HScalar) -> Self {
```

src/transcript.rs
```diff
@@ -43,3 +44,85 @@ impl<F: Field> Hasher<F> for DummyHash<F> {
...
+        self.scalar_hasher.absorb(scalar);
+        self.scalar_needs_squeezing = true;
+    }
+
+    /// Squeeze the transcript to obtain a C::Base value.
+    pub fn squeeze(&mut self) -> C::Base {
+        if self.scalar_needs_squeezing {
+            let transcript_scalar_point =
+                C::Base::from_bytes(&(self.scalar_hasher.squeeze()).to_bytes()).unwrap();
+            self.base_hasher.absorb(transcript_scalar_point);
```
str4d, line 122: This correctly abstracts the existing mechanism for connecting the scalar and base field elements into a single transcript. Please open an issue for fixing this to have the properties we require.

str4d (changes requested):

src/transcript.rs
therealyingtong: This could be related to https://github.com/zcash/halo2/issues/34.

src/poly/multiopen/prover.rs
```diff
@@ -110,8 +108,9 @@ impl<C: CurveAffine> Proof<C> {
 
         let (opening, q_evals) = loop {
             let mut transcript = transcript.clone();
-            let mut transcript_scalar = transcript_scalar.clone();
-            hash_point(&mut transcript, &f_commitment).unwrap();
+            transcript
+                .absorb_point(&f_commitment)
+                .map_err(|_| Error::SamplingError)?;
```
str4d, line 113: Relative to the previous behaviour, we now absorb `f_commitment` before we squeeze the scalar hasher to absorb the scalars from the PLONK prover into the base hasher. I'm not sure if the reordering matters (i.e. whether we should squeeze the scalar hasher into the base hasher if it needs squeezing, when we call `absorb_point` as well as when we call `squeeze`).

src/plonk/prover.rs
```diff
@@ -451,13 +455,9 @@ impl<C: CurveAffine> Proof<C> {
             .chain(permutation_product_inv_evals.iter())
             .chain(permutation_evals.iter().flat_map(|evals| evals.iter()))
         {
-            transcript_scalar.absorb(*eval);
+            transcript.absorb_scalar(*eval);
         }
 
-        let transcript_scalar_point =
-            C::Base::from_bytes(&(transcript_scalar.squeeze()).to_bytes()).unwrap();
-        transcript.absorb(transcript_scalar_point);
```
str4d, line 459: This is the squeeze I'm referring to above, which is now deferred until after `f_commitment` is absorbed. If we conditionally squeezed inside `Transcript::absorb_point`, we'd retain the current behaviour for this specific case.

str4d: (And similarly in `Transcript::absorb_base`)

src/poly/commitment/verifier.rs
```diff
@@ -148,15 +149,12 @@ impl<C: CurveAffine> Proof<C> {
             challenges_sq_packed.push(Challenge(challenge_sq_packed));
         }
 
-        let delta = self.delta.get_xy();
-        if bool::from(delta.is_none()) {
-            return Err(Error::OpeningError);
-        }
-        let delta = delta.unwrap();
+        let delta = self.delta;
```
str4d, line 152:
```suggestion
```

src/poly/commitment/verifier.rs
```diff
@@ -148,15 +149,12 @@ impl<C: CurveAffine> Proof<C> {
...
 
         // Feed delta into the transcript
-        transcript.absorb(delta.0);
-        transcript.absorb(delta.1);
+        transcript
+            .absorb_point(&delta)
```
str4d, line 156:
```suggestion
            .absorb_point(&self.delta)
```

src/transcript.rs
```diff
@@ -43,3 +44,86 @@ impl<F: Field> Hasher<F> for DummyHash<F> {
...
+        };
+        let tmp = tmp.unwrap();
+        self.base_hasher.absorb(tmp.0);
+        self.base_hasher.absorb(tmp.1);
+        Ok(())
+    }
+
+    /// Absorb a base into the base_hasher
+    pub fn absorb_base(&mut self, base: C::Base) {
+        self.base_hasher.absorb(base);
```
str4d, line 109: Is there a possibility that absorbing two bases in a row could be confused with absorbing a point (in terms the transcript malleability)? Where do we use `absorb_base`, and do we need to differentiate between absorbing individual bases and points (with e.g. some kind of domain separator)?

cc @ebfull @daira

str4d (changes requested):

src/transcript.rs
str4d: Non-blocking; open an issue if we don't answer this here.

str4d (approved):
ACK modulo my comment above either being answered, or having an issue opened.

src/plonk/prover.rs
daira: Moving an absorption earlier is safe, i.e. always preserves knowledge soundness (because all squeezed challenges will depend at least on everything they depended on before). Moving a squeeze earlier is not safe in general.

str4d:
Rebased and squashed some of the smaller commits.

str4d (approved):
re-ACK

src/transcript.rs
str4d: Opened https://github.com/zcash/halo2/issues/69.

PR #52, opened 2020-10-30
Remove redundancy in type parameters.

daira:


therealyingtong:
The `Circuit` in the lookup argument implementation has been modified to be unaware of the curve (as before).

PR #53, opened 2020-11-04
Lookup argument implementation

### LINKED ISSUE #44: Implement lookup argument

therealyingtong:
Closes #44

### DIFF
--- a/src/plonk.rs
+++ b/src/plonk.rs
@@ -121,7 +121,7 @@ fn test_proving() {
     use crate::poly::commitment::{Blind, Params};
     use crate::transcript::DummyHash;
     use crate::tweedle::{EqAffine, Fp, Fq};
-    use circuit::{Advice, Any, Column, Fixed};
+    use circuit::{Advice, Column, Fixed};
     use std::marker::PhantomData;
     const K: u32 = 5;
 
@@ -354,11 +354,8 @@ fn test_proving() {
              *
              * ]
              */
-            meta.lookup(&[Column::<Any>::from(a)], &[Column::<Any>::from(sl)]);
-            meta.lookup(
-                &[Column::<Any>::from(a), Column::<Any>::from(b)],
-                &[Column::<Any>::from(sl), Column::<Any>::from(sl2)],
-            );
+            meta.lookup(&[a.into()], &[sl.into()]);
+            meta.lookup(&[a.into(), b.into()], &[sl.into(), sl2.into()]);
 
             meta.create_gate(|meta| {
                 let d = meta.query_advice(d, 1);
--- a/src/plonk.rs
+++ b/src/plonk.rs
@@ -92,6 +92,10 @@ impl<C: CurveAffine> VerifyingKey<C> {
 }
 
 #[derive(Clone, Copy, Debug)]
+struct Theta;
+type ChallengeTheta<F> = ChallengeScalar<F, Theta>;
+
+#[derive(Clone, Copy, Debug)]
 struct Beta;
 type ChallengeBeta<F> = ChallengeScalar<F, Beta>;
 
--- a/src/plonk.rs
+++ b/src/plonk.rs
@@ -121,7 +121,7 @@ fn test_proving() {
     use crate::poly::commitment::{Blind, Params};
     use crate::transcript::DummyHash;
     use crate::tweedle::{EqAffine, Fp, Fq};
-    use circuit::{Advice, Column, Fixed};
+    use circuit::{Advice, Any, Column, Fixed};
     use std::marker::PhantomData;
     const K: u32 = 5;
 
@@ -144,6 +144,8 @@ fn test_proving() {
         sc: Column<Fixed>,
         sm: Column<Fixed>,
         sp: Column<Fixed>,
+        sl: Column<Fixed>,
+        sl2: Column<Fixed>,
 
         perm: usize,
         perm2: usize,
@@ -160,10 +162,12 @@ fn test_proving() {
         fn public_input<F>(&mut self, f: F) -> Result<Variable, Error>
         where
             F: FnOnce() -> Result<FF, Error>;
+        fn lookup_table(&mut self, values: &[Vec<FF>]) -> Result<(), Error>;
     }
 
     struct MyCircuit<F: FieldExt> {
         a: Option<F>,
+        lookup_tables: Vec<Vec<F>>,
     }
 
     struct StandardPLONK<'a, F: FieldExt, CS: Assignment<F> + 'a> {
@@ -297,6 +301,18 @@ fn test_proving() {
 
             Ok(Variable(self.config.a, index))
         }
+        fn lookup_table(&mut self, values: &[Vec<FF>]) -> Result<(), Error> {
+            for (&value_0, &value_1) in values[0].iter().zip(values[1].iter()) {
+                let index = self.current_gate;
+
+                self.current_gate += 1;
+                self.cs
+                    .assign_fixed(self.config.sl, index, || Ok(value_0))?;
+                self.cs
+                    .assign_fixed(self.config.sl2, index, || Ok(value_1))?;
+            }
+            Ok(())
+        }
     }
 
     impl<F: FieldExt> Circuit<F> for MyCircuit<F> {
@@ -319,6 +335,30 @@ fn test_proving() {
             let sb = meta.fixed_column();
             let sc = meta.fixed_column();
             let sp = meta.fixed_column();
+            let sl = meta.fixed_column();
+            let sl2 = meta.fixed_column();
+
+            /*
+             *    A    B        ...  sl   sl2
+             * [
+             *   aux   0        ...  0    0
+             *   a     a        ...  0    0
+             *   a     a^2      ...  0    0
+             *   a     a        ...  0    0
+             *   a     a^2      ...  0    0
+             *   ...   ...      ...  ...  ...
+             *   ...   ...      ...  aux  0
+             *   ...   ...      ...  a    a
+             *   ...   ...      ...  a    a^2
+             *   ...   ...      ...  0    0
+             *
+             * ]
+             */
+            meta.lookup(&[Column::<Any>::from(a)], &[Column::<Any>::from(sl)]);
+            meta.lookup(
+                &[Column::<Any>::from(a), Column::<Any>::from(b)],
+                &[Column::<Any>::from(sl), Column::<Any>::from(sl2)],
+            );
 
             meta.create_gate(|meta| {
                 let d = meta.query_advice(d, 1);
@@ -355,6 +395,8 @@ fn test_proving() {
                 sc,
                 sm,
                 sp,
+                sl,
+                sl2,
                 perm,
                 perm2,
             }
@@ -391,22 +433,33 @@ fn test_proving() {
                 cs.copy(b1, c0)?;
             }
 
+            cs.lookup_table(&self.lookup_tables)?;
+
             Ok(())
         }
     }
 
-    let circuit: MyCircuit<Fp> = MyCircuit {
-        a: Some(Fp::rand()),
+    let a = Fp::rand();
+    let a_squared = a * &a;
+    let aux = Fp::one() + Fp::one();
+    let lookup_table = vec![aux, a, a, Fp::zero()];
+    let lookup_table_2 = vec![Fp::zero(), a, a_squared, Fp::zero()];
+
+    let empty_circuit: MyCircuit<Fp> = MyCircuit {
+        a: None,
+        lookup_tables: vec![lookup_table.clone(), lookup_table_2.clone()],
     };
 
-    let empty_circuit: MyCircuit<Fp> = MyCircuit { a: None };
+    let circuit: MyCircuit<Fp> = MyCircuit {
+        a: Some(a),
+        lookup_tables: vec![lookup_table, lookup_table_2],
+    };
 
     // Initialize the proving key
     let pk = keygen(&params, &empty_circuit).expect("keygen should not fail");
 
     let mut pubinputs = pk.get_vk().get_domain().empty_lagrange();
-    pubinputs[0] = Fp::one();
-    pubinputs[0] += Fp::one();
+    pubinputs[0] = aux;
     let pubinput = params
         .commit_lagrange(&pubinputs, Blind::default())
         .to_affine();
--- a/src/plonk.rs
+++ b/src/plonk.rs
@@ -6,7 +6,9 @@
 //! [plonk]: https://eprint.iacr.org/2019/953
 
 use crate::arithmetic::CurveAffine;
-use crate::poly::{multiopen, Coeff, EvaluationDomain, ExtendedLagrangeCoeff, LagrangeCoeff, Polynomial};
+use crate::poly::{
+    multiopen, Coeff, EvaluationDomain, ExtendedLagrangeCoeff, LagrangeCoeff, Polynomial,
+};
 use crate::transcript::ChallengeScalar;
 
 mod circuit;
--- a/src/plonk.rs
+++ b/src/plonk.rs
@@ -53,6 +53,7 @@ pub struct Proof<C: CurveAffine> {
     advice_commitments: Vec<C>,
     h_commitments: Vec<C>,
     permutations: Option<permutation::Proof<C>>,
+    lookups: Vec<lookup::Proof<C>>,
     advice_evals: Vec<C::Scalar>,
     aux_evals: Vec<C::Scalar>,
     fixed_evals: Vec<C::Scalar>,
--- a/src/plonk.rs
+++ b/src/plonk.rs
@@ -11,6 +11,7 @@ use crate::transcript::ChallengeScalar;
 
 mod circuit;
 mod keygen;
+mod lookup;
 mod permutation;
 mod prover;
 mod verifier;
--- a/src/plonk.rs
+++ b/src/plonk.rs
@@ -6,7 +6,7 @@
 //! [plonk]: https://eprint.iacr.org/2019/953
 
 use crate::arithmetic::CurveAffine;
-use crate::poly::{multiopen, Coeff, EvaluationDomain, ExtendedLagrangeCoeff, Polynomial};
+use crate::poly::{multiopen, Coeff, EvaluationDomain, ExtendedLagrangeCoeff, LagrangeCoeff, Polynomial};
 use crate::transcript::ChallengeScalar;
 
 mod circuit;
@@ -37,6 +37,7 @@ pub struct ProvingKey<C: CurveAffine> {
     vk: VerifyingKey<C>,
     // TODO: get rid of this?
     l0: Polynomial<C::Scalar, ExtendedLagrangeCoeff>,
+    fixed_values: Vec<Polynomial<C::Scalar, LagrangeCoeff>>,
     fixed_polys: Vec<Polynomial<C::Scalar, Coeff>>,
     fixed_cosets: Vec<Polynomial<C::Scalar, ExtendedLagrangeCoeff>>,
     permutations: Vec<permutation::ProvingKey<C>>,
--- a/src/plonk/circuit.rs
+++ b/src/plonk/circuit.rs
@@ -364,6 +364,9 @@ impl<F: Field> ConstraintSystem<F> {
         input_columns: &[Column<Any>],
         table_columns: &[Column<Any>],
     ) -> usize {
+        // The function will panic if the number of input columns and table columns are not the same.
+        assert_eq!(input_columns.len(), table_columns.len());
+
         let index = self.lookups.len();
         if self.lookups.is_empty() {
             let at = Rotation(-1);
--- a/src/plonk/circuit.rs
+++ b/src/plonk/circuit.rs
@@ -359,12 +359,13 @@ impl<F: Field> ConstraintSystem<F> {
     }
 
     /// Add a lookup argument for some input columns and table columns.
+    /// The function will panic if the number of input columns and table
+    /// columns are not the same.
     pub fn lookup(
         &mut self,
         input_columns: &[Column<Any>],
         table_columns: &[Column<Any>],
     ) -> usize {
-        // The function will panic if the number of input columns and table columns are not the same.
         assert_eq!(input_columns.len(), table_columns.len());
 
         let index = self.lookups.len();
--- a/src/plonk/circuit.rs
+++ b/src/plonk/circuit.rs
@@ -358,6 +358,31 @@ impl<F: Field> ConstraintSystem<F> {
         index
     }
 
+    /// Add a lookup argument for some input columns and table columns.
+    pub fn lookup(
+        &mut self,
+        input_columns: &[Column<Any>],
+        table_columns: &[Column<Any>],
+    ) -> usize {
+        let index = self.lookups.len();
+        if self.lookups.is_empty() {
+            let at = Rotation(-1);
+            let len = self.rotations.len();
+            self.rotations.entry(at).or_insert(PointIndex(len));
+        }
+
+        for input in input_columns {
+            self.query_any_index(*input, 0);
+        }
+        for table in table_columns {
+            self.query_any_index(*table, 0);
+        }
+        self.lookups
+            .push(lookup::Argument::new(input_columns, table_columns));
+
+        index
+    }
+
     fn query_fixed_index(&mut self, column: Column<Fixed>, at: i32) -> usize {
         let at = Rotation(at);
         {
--- a/src/plonk/circuit.rs
+++ b/src/plonk/circuit.rs
@@ -4,7 +4,7 @@ use ff::Field;
 use std::collections::BTreeMap;
 use std::convert::TryFrom;
 
-use super::{permutation, Error};
+use super::{lookup, permutation, Error};
 use crate::poly::Rotation;
 
 /// A column type
@@ -313,6 +313,10 @@ pub struct ConstraintSystem<F> {
     // Vector of permutation arguments, where each corresponds to a sequence of columns
     // that are involved in a permutation argument.
     pub(crate) permutations: Vec<permutation::Argument>,
+
+    // Vector of lookup arguments, where each corresponds to a sequence of
+    // input columns and a sequence of table columns involved in the lookup.
+    pub(crate) lookups: Vec<lookup::Argument>,
 }
 
 impl<F: Field> Default for ConstraintSystem<F> {
@@ -330,6 +334,7 @@ impl<F: Field> Default for ConstraintSystem<F> {
             aux_queries: Vec::new(),
             rotations,
             permutations: Vec::new(),
+            lookups: Vec::new(),
         }
     }
 }
--- a/src/plonk/circuit.rs
+++ b/src/plonk/circuit.rs
@@ -344,9 +344,7 @@ impl<F: Field> ConstraintSystem<F> {
     pub fn permutation(&mut self, columns: &[Column<Advice>]) -> usize {
         let index = self.permutations.len();
         if self.permutations.is_empty() {
-            let at = Rotation(-1);
-            let len = self.rotations.len();
-            self.rotations.entry(at).or_insert(PointIndex(len));
+            self.add_rotation(Rotation(-1));
         }
 
         for column in columns {
@@ -370,9 +368,7 @@ impl<F: Field> ConstraintSystem<F> {
 
         let index = self.lookups.len();
         if self.lookups.is_empty() {
-            let at = Rotation(-1);
-            let len = self.rotations.len();
-            self.rotations.entry(at).or_insert(PointIndex(len));
+            self.add_rotation(Rotation(-1));
         }
 
         for input in input_columns {
@@ -389,10 +385,7 @@ impl<F: Field> ConstraintSystem<F> {
 
     fn query_fixed_index(&mut self, column: Column<Fixed>, at: i32) -> usize {
         let at = Rotation(at);
-        {
-            let len = self.rotations.len();
-            self.rotations.entry(at).or_insert(PointIndex(len));
-        }
+        self.add_rotation(at);
 
         // Return existing query, if it exists
         for (index, fixed_query) in self.fixed_queries.iter().enumerate() {
@@ -415,10 +408,7 @@ impl<F: Field> ConstraintSystem<F> {
 
     pub(crate) fn query_advice_index(&mut self, column: Column<Advice>, at: i32) -> usize {
         let at = Rotation(at);
-        {
-            let len = self.rotations.len();
-            self.rotations.entry(at).or_insert(PointIndex(len));
-        }
+        self.add_rotation(at);
 
         // Return existing query, if it exists
         for (index, advice_query) in self.advice_queries.iter().enumerate() {
@@ -441,10 +431,7 @@ impl<F: Field> ConstraintSystem<F> {
 
     fn query_aux_index(&mut self, column: Column<Aux>, at: i32) -> usize {
         let at = Rotation(at);
-        {
-            let len = self.rotations.len();
-            self.rotations.entry(at).or_insert(PointIndex(len));
-        }
+        self.add_rotation(at);
 
         // Return existing query, if it exists
         for (index, aux_query) in self.aux_queries.iter().enumerate() {
@@ -567,5 +554,10 @@ impl<F: Field> ConstraintSystem<F> {
         };
         self.num_aux_columns += 1;
         tmp
+    }
+
+    fn add_rotation(&mut self, at: Rotation) {
+        let len = self.rotations.len();
+        self.rotations.entry(at).or_insert(PointIndex(len));
     }
 }
--- a/src/plonk/circuit.rs
+++ b/src/plonk/circuit.rs
@@ -432,13 +432,11 @@ impl<F: Field> ConstraintSystem<F> {
     }
 
     fn query_any_index(&mut self, column: Column<Any>, at: i32) -> usize {
-        let index = match column.column_type() {
+        match column.column_type() {
             Any::Advice => self.query_advice_index(Column::<Advice>::try_from(column).unwrap(), at),
             Any::Fixed => self.query_fixed_index(Column::<Fixed>::try_from(column).unwrap(), at),
             Any::Aux => self.query_aux_index(Column::<Aux>::try_from(column).unwrap(), at),
-        };
-
-        index
+        }
     }
 
     /// Query an Any column at a relative position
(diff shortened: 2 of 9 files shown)


src/plonk/lookup/prover.rs
```diff
@@ -0,0 +1,536 @@
...
+            parallelize(&mut first_lookup_poly, |first_lookup_poly, start| {
+                for (((first_lookup_poly, input), table), l0) in first_lookup_poly
+                    .iter_mut()
+                    .zip(permuted.permuted_input_coset[start..].iter())
+                    .zip(permuted.permuted_table_coset[start..].iter())
+                    .zip(pk.l0[start..].iter())
+                {
+                    *first_lookup_poly += &(*l0 * &(*input - table));
+                }
+            });
```
daira, line 484: We don't need `parallelize` here, the operations in `poly.rs` already do that.

daira: Similarly in `plonk/prover.rs`.

daira (approved):
utACK

src/plonk/lookup/prover.rs
```diff
@@ -0,0 +1,539 @@
...
+            // If input value is repeated
+            } else {
+                repeated_input_rows.push(row);
+            }
+        }
+
+        // Populate permuted table at unfilled rows with leftover table elements
+        for (coeff, count) in leftover_table_map.iter() {
+            for _ in 0..*count {
+                permuted_table_coeffs[repeated_input_rows.remove(0) as usize] = *coeff;
```
daira, line 166:
```suggestion
                permuted_table_coeffs[repeated_input_rows.pop() as usize] = *coeff;
```
[remove(0) is O(length)](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.remove), which would make the overall complexity quadratic. Since the order doesn't matter, we can use pop() which is O(1).

src/plonk/lookup/prover.rs
```diff
@@ -0,0 +1,539 @@
...
+    ) -> Product<C> {
+        let permuted = self.permuted.clone().unwrap();
+        let unpermuted_input_values: Vec<Polynomial<C::Scalar, LagrangeCoeff>> = self
+            .lookup
+            .input_columns
+            .iter()
+            .map(|&input| match input.column_type {
+                Any::Advice => advice_values[input.index].clone(),
+                Any::Fixed => fixed_values[input.index].clone(),
+                _ => unreachable!(),
```
daira, line 193: I don't think this is unreachable, since `ConstraintSystem::lookup` allows `Column<Any>` for the input and table columns. I can't immediately think of a use-case for an aux column in a lookup, but I also can't think of a reason to prohibit it.

src/plonk/lookup/prover.rs
```diff
@@ -0,0 +1,539 @@
...
+                    .vk
+                    .cs
+                    .get_advice_query_index(Column::<Advice>::from(input), 0)]
+                .clone(),
+                Any::Fixed => fixed_cosets[pk
+                    .vk
+                    .cs
+                    .get_fixed_query_index(Column::<Fixed>::from(input), 0)]
+                .clone(),
+                _ => unreachable!(),
```
daira, line 382: Don't forget to change this as well (and below) if aux columns are allowed in lookups.

src/plonk/lookup/prover.rs
```diff
@@ -0,0 +1,539 @@
...
+                    *left *= &(*permuted_input + &beta);
+                    *left *= &(*permuted_table + &gamma);
+                }
+            });
+
+            //  z'(\omega^{-1} X) (a_1(X) + \theta a_2(X) + ... + \beta) (s_1(X) + \theta s_2(X) + ... + \gamma)
+            let mut right = product.product_inv_coset.clone();
+            let mut input_terms = pk.vk.domain.empty_extended();
+
+            // Compress the unpermuted Input columns
```
daira, line 443:
```suggestion
            // Compress the unpermuted input columns
```

src/plonk/lookup/prover.rs
```diff
@@ -0,0 +1,539 @@
...
+                parallelize(&mut input_terms, |input_term, start| {
+                    for (input_term, input) in input_term.iter_mut().zip(input[start..].iter()) {
+                        *input_term *= &theta;
+                        *input_term += input;
+                    }
+                });
+            }
+
+            let mut table_terms = pk.vk.domain.empty_extended();
+            // Compress the unpermuted Table columns
```
daira, line 455:
```suggestion
            // Compress the unpermuted table columns
```

daira (approved):
ACK modulo comments.

src/plonk/lookup/prover.rs
```diff
@@ -0,0 +1,539 @@
...
+                repeated_input_rows.push(row);
+            }
+        }
+
+        // Populate permuted table at unfilled rows with leftover table elements
+        for (coeff, count) in leftover_table_map.iter() {
+            for _ in 0..*count {
+                permuted_table_coeffs[repeated_input_rows.remove(0) as usize] = *coeff;
+            }
+        }
```
daira, line 168:
```suggestion
        }
        assert!(permuted_table_coeffs.is_empty());
```

src/plonk/lookup/prover.rs
```diff
@@ -0,0 +1,547 @@
...
+            // If this is the first occurence of `input_value` in the input column
+            if row == 0 || input_value != input_coeffs[row - 1] {
+                permuted_table_coeffs[row] = input_value;
+                // Remove one instance of input_value from leftover_table_map
+                if let Some(count) = leftover_table_map.get_mut(&input_value) {
+                    assert!(*count > 0);
+                    *count -= 1;
+                } else {
+                    // Panic if input_value not found
+                    panic!("Input value not found in table.");
```
daira, line 156: I think we should return a `ConstraintSystemFailure` error here rather than panicking. I know that means refactoring a bunch of things to return `Result`s, sorry. But it's more consistent with how we handle constraint errors elsewhere, and it's an error made by the caller rather than an unreachable condition in this implementation (unlike the assertions and the unwrap).

daira: Hmm, `ConstraintSystemFailure` [only seems to be used in the verifier](https://github.com/zcash/halo2/search?q=ConstraintSystemFailure). That's not what I expected based on its documentation. In any case, I think there should be an error for the situation where a constraint violation is detected during proving.

src/plonk.rs
```diff
@@ -38,6 +39,7 @@ pub struct ProvingKey<C: CurveAffine> {
     vk: VerifyingKey<C>,
     // TODO: get rid of this?
     l0: Polynomial<C::Scalar, ExtendedLagrangeCoeff>,
+    fixed_values: Vec<Polynomial<C::Scalar, LagrangeCoeff>>,
```
ebfull, line 42: Let's make an issue for removing entries in this vector that we don't actually use because they aren't involved in a lookup argument.

src/plonk/lookup/prover.rs
```diff
@@ -0,0 +1,547 @@
...
+    pub fn new(lookup: &Lookup) -> Self {
+        assert_eq!(lookup.input_columns.len(), lookup.table_columns.len());
+        LookupData {
+            lookup: lookup.clone(),
+            permuted: None,
+            product: None,
+        }
+    }
+
+    pub fn construct_permuted(
```
ebfull, line 29: Add documentation for this method so that it's clear what it's doing.

src/plonk/lookup/prover.rs
```diff
@@ -0,0 +1,547 @@
...
+            permuted_table_coset,
+            permuted_table_blind,
+            permuted_table_commitment,
+        };
+
+        self.permuted = Some(permuted.clone());
+        permuted
+    }
+
+    fn permute_column_pair(
```
ebfull, line 121: Add documentation.

ebfull (changes requested):

src/plonk/lookup/prover.rs
```diff
@@ -0,0 +1,536 @@
...
+                .fold(C::Scalar::zero(), |acc, table| {
+                    acc * &theta + &table.get_values()[i]
+                });
+
+            input_term += &beta;
+            table_term += &gamma;
+            right *= &(input_term * &table_term);
+
+            assert_eq!(left, right);
+        }
```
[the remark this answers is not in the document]
therealyingtong: This test only checks intermediate representations within the `construct_product()` method, so it may be more useful for debugging than as a unit test. I've commented it out with a remark to that effect.

src/plonk.rs
therealyingtong: Opened https://github.com/zcash/halo2/issues/57.

src/plonk/lookup/prover.rs
```diff
@@ -0,0 +1,536 @@
...
+        // and i is the ith row of the wire.
+
+        // Compute the evaluations of the lookup product polynomial
+        // over our domain, starting with z[0] = 1
+        let mut z = vec![C::Scalar::one()];
+        for row in 1..(params.n as usize) {
+            let mut tmp = z[row - 1];
+            tmp *= &lookup_product[row];
+            z.push(tmp);
+        }
```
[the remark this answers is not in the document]
therealyingtong: i'm not confident if this will be an optimisation, because we currently use `parallelize()` and `batch_invert()` in the loop above.

src/plonk/verifier.rs
```diff
@@ -177,6 +192,38 @@ impl<'a, C: CurveAffine> Proof<C> {
...
+                commitment: &lookup.permuted_table_commitment,
+                eval: lookup.permuted_table_eval,
+            });
+
+            // Open lookup product commitments at \omega^{-1} x_3
+            let x_3_inv = vk.domain.rotate_omega(x_3, Rotation(-1));
+            queries.push(VerifierQuery {
+                point: x_3_inv,
+                commitment: &lookup.product_commitment,
+                eval: lookup.product_inv_eval,
```
ebfull, line 223: We are missing `lookup.permuted_input_inv_eval` here, I think.

src/plonk/verifier.rs
```diff
@@ -64,7 +79,7 @@ impl<'a, C: CurveAffine> Proof<C> {
 
         // This check ensures the circuit is satisfied so long as the polynomial
         // commitments open to the correct values.
-        self.check_hx(params, vk, x_0, x_1, x_2, x_3)?;
+        self.check_hx(params, vk, beta, gamma, x_2, x_3, theta)?;
 
         // Hash together all the openings provided by the prover into a new
         // transcript on the scalar field.
```
ebfull, line 85: We need to hash the openings for the lookup argument into the transcript like we do with the permutation arguments (expand the code just below this comment).

src/plonk/prover.rs
```diff
@@ -558,6 +626,42 @@ impl<C: CurveAffine> Proof<C> {
...
+                eval: proof.permuted_table_eval,
+            });
+
+            let x_3_inv = domain.rotate_omega(x_3, Rotation(-1));
+            // Open lookup product commitments at \omega^{-1} x_3
+            instances.push(ProverQuery {
+                point: x_3_inv,
+                poly: &lookup.product.as_ref().unwrap().product_poly,
+                blind: lookup.product.clone().unwrap().product_blind,
+                eval: proof.product_inv_eval,
```
ebfull, line 661: Missing `proof.permuted_input_inv_eval`.

src/plonk/lookup/verifier.rs
```diff
@@ -0,0 +1,79 @@
...
+        // - z'(\omega^{-1} X) (a_1(X) + \theta a_2(X) + ... + \beta) (s_1(X) + \theta s_2(X) + ... + \gamma)
+        {
+            let left = self.product_eval
+                * &(self.permuted_input_eval + &beta)
+                * &(self.permuted_table_eval + &gamma);
+
+            let mut right = self.product_inv_eval;
+            let mut input_term = C::Scalar::zero();
+            for &input in lookup.input_columns.iter() {
+                let eval = advice_evals[cs.get_any_query_index(input, 0)].clone();
```
ebfull, line 39: This code is implicitly assuming that all of the "input" columns are advice columns because it's taking the index from `get_any_query_index` and using it regardless of `input`'s column type.

src/plonk/lookup/verifier.rs
```diff
@@ -0,0 +1,79 @@
...
+                input_term *= &theta;
+                input_term += &eval;
+            }
+            input_term += &beta;
+
+            let mut table_term = C::Scalar::zero();
+            for &table in lookup.table_columns.iter() {
+                let eval = fixed_evals[cs.get_any_query_index(table, 0)].clone();
```
ebfull, line 47: Similarly this is assuming "table columns" are always fixed.

ebfull (changes requested):

src/poly.rs
```diff
@@ -108,6 +108,19 @@ impl<F, B> DerefMut for Polynomial<F, B> {
 }
 
 impl<F, B> Polynomial<F, B> {
+    /// Make new polynomial
+    pub fn new(values: Vec<F>) -> Self {
```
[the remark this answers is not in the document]
ebfull: (in other words i recommend getting rid of this `new` method)

src/plonk/verifier.rs
```diff
@@ -177,6 +192,38 @@ impl<'a, C: CurveAffine> Proof<C> {
             }
         }
 
+        // Handle lookup arguments, if any exist
```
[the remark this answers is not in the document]
ebfull: I think this suggests that we should change the way we're introducing the lookup argument as a modularized protocol. Let's sketch something out together.

src/plonk/verifier.rs
```diff
@@ -235,6 +282,10 @@ impl<'a, C: CurveAffine> Proof<C> {
             return Err(Error::IncompatibleParams);
         }
 
+        if self.lookup_proofs.len() != vk.cs.lookups.len() {
+            return Err(Error::IncompatibleParams);
+        }
+
```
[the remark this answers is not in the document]
therealyingtong: but we have no vectors inside `lookup::Proof`:

```
pub struct Proof<C: CurveAffine> {
    pub product_commitment: C,
    pub product_eval: C::Scalar,
    pub product_inv_eval: C::Scalar,
    pub permuted_input_commitment: C,
    pub permuted_table_commitment: C,
    pub permuted_input_eval: C::Scalar,
    pub permuted_input_inv_eval: C::Scalar,
    pub permuted_table_eval: C::Scalar,
}
```

src/plonk/lookup/prover.rs
```diff
@@ -0,0 +1,712 @@
...
+        let permuted_table_coset = pk
+            .vk
+            .domain
+            .coeff_to_extended(permuted_table_poly.clone(), Rotation::default());
+        let permuted_table_blind = Blind(C::Scalar::rand());
+        let permuted_table_commitment = params
+            .commit_lagrange(&permuted_table_value, permuted_table_blind)
+            .to_affine();
+
+        // Hash each permuted input commitment
```
ebfull, line 156: Remove `each` because there's only one of these per argument.

src/plonk/lookup/prover.rs
```diff
@@ -0,0 +1,712 @@
...
+        transcript
+            .absorb_point(&permuted_input_commitment)
+            .map_err(|_| Error::TranscriptError)?;
+
+        // Hash each permuted table commitment
```
ebfull, line 161: Ditto.

src/plonk/lookup/prover.rs
```diff
@@ -0,0 +1,712 @@
...
+            permuted_table_commitment,
+        })
+    }
+
+    /// Given a Lookup with input columns, table columns, and the permuted
+    /// input column and permuted table column, this method constructs the
+    /// grand product polynomial over the lookup. The grand product polynomial
+    /// is used to populate the Product<C> struct. The Product<C> struct is
+    /// added to the Lookup and finally returned by the method.
+    pub(in crate::plonk) fn commit_product<
```
ebfull, line 186: I'm somewhat confused about why this method is defined here. It should be defined on `Permuted<C>` right?

src/plonk/lookup/prover.rs
```diff
@@ -0,0 +1,712 @@
...
+        multiopen::ProverQuery,
+        Coeff, EvaluationDomain, ExtendedLagrangeCoeff, LagrangeCoeff, Polynomial, Rotation,
+    },
+    transcript::{Hasher, Transcript},
+};
+use ff::Field;
+use std::{collections::BTreeMap, convert::TryFrom, iter};
+
+#[derive(Clone, Debug)]
+pub(crate) struct Permuted<C: CurveAffine> {
```
ebfull, line 19: I don't think any of these types should be `Clone` and in the permutation argument I believe it was a mistake that the `Committed` structure implemented `Clone`. (Opened #83 to deal with that and other random things I see.)

src/plonk/lookup/prover.rs
```diff
@@ -0,0 +1,712 @@
...
+        fixed_values: &[Polynomial<C::Scalar, LagrangeCoeff>],
+        aux_values: &[Polynomial<C::Scalar, LagrangeCoeff>],
+        transcript: &mut Transcript<C, HBase, HScalar>,
+    ) -> Result<Permuted<C>, Error> {
+        // Values of input columns involved in the lookup
+        let unpermuted_input_values: Vec<Polynomial<C::Scalar, LagrangeCoeff>> = self
+            .input_columns
+            .iter()
+            .map(|&input| match input.column_type() {
+                Any::Advice => advice_values[input.index()].clone(),
```
ebfull, line 99: Instead of cloning these, obtain references to them (we're not mutating them). Further, place them into `Permuted<C>` so that when you call `commit_product` you don't need to retrieve them again -- which will then allow you to define `commit_product` on `Permuted<C>` rather than on `Argument`. (This will make it so that `Permuted<C>` will need a lifetime bound added to it.)

src/plonk/lookup/prover.rs
ebfull: I see what's going on. You needed `unpermuted_input_values` and `unpermuted_table_values` again here and thus needed `Argument` to get them. Instead, as per my other comment, this function should be defined on `Permuted<C>` and the vectors `unpermuted_input_values` / `unpermuted_table_values` should be vectors of references to polynomials and placed in `Permuted<C>` so that you don't need `Argument` again.

src/plonk/lookup/prover.rs
```diff
@@ -0,0 +1,712 @@
...
+    ///   obtaining A' and S', and
+    /// - constructs Permuted<C> struct using permuted_input_value = A', and
+    ///   permuted_table_value = S'.
+    /// The Permuted<C> struct is used to update the Lookup, and is then returned.
+    pub(in crate::plonk) fn commit_permuted<
+        C: CurveAffine,
+        HBase: Hasher<C::Base>,
+        HScalar: Hasher<C::Scalar>,
+    >(
+        &self,
```
ebfull, line 84: I think you should be able to take this by value, aka just `self`, right?

src/plonk/lookup/prover.rs
```diff
@@ -0,0 +1,712 @@
...
+    /// input column and permuted table column, this method constructs the
+    /// grand product polynomial over the lookup. The grand product polynomial
+    /// is used to populate the Product<C> struct. The Product<C> struct is
+    /// added to the Lookup and finally returned by the method.
+    pub(in crate::plonk) fn commit_product<
+        C: CurveAffine,
+        HBase: Hasher<C::Base>,
+        HScalar: Hasher<C::Scalar>,
+    >(
+        &self,
```
ebfull, line 191: When this method is relocated, you should be able to make it take `self` by value.

ebfull (changes requested):

src/plonk/lookup/prover.rs
ebfull: Ah, you might not be able to do it for this one but for the others you should try to.

src/plonk/lookup/prover.rs
```diff
@@ -0,0 +1,712 @@
...
+        parallelize(&mut lookup_product, |lookup_product, start| {
+            for ((lookup_product, permuted_input_value), permuted_table_value) in lookup_product
+                .iter_mut()
+                .zip(permuted.permuted_input_value[start..].iter())
+                .zip(permuted.permuted_table_value[start..].iter())
+            {
+                *lookup_product *= &(*beta + permuted_input_value);
+                *lookup_product *= &(*gamma + permuted_table_value);
+            }
+        });
```
str4d, line 246: This would be more efficient (half the number of multiplications):
```suggestion
        let mut lookup_product = vec![C::Scalar::zero(); params.n as usize];

        // Denominator uses the permuted input column and permuted table column
        parallelize(&mut lookup_product, |lookup_product, start| {
            for ((lookup_product, permuted_input_value), permuted_table_value) in lookup_product
                .iter_mut()
                .zip(permuted.permuted_input_value[start..].iter())
                .zip(permuted.permuted_table_value[start..].iter())
            {
                *lookup_product = (*beta + permuted_input_value) * &(*gamma + permuted_table_value);
            }
        });
```

str4d (changes requested):

src/plonk/lookup/prover.rs
```diff
@@ -176,6 +176,200 @@ impl Argument {
...
+        // and i is the ith row of the column.
+
+        // Compute the evaluations of the lookup product polynomial
+        // over our domain, starting with z[0] = 1
+        let mut z = vec![C::Scalar::one()];
+        for row in 1..(params.n as usize) {
+            let mut tmp = z[row - 1];
+            tmp *= &lookup_product[row];
+            z.push(tmp);
+        }
```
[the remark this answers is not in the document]
ebfull: Yeah, `let mut z = vec![C::Scalar::one()];` initializes the vector with the first element being one, with the intention to show that the product of the numerators is the inverse of the product of the denominators. Side note, this code should probably be altered so that the z vector is initialized with the correct capacity (`params.n`) to reduce allocations.

src/plonk/lookup/prover.rs
ebfull: Nice, note that some of these optimizations also apply to the permutation argument probably so it'd be nice to mirror them over there as well.

src/plonk/lookup/prover.rs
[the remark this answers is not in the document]
str4d: Mmk, so `z` ends up containing in its last entry the product of all parts of `lookup_product` except the first? It's not clear to me where that last part is accounted for.

Re: reducing allocations, AFAIK the chain-and-collects I propose above and in other comments should also achieve that. I should dig into `collect` to check this...

src/plonk/prover.rs
```diff
@@ -192,6 +214,40 @@ impl<C: CurveAffine> Proof<C> {
...
         };
 
+        // Construct products for each lookup
+        let lookups_products = pk
+            .vk
+            .cs
+            .lookups
+            .iter()
+            .zip(lookups_permuted.iter())
+            .map(|(lookup, permuted)| {
```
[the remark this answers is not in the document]
therealyingtong: we no longer need to zip the `pk.vk.cs.lookups` in here.

src/plonk/lookup/prover.rs
```diff
@@ -176,6 +176,200 @@ impl Argument {
...
+                {
+                    *input_term *= &theta;
+                    *input_term += input_value;
+                }
+            });
+        }
+
+        // Compress unpermuted table columns
+        let mut table_term = vec![C::Scalar::zero(); params.n as usize];
+        for unpermuted_table_value in unpermuted_table_values.iter() {
```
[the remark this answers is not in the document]
therealyingtong: i've updated the documentation to match the reverse progression `\theta^{m-1} a_1(X) + ... + \theta a_{m-1}(X) + a_m(X)`, where `m` is the number of input columns = the number of table columns.

src/plonk/lookup/prover.rs
```diff
@@ -176,6 +176,200 @@ impl Argument {
...
+            .map(|&table| match table.column_type() {
+                Any::Advice => advice_values[table.index()].clone(),
+                Any::Fixed => fixed_values[table.index()].clone(),
+                Any::Aux => aux_values[table.index()].clone(),
+            })
+            .collect();
+
+        // Goal is to compute the products of fractions
+        //
+        // (a_1(\omega^i) + \theta a_2(\omega^i) + ... + beta)(s_1(\omega^i) + \theta(\omega^i) + ... + \gamma) /
```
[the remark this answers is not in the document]
therealyingtong: thanks for the catch :+1:

src/plonk/lookup/prover.rs
```diff
@@ -51,3 +64,191 @@ pub(crate) struct Evaluated<C: CurveAffine> {
...
+                *count -= 1;
+            } else {
+                // Return error if input_value not found
+                return Err(Error::ConstraintSystemFailure);
+            }
+        // If input value is repeated
+        } else {
+            repeated_input_rows.push(row);
+        }
+    }
```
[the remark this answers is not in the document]
therealyingtong: did not ignore!

str4d (approved):
utACK, with some non-blocking comments.

src/plonk/lookup/prover.rs
```diff
@@ -0,0 +1,664 @@
+use super::super::{
+    circuit::{Advice, Any, Aux, Column, Fixed},
```
therealyingtong, line 2:
```suggestion
    circuit::{Any, Column},
```

src/plonk/lookup/prover.rs
```diff
@@ -0,0 +1,664 @@
...
+    arithmetic::{eval_polynomial, parallelize, BatchInvert, Curve, CurveAffine, FieldExt},
+    poly::{
+        commitment::{Blind, Params},
+        multiopen::ProverQuery,
+        Coeff, EvaluationDomain, ExtendedLagrangeCoeff, LagrangeCoeff, Polynomial, Rotation,
+    },
+    transcript::{Hasher, Transcript},
+};
+use ff::Field;
+use std::{collections::BTreeMap, convert::TryFrom, iter};
```
therealyingtong, line 16:
```suggestion
use std::{collections::BTreeMap, iter};
```

src/plonk/lookup/verifier.rs
```diff
@@ -0,0 +1,157 @@
+use std::iter;
+
+use super::super::circuit::Any;
```
therealyingtong, line 3:
```suggestion
use super::super::circuit::{Any, Column};
```

src/plonk/lookup/prover.rs
```diff
@@ -0,0 +1,664 @@
...
+        }
+
+        let product_blind = Blind(C::Scalar::rand());
+        let product_commitment = params.commit_lagrange(&z, product_blind).to_affine();
+        let z = pk.vk.domain.lagrange_to_coeff(z);
+        let product_coset = pk
+            .vk
+            .domain
+            .coeff_to_extended(z.clone(), Rotation::default());
+        let product_inv_coset = pk.vk.domain.coeff_to_extended(z.clone(), Rotation(-1));
```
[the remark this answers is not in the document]
therealyingtong: i've opened https://github.com/zcash/halo2/issues/new. i think `transcript.absorb_point` should remain outside though because the transcript is separate from the `poly` crate.

src/plonk/lookup/prover.rs
```diff
@@ -0,0 +1,652 @@
...
+        parallelize(&mut lookup_product, |product, start| {
+            for ((product, input_term), table_term) in product
+                .iter_mut()
+                .zip(input_term[start..].iter())
+                .zip(table_term[start..].iter())
+            {
+                *product *= &(*input_term + &beta);
+                *product *= &(*table_term + &gamma);
+            }
+        });
```
ebfull, line 275: These three loops could be integrated together to avoid the extra allocations and choke points for parallelism, and also make it more clear what's happening.

```suggestion
        // Finish the computation of the entire fraction by computing the numerators
        // (\theta^{m-1} a_1(\omega^i) + \theta^{m-2} a_2(\omega^i) + ... + \theta a_{m-1}(\omega^i) + a_m(\omega^i) + \beta)
        // * (\theta^{m-1} s_1(\omega^i) + \theta^{m-2} s_2(\omega^i) + ... + \theta s_{m-1}(\omega^i) + s_m(\omega^i) + \gamma)
        parallelize(&mut lookup_product, |product, start| {
            for (i, product) in product.iter_mut().enumerate() {
                let i = i + start;

                // Compress unpermuted input columns
                let mut input_term = C::Scalar::zero();
                for unpermuted_input_column in self.unpermuted_input_columns.iter() {
                    input_term *= &theta;
                    input_term += &unpermuted_input_column[i];
                }

                // Compress unpermuted table columns
                let mut table_term = C::Scalar::zero();
                for unpermuted_table_column in self.unpermuted_table_columns.iter() {
                    table_term *= &theta;
                    table_term += &unpermuted_table_column[i];
                }

                *product *= &(input_term + &beta);
                *product *= &(table_term + &gamma);
            }
        });
```

src/plonk/lookup/prover.rs
```diff
@@ -0,0 +1,652 @@
...
+                parallelize(&mut right, |right, start| {
+                    for ((right, input_term), table_term) in right
+                        .iter_mut()
+                        .zip(input_terms[start..].iter())
+                        .zip(table_terms[start..].iter())
+                    {
+                        *right *= &(*input_term + &(*beta));
+                        *right *= &(*table_term + &(*gamma));
+                    }
+                });
```
ebfull, line 442: Similar to above, we can combine these loops.

```suggestion
                parallelize(&mut right, |right, start| {
                    for (i, right) in right.iter_mut().enumerate() {
                        let i = i + start;

                        // Compress the unpermuted input columns
                        let mut input_term = C::Scalar::zero();
                        for input in permuted.unpermuted_input_cosets.iter() {
                            input_term *= &theta;
                            input_term += &input[i];
                        }

                        // Compress the unpermuted table columns
                        let mut table_term = C::Scalar::zero();
                        for table in permuted.unpermuted_table_cosets.iter() {
                            table_term *= &theta;
                            table_term += &table[i];
                        }

                        // Add \beta and \gamma offsets
                        *right *= &(input_term + &beta);
                        *right *= &(table_term + &gamma);
                    }
                });
```

ebfull (changes requested):
Looks like we're almost done here! Good job!

src/plonk.rs
```diff
@@ -310,6 +335,30 @@ fn test_proving() {
...
+             *   ...   ...      ...  a    a^2
+             *   ...   ...      ...  0    0
+             *
+             * ]
+             */
+            meta.lookup(&[Column::<Any>::from(a)], &[Column::<Any>::from(sl)]);
+            meta.lookup(
+                &[Column::<Any>::from(a), Column::<Any>::from(b)],
+                &[Column::<Any>::from(sl), Column::<Any>::from(sl2)],
+            );
```
therealyingtong, line 361:
```suggestion
            meta.lookup(&[a.into()], &[sl.into()]);
            meta.lookup(&[a.into(), b.into()], &[sl.into(), sl2.into()]);
```

src/plonk/circuit.rs
```diff
@@ -353,6 +358,34 @@ impl<F: Field> ConstraintSystem<F> {
         index
     }
 
+    /// Add a lookup argument for some input columns and table columns.
+    pub fn lookup(
```
therealyingtong, line 362:
```suggestion
    /// Add a lookup argument for some input columns and table columns.
    /// The function will panic if the number of input columns and table
    /// columns are not the same.
    pub fn lookup(
```

src/plonk/circuit.rs
```diff
@@ -353,6 +358,34 @@ impl<F: Field> ConstraintSystem<F> {
...
+        &mut self,
+        input_columns: &[Column<Any>],
+        table_columns: &[Column<Any>],
+    ) -> usize {
+        // The function will panic if the number of input columns and table columns are not the same.
```
therealyingtong, line 367:
```suggestion
```

src/plonk/lookup.rs
```diff
@@ -0,0 +1,33 @@
+use super::circuit::{Any, Column};
+use crate::arithmetic::CurveAffine;
+
+pub(crate) mod prover;
```
therealyingtong, line 4:
```suggestion
mod prover;
```

src/plonk/lookup/verifier.rs
```diff
@@ -0,0 +1,149 @@
...
+        theta: ChallengeTheta<C::Scalar>,
+        beta: ChallengeBeta<C::Scalar>,
+        gamma: ChallengeGamma<C::Scalar>,
+        advice_evals: &[C::Scalar],
+        fixed_evals: &[C::Scalar],
+        aux_evals: &[C::Scalar],
+    ) -> impl Iterator<Item = C::Scalar> + 'a {
+        let product_expression = || {
+            // z'(X) (a'(X) + \beta) (s'(X) + \gamma)
+            // - z'(\omega^{-1} X) (\theta^m a_1(X) + \theta^{m-1} a_2(X) + ... + a_m(X) + \beta) (\theta^m s_1(X) + \theta^{m-1} s_2(X) + ... + s_m(X) + \gamma)
```
daira, line 55: The comments in this file also have the off-by-one error.

src/plonk/lookup/prover.rs
```diff
@@ -0,0 +1,621 @@
...
+        // (\theta^{m-1} a_1(\omega^i) + \theta^{m-2} a_2(\omega^i) + ... + \theta a_{m-1}(\omega^i) + a_m(\omega^i) + \beta)
+        // * (\theta^{m-1} s_1(\omega^i) + \theta^{m-2} s_2(\omega^i) + ... + \theta s_{m-1}(\omega^i) + s_m(\omega^i) + \gamma)
+        parallelize(&mut lookup_product, |product, start| {
+            for (i, product) in product.iter_mut().enumerate() {
+                let i = i + start;
+
+                // Compress unpermuted input columns
+                let mut input_term = C::Scalar::zero();
+                for unpermuted_input_column in self.unpermuted_input_columns.iter() {
+                    input_term *= &theta;
```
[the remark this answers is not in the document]
daira: Hmm, clippy also seems to be complaining in a similar way about a bunch of code not added in this PR. Must be due to some other recent refactoring.

daira (changes requested):
The inconsistencies in the comments about compressing by theta are blocking.

src/plonk/lookup/prover.rs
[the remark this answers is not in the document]
therealyingtong: yeah, it seems like clippy isn't using our `Deref` impl for `ChallengeScalar`.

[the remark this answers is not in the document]
ebfull: There are many issues with clippy that we'll be resolving once we've emptied the PR queue, because we introduced clippy lints in CI during a lot of ongoing development.

ebfull (approved):

daira (approved):
utACK

PR #59, opened 2020-11-12
Documentation updates + clippy fixes

therealyingtong:


src/poly/commitment.rs
```diff
@@ -279,14 +279,13 @@ fn test_opening_proof() {
     transcript.absorb(Fp::from_bytes(&v.to_bytes()).unwrap()); // unlikely to fail since p ~ q
 
     loop {
-        let transcript_dup = transcript.clone();
+        let mut transcript_dup = transcript.clone();
 
         let opening_proof = Proof::create(&params, &mut transcript, &px, blind, x);
         if opening_proof.is_err() {
```
daira, line 285: Style nit: use matching.

src/poly/commitment/verifier.rs
```diff
@@ -32,14 +32,14 @@ impl<'a, C: CurveAffine> Guard<'a, C> {
     /// scalars and points.
     pub fn use_challenges(mut self) -> MSM<'a, C> {
         let s = compute_s(&self.challenges_sq, self.allinv * &self.neg_z1);
-        self.msm.add_to_g(&s);
-        self.msm.add_to_h(self.neg_z1);
+        self.msm.add_to_g_scalars(&s);
```
daira, line 35: `add` -> `append`

src/poly/commitment/verifier.rs
```diff
@@ -162,9 +162,9 @@ impl<C: CurveAffine> Proof<C> {
         let c_packed = transcript.squeeze().get_lower_128();
         let c: C::Scalar = get_challenge_scalar(Challenge(c_packed));
 
-        // Check
-        // [c] P + [c * v] U + [c] sum(L_i * u_i^2) + [c] sum(R_i * u_i^-2) + delta - [z1] G - [z1 * b] U - [z1 - z2] H
-        // = 0
+        // Construct
+        // [c] P + [c * v] U + [c] sum(L_i * u_i^2) + [c] sum(R_i * u_i^-2) + delta - [z1 * b] U + [z1 - z2] H
+        // = [z1] (G + H)
```
daira, line 167: Say where the equality is checked.

src/poly/commitment/verifier.rs
ebfull: It isn't appended, it's actually added to the scalar.

src/poly/commitment.rs
ebfull: Agreed and in fact I think this should just be changed to use matching.

src/poly/commitment/verifier.rs
ebfull: The comment is correct to note that we've modified where the actual computation of `[z1] (G + H)` happens. If we just add a detail here that "the computation of `[z1] (G + H)` happens in either `Guard::use_challenges` or `Guard::use_g`" that would be cool.

ebfull: I double-checked and yeah, good catch on the `+ [z1 - z2] H` thing.

src/poly/commitment/verifier.rs
therealyingtong: right, i think we want `add_term` -> `append_term`.

src/poly/commitment/msm.rs
```diff
@@ -34,52 +34,58 @@ impl<'a, C: CurveAffine> MSM<'a, C> {
...
+    pub fn append_term(&mut self, scalar: C::Scalar, point: C) {
         self.other_scalars.push(scalar);
         self.other_bases.push(point);
     }
 
     /// Add a vector of scalars to `g_scalars`. This function will panic if the
     /// caller provides a slice of scalars that is not of length `params.n`.
-    // TODO: parallelize
-    pub fn add_to_g(&mut self, scalars: &[C::Scalar]) {
+    pub fn add_to_g_scalars(&mut self, scalars: &[C::Scalar]) {
```
daira, line 53: `append_to_g_scalars`

src/poly/commitment/msm.rs
```diff
@@ -34,52 +34,58 @@ impl<'a, C: CurveAffine> MSM<'a, C> {
...
+            })
         } else {
             self.g_scalars = Some(scalars.to_vec());
         }
     }
 
-    /// Add term to h
-    pub fn add_to_h(&mut self, scalar: C::Scalar) {
+    /// Add to `h_scalar`
+    pub fn add_to_h_scalar(&mut self, scalar: C::Scalar) {
```
daira, line 67: `append_to_h_scalar`

daira (approved):
utACK with non-blocking comments.

src/poly/commitment/msm.rs
therealyingtong: this does actually add to the scalar; the confusingly named method was `add_term()` which we've changed in this PR to `append_term()`.

src/poly/commitment/msm.rs
therealyingtong: this does actually add to the scalar; the confusingly named method was `add_term()` which we've changed in this PR to `append_term()`.

daira (approved):
utACK

src/poly/commitment.rs
daira: clippy is still complaining, and I agree.

str4d: Agreed; this should be:
```
        if let Ok(opening_proof) = opening_proof {
            // Code currently in the `else` clause
        } else {
            transcript = transcript_dup;
            transcript.absorb_base(Field::one());
        }
```

str4d (approved):
utACK mod my comment in https://github.com/zcash/halo2/pull/59/files#r532633044

str4d (approved):

PR #61, opened 2020-11-13
Migrate to ff traits

str4d:
The `Field` trait in this crate is now `FieldExt: ff::PrimeField`.

### DIFF
--- a/src/tweedle/fields/fp.rs
+++ b/src/tweedle/fields/fp.rs
@@ -1,10 +1,11 @@
+use bitvec::{array::BitArray, order::Lsb0};
 use core::convert::TryInto;
 use core::fmt;
 use core::ops::{Add, Mul, Neg, Sub};
-
+use rand::RngCore;
 use subtle::{Choice, ConditionallySelectable, ConstantTimeEq, CtOption};
 
-use crate::arithmetic::{adc, mac, sbb, Field, Group};
+use crate::arithmetic::{adc, mac, sbb, FieldExt, Group};
 
 /// This represents an element of $\mathbb{F}_p$ where
 ///
@@ -101,6 +102,19 @@ const MODULUS: Fp = Fp([
     0x4000000000000000,
 ]);
 
+/// The modulus as u32 limbs.
+#[cfg(not(target_pointer_width = "64"))]
+const MODULUS_LIMBS_32: [u32; 8] = [
+    0x0000_0001,
+    0xa140_64e2,
+    0x6c3f_59b9,
+    0x038a_a127,
+    0x0000_0000,
+    0x0000_0000,
+    0x0000_0000,
+    0x4000_0000,
+];
+
 impl<'a> Neg for &'a Fp {
     type Output = Fp;
 
@@ -174,6 +188,15 @@ const R3: Fp = Fp([
     0xf9fdbeb55b7eb87c,
     0x63f75cb999eafa89,
     0x217cb214ebb8fc72,
+]);
+
+/// `GENERATOR = 5 mod p` is a generator of the `p - 1` order multiplicative subgroup, and
+/// is also a quadratic non-residue.
+const GENERATOR: Fp = Fp::from_raw([
+    0x0000_0000_0000_0005,
+    0x0000_0000_0000_0000,
+    0x0000_0000_0000_0000,
+    0x0000_0000_0000_0000,
 ]);
 
 const S: u32 = 33;
@@ -416,6 +439,12 @@ impl Fp {
     }
 }
 
+impl From<Fp> for [u8; 32] {
+    fn from(value: Fp) -> [u8; 32] {
+        value.to_bytes()
+    }
+}
+
 impl<'a> From<&'a Fp> for [u8; 32] {
     fn from(value: &'a Fp) -> [u8; 32] {
         value.to_bytes()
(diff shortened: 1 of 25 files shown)


src/tweedle/fields/fp.rs
```diff
@@ -569,6 +544,150 @@ impl Field for Fp {
...
+            u64::from_le_bytes(bytes[8..16].try_into().unwrap()),
+            u64::from_le_bytes(bytes[16..24].try_into().unwrap()),
+            u64::from_le_bytes(bytes[24..32].try_into().unwrap()),
+        ];
+
+        BitArray::new(limbs)
+    }
+
+    fn is_odd(&self) -> bool {
+        todo!()
```
str4d, line 615: Should this be the usual definition?
```suggestion
        self.to_bytes()[0] & 1 == 1
```

src/tweedle/fields/fp.rs
```diff
@@ -569,6 +544,150 @@ impl Field for Fp {
...
+        {
+            BitArray::new(MODULUS_LIMBS_32)
+        }
+
+        #[cfg(target_pointer_width = "64")]
+        BitArray::new(MODULUS.0)
+    }
+
+    fn multiplicative_generator() -> Self {
+        todo!()
```
str4d, line 629: @ebfull is `GENERATOR` defined anywhere with a different name? Both `ROOT_OF_UNITY` and `DELTA` have doc comments from `GENERATOR`.

src/tweedle/fields/fq.rs
```diff
@@ -584,6 +559,150 @@ impl Field for Fq {
...
+            u64::from_le_bytes(bytes[8..16].try_into().unwrap()),
+            u64::from_le_bytes(bytes[16..24].try_into().unwrap()),
+            u64::from_le_bytes(bytes[24..32].try_into().unwrap()),
+        ];
+
+        BitArray::new(limbs)
+    }
+
+    fn is_odd(&self) -> bool {
+        todo!()
```
str4d, line 630: Ditto.

src/tweedle/fields/fq.rs
```diff
@@ -584,6 +559,150 @@ impl Field for Fq {
...
+        {
+            BitArray::new(MODULUS_LIMBS_32)
+        }
+
+        #[cfg(target_pointer_width = "64")]
+        BitArray::new(MODULUS.0)
+    }
+
+    fn multiplicative_generator() -> Self {
+        todo!()
```
str4d, line 644: Ditto.

src/tweedle/fields/fq.rs
```diff
@@ -101,6 +102,19 @@ const MODULUS: Fq = Fq([
...
+#[cfg(not(target_pointer_width = "64"))]
+const MODULUS_LIMBS_32: [u32; 8] = [
+    0x0000_0001,
+    0x842c_afd4,
+    0x6962_86c9,
+    0x038a_a127,
+    0x0000_0000,
+    0x0000_0000,
+    0x0000_0000,
+    0x4000_0000,
```
daira, line 115: I checked this.

src/tweedle/fields/fq.rs
```diff
@@ -584,6 +559,150 @@ impl Field for Fq {
...
+        0xd023bfdccccccccd,
+        0x360880ec544ed23a,
+        0x3333333333333333,
+        0x3333333333333333,
+    ];
+    const ZETA: Self = Fq::from_raw([
+        0x4394c2bd148fa4fd,
+        0x69cf8de720e52ec1,
+        0x87ad8b5ff9731ffe,
+        0x36c66d3a1e049a58,
```
daira, line 691: I did *not* check these constants.

daira (approved):
ACK modulo @str4d's comments.

src/tweedle/fields/fp.rs
ebfull: Yes.

src/tweedle/fields/fp.rs
ebfull: `GENERATOR` = 5 for both fields.

ebfull: (but it's not defined anywhere in the code currently)

src/tweedle/fields/fp.rs
```diff
@@ -569,6 +553,150 @@ impl Field for Fp {
         CtOption::new(tmp, !self.ct_eq(&Self::zero()))
     }
 
+    fn pow_vartime<S: AsRef<[u64]>>(&self, exp: S) -> Self {
```
ebfull, line 556: This is provided by `Field` already, right?

src/tweedle/fields/fp.rs
```diff
@@ -176,6 +190,15 @@ const R3: Fp = Fp([
     0x217cb214ebb8fc72,
 ]);
 
+/// `GENERATOR = 5 mod p` is a generator of the `p - 1` order multiplicative subgroup, and
+/// is also a quadratic non-residue.
+const GENERATOR: Fp = Fp([
```
ebfull, line 195: Can we do `Fp::from_raw([5, 0, 0, 0])` here like we do below?

src/tweedle/fields/fq.rs
```diff
@@ -176,6 +190,15 @@ const R3: Fq = Fq([
     0xa42cdf95f959577,
 ]);
 
+/// `GENERATOR = 5 mod q` is a generator of the `q - 1` order multiplicative subgroup, and
+/// is also a quadratic non-residue.
+const GENERATOR: Fq = Fq([
```
ebfull, line 195: Similarly, `Fq::from_raw([5, 0, 0, 0])`

src/tweedle/fields/fq.rs
```diff
@@ -584,6 +568,150 @@ impl Field for Fq {
         CtOption::new(tmp, !self.ct_eq(&Self::zero()))
     }
 
+    fn pow_vartime<S: AsRef<[u64]>>(&self, exp: S) -> Self {
```
ebfull, line 571: Similarly, this is provided by `Field` already.

ebfull (changes requested):

src/tweedle/fields/fp.rs
str4d: Yes, but the `ff` implementation doesn't skip squarings before finding a one, and the existing impl here did, so I assumed you wanted this behaviour.

ebfull (approved):

PR #70, opened 2020-11-25
Extract permutation argument and introduce typed challenges

### LINKED ISSUE #62: Modularise prover and verifier by introducing internal APIs

str4d:
Part of #62.

### DIFF
--- a/src/plonk.rs
+++ b/src/plonk.rs
@@ -7,7 +7,8 @@
 
 use crate::arithmetic::CurveAffine;
 use crate::poly::{
-    multiopen, Coeff, EvaluationDomain, ExtendedLagrangeCoeff, LagrangeCoeff, Polynomial,
+    commitment::ChallengeScalar, multiopen, Coeff, EvaluationDomain, ExtendedLagrangeCoeff,
+    LagrangeCoeff, Polynomial,
 };
 
 mod circuit;
@@ -92,6 +93,22 @@ impl<C: CurveAffine> VerifyingKey<C> {
         &self.domain
     }
 }
+
+#[derive(Clone, Copy, Debug)]
+pub(crate) struct Beta;
+pub(crate) type ChallengeBeta<F> = ChallengeScalar<F, Beta>;
+
+#[derive(Clone, Copy, Debug)]
+pub(crate) struct Gamma;
+pub(crate) type ChallengeGamma<F> = ChallengeScalar<F, Gamma>;
+
+#[derive(Clone, Copy, Debug)]
+pub(crate) struct Y;
+pub(crate) type ChallengeY<F> = ChallengeScalar<F, Y>;
+
+#[derive(Clone, Copy, Debug)]
+pub(crate) struct X;
+pub(crate) type ChallengeX<F> = ChallengeScalar<F, X>;
 
 #[test]
 fn test_proving() {
--- a/src/plonk.rs
+++ b/src/plonk.rs
@@ -8,7 +8,7 @@
 use crate::arithmetic::CurveAffine;
 use crate::poly::{
     commitment::ChallengeScalar, multiopen, Coeff, EvaluationDomain, ExtendedLagrangeCoeff,
-    LagrangeCoeff, Polynomial,
+    Polynomial,
 };
 
 mod circuit;
@@ -28,7 +28,7 @@ pub use verifier::*;
 pub struct VerifyingKey<C: CurveAffine> {
     domain: EvaluationDomain<C::Scalar>,
     fixed_commitments: Vec<C>,
-    permutation_commitments: Vec<Vec<C>>,
+    permutations: Vec<permutation::VerifyingKey<C>>,
     cs: ConstraintSystem<C::Scalar>,
 }
 
@@ -41,9 +41,7 @@ pub struct ProvingKey<C: CurveAffine> {
     l0: Polynomial<C::Scalar, ExtendedLagrangeCoeff>,
     fixed_polys: Vec<Polynomial<C::Scalar, Coeff>>,
     fixed_cosets: Vec<Polynomial<C::Scalar, ExtendedLagrangeCoeff>>,
-    permutations: Vec<Vec<Polynomial<C::Scalar, LagrangeCoeff>>>,
-    permutation_polys: Vec<Vec<Polynomial<C::Scalar, Coeff>>>,
-    permutation_cosets: Vec<Vec<Polynomial<C::Scalar, ExtendedLagrangeCoeff>>>,
+    permutations: Vec<permutation::ProvingKey<C>>,
 }
 
 /// This is an object which represents a (Turbo)PLONK proof.
--- a/src/plonk.rs
+++ b/src/plonk.rs
@@ -6,10 +6,8 @@
 //! [plonk]: https://eprint.iacr.org/2019/953
 
 use crate::arithmetic::CurveAffine;
-use crate::poly::{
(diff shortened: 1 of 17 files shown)


src/plonk/verifier.rs
```diff
@@ -52,10 +51,8 @@ impl<'a, C: CurveAffine> Proof<C> {
         let x_1: C::Scalar = get_challenge_scalar(Challenge(transcript.squeeze().get_lower_128()));
```
therealyingtong, line 51: it seems asymmetric that `x_0`, `x_1` are sampled within the permutation argument on the prover's side, but within the main PLONK argument on the verifier's side.

str4d:
Rebased to fix merge conflicts with #59.

src/plonk/permutation.rs
```diff
@@ -0,0 +1,14 @@
+//! Implementation of a PLONK permutation argument.
+
+use crate::arithmetic::CurveAffine;
+
+mod prover;
+mod verifier;
+
+#[derive(Debug, Clone)]
+pub(crate) struct Proof<C: CurveAffine> {
```
str4d, line 9: This currently uses a single permutation "proof" to represent all the permutations added to the circuit. This was the most straight-forward refactoring of the existing code (making review of the moving pieces easier), but we will probably want to alter this in a subsequent PR to instead have a proof per permutation (similar to the lookup argument PR).

str4d: I've now extracted the permutation keygen logic as well, and I've done so in the same way as the lookup argument (`Vec<permutation::Argument>`).

src/plonk/permutation/prover.rs
```diff
@@ -0,0 +1,391 @@
...
+}
+
+pub(crate) struct Evaluated<C: CurveAffine> {
+    constructed: Constructed<C>,
+    permutation_product_evals: Vec<C::Scalar>,
+    permutation_product_inv_evals: Vec<C::Scalar>,
+    permutation_evals: Vec<Vec<C::Scalar>>,
+}
+
+impl<C: CurveAffine> Proof<C> {
```
therealyingtong, line 37: i think this should be `impl<C: CurveAffine> Argument<C>` instead.

src/plonk/permutation/verifier.rs
```diff
@@ -0,0 +1,157 @@
...
+        self.permutation_product_evals
+            .iter()
+            .chain(self.permutation_product_inv_evals.iter())
+            .chain(self.permutation_evals.iter().flat_map(|evals| evals.iter()))
+    }
+
+    pub(crate) fn queries<'a>(
+        &'a self,
+        vk: &'a VerifyingKey<C>,
+        x: C::Scalar,
```
therealyingtong, line 114: should this also be `x: ChallengeX<C::Scalar>`?

str4d: Good catch! I found a different place this was forgotten yesterday, but didn't spot this one.

src/poly/commitment.rs
```diff
@@ -15,6 +16,78 @@ mod verifier;
...
+impl<F: Field, T> Deref for ChallengeScalar<F, T> {
+    type Target = F;
+
+    fn deref(&self) -> &F {
+        &self.inner
+    }
+}
+
+#[derive(Clone, Copy, Debug)]
+pub(crate) struct X6 {}
```
ebfull, line 88: Can we call this challenge `Z` instead?

src/poly/commitment/prover.rs
```diff
@@ -20,12 +19,12 @@ impl<C: CurveAffine> Proof<C> {
     /// opening v, and the point x. It's probably also nice for the transcript
     /// to have seen the elliptic curve description and the SRS, if you want to
     /// be rigorous.
-    pub fn create<HBase, HScalar>(
+    pub(crate) fn create<HBase, HScalar>(
```
ebfull, line 22: This effectively hides the `Proof` functionality behind the `multiopen` module. I think that's okay, since it provides a strict superset of behavior, but is it what you intended?

src/poly/multiopen/prover.rs
```diff
@@ -29,8 +28,8 @@ impl<C: CurveAffine> Proof<C> {
     where
         I: IntoIterator<Item = ProverQuery<'a, C>> + Clone,
     {
-        let x_4: C::Scalar = get_challenge_scalar(Challenge(transcript.squeeze().get_lower_128()));
-        let x_5: C::Scalar = get_challenge_scalar(Challenge(transcript.squeeze().get_lower_128()));
+        let x_4 = ChallengeScalar::<_, ()>::get(transcript);
+        let x_5 = ChallengeScalar::<_, ()>::get(transcript);
```
ebfull, line 32: Let's rename these challenges to

`X1`, `X2`, (and `X3`, `X4` below) since the numbering no longer makes sense. Regarding the effective shadowing of challenge names, see my other comment.

ebfull (changes requested):

str4d:
Rebased to fix conflicts with #61.

src/plonk.rs
```diff
@@ -95,6 +92,22 @@ impl<C: CurveAffine> VerifyingKey<C> {
     }
 }
 
+#[derive(Clone, Copy, Debug)]
+pub(crate) struct Beta;
```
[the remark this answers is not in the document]
str4d: This was required because they were used in `pub(crate)` interfaces, but is fixed by the `pub(in crate::plonk)` suggestion 🙂

src/plonk/permutation/prover.rs
```diff
@@ -0,0 +1,391 @@
...
+impl<C: CurveAffine> Constructed<C> {
+    pub(crate) fn evaluate<HBase: Hasher<C::Base>, HScalar: Hasher<C::Scalar>>(
+        self,
+        pk: &ProvingKey<C>,
+        x: ChallengeX<C::Scalar>,
+        transcript: &mut Transcript<C, HBase, HScalar>,
+    ) -> Evaluated<C> {
+        let domain = &pk.vk.domain;
+
+        let permutation_product_evals: Vec<C::Scalar> = self
```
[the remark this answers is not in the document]
str4d: They need `Vec<_>` at a minimum, because we then need to iterate over them to add them to the transcript (and the type needs to be known at that point). But the `C::Scalar` is not necessary here.

src/plonk/prover.rs
```diff
@@ -170,203 +167,52 @@ impl<C: CurveAffine> Proof<C> {
...
-                .absorb_point(c)
-                .map_err(|_| Error::TranscriptError)?;
-        }
+        // Sample beta challenge
+        let beta = ChallengeBeta::get(&mut transcript);
+
+        // Sample gamma challenge
+        let gamma = ChallengeGamma::get(&mut transcript);
+
+        // Commit to permutations, if any.
```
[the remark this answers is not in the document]
str4d: Let's look at this again once the lookups PR is merged. Opened #82

src/plonk/permutation/prover.rs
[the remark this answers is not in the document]
ebfull: ah. that does seems like a defect in the rust compiler's inference capabilities though

src/poly/commitment/prover.rs
str4d: This was mainly because it would require making `ChallengeZ` public, whereas the `multiopen` prover doesn't take in any challenges at all in its public API. If you're fine with `ChallengeZ` being public, I can make this `pub` again.

str4d: Fixed.

ebfull (approved):

PR #71, opened 2020-11-25
Use latest stable Rust

str4d:
We don't need to commit to an MSRV yet, and this avoids the problem
where some of our in-flux dependencies bump their MSRV (or don't have
a defined MSRV and make changes to use more recent features) causing
our CI to break.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions-rs/toolchain@v1
         with:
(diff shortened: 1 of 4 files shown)


daira (approved):
utACK

daira:
Looks like the stable clippy CI checks are not working, though.

daira:
I would suggest disabling the `renamed-and-removed-lints` error (on the command line) for stable.

str4d:
Given that after this PR we target stable, we can happily rename that lint.

PR #89, opened 2020-12-15
SHA-256 gadget

str4d:


src/gadget/sha256.rs
```diff
@@ -0,0 +1,274 @@
...
+}
+
+/// Implementation of the compression function for SHA-256.
+#[derive(Debug)]
+pub struct Sha256Compress<CS: Sha256CompressInstructions> {
+    _cs: PhantomData<CS>,
+}
+
+impl<CS: Sha256CompressInstructions> Sha256Compress<CS> {
+    /// Starting from the given intial state, processes a block of input and returns the
```
rex4539, line 138:
```suggestion
    /// Starting from the given initial state, processes a block of input and returns the
```

src/gadget/sha256.rs
```diff
@@ -0,0 +1,274 @@
...
+        let mut g = initial_state.h_6.clone();
+        let mut h = initial_state.h_7.clone();
+
+        let w = CS::message_schedule(layer, input)?;
+
+        for (&k_t, w_t) in ROUND_CONSTANTS.iter().zip(w.iter()) {
+            let h_prime = CS::h_prime(layer, &e, &f, &g, h, k_t, w_t)?;
+
+            let a_new = CS::a_new(layer, &a, &b, &c, &h_prime)?;
+            let e_new = CS::e_new(layer, d, &h_prime)?;
```
daira, line 160: A chip may have an implementation of `h_prime`, `a_new`, and `e_new` that use relative cell references to pass variables between rows. The permutation arguments may not cover all advice columns, and so relative references may be required for efficiency. The layouter is supposed to be a general algorithm; the chip could require certain capabilities from it, but it's not specific to this chip. And the gadget could call the instruction methods in any order.

How will the layouter construct a layout that wires up the relative references correctly?

daira: One possibility is that you can only use relative references within an instruction, but I think this is too restrictive. Consider a scalar multiplication or Sinsemilla gadget: we know that there's a natural ordering of the steps of Sinsemilla or of a particular scalar multiplication algorithm. The problem is to design an API that separates concerns between the gadget and the chip, while constraining the gadget to make calls in the "right" order to avoid the layouter having to solve an unnecessary hard layout problem. One way to do it is using session types: the chip defines an API that enforces a particular ordering of calls from the gadget. But I have a feeling we're missing something, since this seems overcomplicated. In the cases we want to immediately support (scalar mul, Sinsemilla, Rescue and Poseidon), the required ordering is just a linear sequence of steps. This suggests a more functional API: we construct a vector of steps for a gadget and pass that to the layouter all at once. Then the layouter has a much easier job. In general it might need to lay out instructions from different gadgets in parallel, but we could reasonably require (at least for now) that the permutation argument covers enough columns to allow loading data in and out of gadget blocks that are laid out in parallel.

daira: On further consideration, I think we cannot meaningfully separate concerns between the gadget and the chip without letting the chip essentially implement the whole high-level algorithm.

I started off by trying to change the Chip API in the simplest way I could think of to avoid the call ordering problem described above:
```
impl<Chip: Sha256CompressInstructions> Sha256Compress<Chip> {
    fn process(
        layer: &mut impl LayoutStrategy<Chip>,
        initial_state: &State<Chip>,
        input: [Chip::BlockWord; 16],
    ) -> Result<State<Chip>, Error> {
        let mut state = initial_state.clone();
        let w = Chip::message_schedule(layer, input)?;
        for (&k_t, w_t) in ROUND_CONSTANTS.iter().zip(w.iter()) {
            state = Chip::round(&state, k_t, w_t)?;
        }
        Chip::feedforward(&state, &initial_state)?;
}
```
with the assumption that `round` will be called in the obvious order.

But at this point we're not really gaining anything from separating out the `message_schedule`, `round`, and `feedforward` functions. We might as well just define a `compress` instruction that does the whole thing.

str4d: @daira and I discussed this in a pairing, and we discussed a variety of issues for which I now have notes 😄 For this PR, we will take the approach of defining a single `compress` instruction. My decision to agree on this was in part influenced by seeing that the Intel SHA extensions define an instruction that operates on two compression rounds simultaneously, which is less granular than the instructions we'd defined here.

nuttycom: I would really like to look at how we might construct this API starting from the high-level view of a gadget/chip/whateveryouwanttocallit as a transformation on tables, and then from there deconstruct to the smaller transformations on tables that need to be composed to achieve the larger one. I really think that designing this API imperatively is muddying the waters. One can always implement a declarative API in terms of mutation.

daira: I agree with @nuttycom that it is simpler and more general to start from a declarative API design.

In #93 I sketched out a simple layout algorithm that should suffice for the circuits we need in the short term. This clarifies that all the input a layout algorithm needs is the block shapes (where relative references are only allowed within a block). So I claim that a gadget should just output a set of blocks. There's no requirement that it use an imperative API to do so.

src/gadget.rs
```diff
@@ -0,0 +1,175 @@
...
+
+    /// Assign a fixed value
+    fn assign_fixed<'v>(
+        &'v mut self,
+        column: Column<Fixed>,
+        offset: usize,
+        to: &'v mut (dyn FnMut() -> Result<C::Field, Error> + 'v),
+    ) -> Result<Cell, Error>;
+
+    /// Constraint two cells to have the same value.
```
daira, line 90:
```suggestion
    /// Constrain two cells to have the same value.
```

src/gadget.rs
```diff
@@ -0,0 +1,175 @@
...
+    fn assign_fixed<'v>(
+        &'v mut self,
+        column: Column<Fixed>,
+        offset: usize,
+        mut to: impl FnMut() -> Result<C::Field, Error> + 'v,
+    ) -> Result<Cell, Error> {
+        self.region.assign_fixed(column, offset, &mut to)
+    }
+
+    /// Constraint two cells to have the same value.
```
daira, line 137:
```suggestion
    /// Constrain two cells to have the same value.
```

src/gadget.rs
```diff
@@ -0,0 +1,175 @@
...
+        }
+    }
+}
+
+/// This trait allows a [`Chip`] to direct a [`Layouter`] to assign cells within a
+/// region.
+///
+/// TODO: It would be great if we could constrain the columns in these types to be
+/// "logical" columns that are guaranteed to correspond to the chip (and have come from
+/// `Chip::Config`).
```
daira, line 72: One way to do this would be to have `DynRegion` and the chip parameterized on the type of columns that it uses. Then it would only be able to refer to its own columns in places where that type variable is used (even though it is always instantiated as Column<Any>, it's still treated as a different type per chip instance).

Unfortunately this is is complicated by the categorization of columns. It's still possible, for example we could use `DynRegion<CHIP: Chip, CELL: Cell, CFixed: Column<Fixed>, CAdvice: Column<Advice>>`. (This variant also constrains Cells to be in the right column.)

src/gadget.rs
```diff
@@ -0,0 +1,175 @@
+//! Self-contained circuit implementations of various primitives.
+
+use std::fmt;
+
+use crate::{
+    arithmetic::FieldExt,
+    plonk::{Advice, Any, Column, ConstraintSystem, Error, Fixed},
+};
+
+pub mod sha256;
```
daira, line 10: This module should not depend on `sha256`.

src/gadget.rs
daira: Another possibility, that is probably cleaner now that I think of it, is to parameterize on a Region `R`, i.e. use `Cell<R>` and `Column<R, {Any, Fixed, Advice, Aux}>`.

str4d: A chip might contains multiple regions, but if this only needs to be a distinguisher then we could use the chip itself as the "region" type here (or a concrete type that it knows).

src/gadget.rs
str4d: It doesn't? Nothing in this module depends on this line; it just exposes the `sha256` submodule to crate users.

src/gadget.rs
daira: We'd also need to be able to refer to columns that are *not* constrained to a chip. I think maybe that pushes the complexity of the `Cell<R>` and `Column<R, {Any, Fixed, Advice, Aux}>` approach above that of the `DynRegion<CHIP: Chip, CELL: Cell, CFixed: Column<Fixed>, CAdvice: Column<Advice>>` approach.

src/gadget.rs
daira: What's special about the `sha256` module that we should publically expose it here, rather than having its users import it directly? IOW I think it should not be a submodule of `gadget`, because user-implemented concrete gadget modules won't be.

str4d: Am I misunderstanding you? A user can't "import it directly" in Rust. If the module files are located in `halo2::gadget::sha256` (which they currently are), then that module path needs to be exposed in the public API. Alternatively, the files could be moved somewhere else, or re-exported as e.g. `halo::sha256` (which starts polluting the top-level namespace of the `halo2` crate). But there *must* be a `pub mod sha256` somewhere, or else the gadget is unreachable by end users.

src/gadget/sha256/table16/gates.rs
```diff
@@ -0,0 +1,116 @@
...
+
+        // Scale the whole expression by factor to avoid divisions
+        let factor = factorial((deg - 1) as u64);
+
+        let numerator = |var: Expression<F>, eval: u32, idx: u64| {
+            let mut expr = Self::ONES;
+            for i in 0..deg {
+                let i = i as u64;
+                if i != idx {
+                    expr = expr * (Self::ONES * (-F::one()) * F::from_u64(i) + var.clone());
```
daira, line 34: Consider adding a `From` impl so that we can say
```suggestion
                    expr = expr * ((-F::one()) * F::from_u64(i) + var.clone()).into();
```

therealyingtong:
Superseded by https://github.com/zcash/halo2/pull/127.

Rationale for closing: Most comments on this PR were made on the early commits, which have been extracted into https://github.com/zcash/halo2/pull/125. Rebasing this PR on top of #125 would lose these comments, but closing the PR preserves them.

PR #95, opened 2020-12-22
Add MockProver for developing circuits

str4d:


src/dev.rs
```diff
@@ -0,0 +1,199 @@
...
+            .max()
+            .unwrap_or(1);
+
+        // Account for each gate to ensure our quotient polynomial is the
+        // correct degree and that our extended domain is the right size.
+        for poly in cs.gates.iter() {
+            degree = std::cmp::max(degree, poly.degree());
+        }
+
+        let domain = EvaluationDomain::new(degree as u32, k);
```
str4d, line 111: Currently nothing in `MockProver` actually relies on this, so if neither the permutation check nor anything else needs it, I can remove it to speed up tests.

str4d:
Rebased on main to bring in the CI fixes from #96.

src/dev.rs
```diff
@@ -0,0 +1,293 @@
...
+use crate::{
+    arithmetic::{FieldExt, Group},
+    plonk::{Any, Assignment, Circuit, Column, ConstraintSystem, Error},
+    poly::{EvaluationDomain, LagrangeCoeff, Polynomial},
+};
+
+#[derive(Debug)]
+struct Cell(usize, usize);
+
+/// The reasons why a particular circuit is not satisfied.
```
daira, line 14: Maybe we should move this out of dev so that we can more precisely report unsatisfiedness (dissatisfaction?) errors in the normal case. I think it's always useful to do that even if it is more expensive, since unsatisfiedness indicates a serious bug in either the circuit implementation or the provider of the high-level witness.

src/dev.rs
```diff
@@ -0,0 +1,293 @@
...
+pub enum VerifyFailure {
+    /// A gate was not satisfied for a particular row.
+    Gate { gate_index: usize, row: usize },
+    /// A lookup input did not exist in its corresponding table.
+    Lookup { lookup_index: usize, row: usize },
+    /// A permutation did not preserve the original value of a cell.
+    Permutation {
+        perm_index: usize,
+        column: usize,
+        row: usize,
```
daira, line 25: If the size is correct then there must be *both* some value in the input that was included too many times, and some value in the table that wasn't included enough times. Which one are we reporting? Should we report both?

src/dev.rs
```diff
@@ -0,0 +1,293 @@
...
+    }
+
+    fn copy(
+        &mut self,
+        permutation: usize,
+        left_column: usize,
+        left_row: usize,
+        right_column: usize,
+        right_row: usize,
+    ) -> Result<(), crate::plonk::Error> {
```
daira, line 90: This seems to be duplicated from the normal prover; does it need to be?

src/dev.rs
```diff
@@ -0,0 +1,293 @@
...
+        }
+
+        // Check that all lookups exist in their respective tables.
+        for (lookup_index, lookup) in self.cs.lookups.iter().enumerate() {
+            for input_row in 0..n {
+                let load = |column: &Column<Any>, row| match column.column_type() {
+                    Any::Fixed => self.fixed[column.index()][row as usize],
+                    Any::Advice => self.advice[column.index()][row as usize],
+                    Any::Aux => self.aux[column.index()][row as usize],
+                };
```
daira, line 243: This seems like it is generally useful and should be abstracted out.

src/dev.rs
str4d: We obviously can't use this for normal verification, because we don't have any of the witness information 🙂

We could possibly include this within the prover's errors (by caching more information during the proof-creation process, to enable this granularity), but in general it would be better to encourage `Circuit` structs to be designed in a way that makes it impossible to construct invalid witness assignments.

An alternative we could explore is that if creation of a proof fails, we immediately run that circuit assignment through the `MockProver` to get the more granular details.

src/dev.rs
daira: I realized after I wrote this that we are checking whether a *specific* permutation is correct, when we report this error. Can we make this clearer in the doc?

src/dev.rs
str4d: I have de-duplicated this.

src/dev.rs
str4d: Opened #99 for this.

src/dev.rs
ebfull: Yep, from the circuit's perspective the domain should be irrelevant due to the `Rotation` API keeping the actual points in the domain opaque. You can remove this.

src/dev.rs
```diff
@@ -0,0 +1,343 @@
...
+                            perm_index,
+                            column,
+                            row,
+                        });
+                    }
+                }
+            }
+        }
+
+        // TODO: Implement the rest of the verification checks.
```
ebfull, line 339: There shouldn't be any additional checks right?

src/plonk.rs
```diff
@@ -14,7 +14,7 @@ use crate::transcript::ChallengeScalar;
 mod circuit;
 mod keygen;
 mod lookup;
-mod permutation;
+pub(crate) mod permutation;
```
ebfull, line 17: Ideally we wouldn't need to do this (the permutation module here should use a shared implementation of @daira and @defuse's permutation algorithm) but this is fine for now.

src/dev.rs
```diff
@@ -0,0 +1,343 @@
...
+use ff::Field;
+
+use crate::{
+    arithmetic::{FieldExt, Group},
+    plonk::{permutation, Any, Assignment, Circuit, Column, ConstraintSystem, Error},
+    poly::{EvaluationDomain, LagrangeCoeff, Polynomial},
+};
+
+#[derive(Debug)]
+struct Cell(usize, usize);
```
ebfull, line 12: Does this need to be introduced in this PR? It's not used anywhere AFAICT. If it's helpful, though, let's introduce it in the `circuit` module instead.

Also, consider doing `(Column, usize)` instead of `(usize, usize)`.

ebfull (changes requested):
Minor nits.

src/dev.rs
str4d: Now that the permutation checks have been added, I think this comment is moot.

src/dev.rs
str4d: It is leftover from my placeholder Permutation collector, which Ying Tong replaced. I'll remove it.

daira (approved):
ACK

str4d:
Rebased on main to bring in the CI fix from #119.

PR #96, opened 2020-12-22
Bump metrics crate and pin its dependencies

str4d:
Fixes CI breakage caused by the latest metrics alpha release.


str4d:
Trivial PR; self-ACK.

PR #110, opened 2020-12-23
Remove rotations from ConstraintSystem & use Rotation instead of i32 for circuit methods.

nuttycom:
The first commit of this branch just removes the unused `rotations` field from ConstraintSystem.
The second commit switches to using Rotation in the circuit interface rather than i32; it can be dropped if folks don't care for it.

### DIFF
--- a/src/plonk/lookup/prover.rs
+++ b/src/plonk/lookup/prover.rs
@@ -94,7 +94,7 @@ impl Argument {
                     };
                     (
                         &values[column.index()],
-                        &cosets[pk.vk.cs.get_any_query_index(column, 0)],
+                        &cosets[pk.vk.cs.get_any_query_index(column, Rotation::cur())],
                     )
                 })
                 .unzip();
@@ -148,7 +148,7 @@ impl Argument {
         let permuted_input_coset = pk
             .vk
             .domain
-            .coeff_to_extended(permuted_input_poly.clone(), Rotation::default());
+            .coeff_to_extended(permuted_input_poly.clone(), Rotation::cur());
         let permuted_input_inv_coset = pk
             .vk
             .domain
@@ -156,7 +156,7 @@ impl Argument {
         let permuted_table_coset = pk
             .vk
             .domain
-            .coeff_to_extended(permuted_table_poly.clone(), Rotation::default());
+            .coeff_to_extended(permuted_table_poly.clone(), Rotation::cur());
 
         Ok(Permuted {
             unpermuted_input_columns,
@@ -310,11 +310,8 @@ impl<'a, C: CurveAffine> Permuted<'a, C> {
         let product_blind = Blind(C::Scalar::rand());
         let product_commitment = params.commit_lagrange(&z, product_blind).to_affine();
         let z = pk.vk.domain.lagrange_to_coeff(z);
-        let product_coset = pk
-            .vk
-            .domain
-            .coeff_to_extended(z.clone(), Rotation::default());
-        let product_inv_coset = pk.vk.domain.coeff_to_extended(z.clone(), Rotation(-1));
+        let product_coset = pk.vk.domain.coeff_to_extended(z.clone(), Rotation::cur());
+        let product_inv_coset = pk.vk.domain.coeff_to_extended(z.clone(), Rotation::prev());
 
         // Hash product commitment
         transcript
--- a/src/plonk/lookup/verifier.rs
+++ b/src/plonk/lookup/verifier.rs
@@ -120,7 +120,7 @@ impl<C: CurveAffine> Evaluated<C> {
                 columns
                     .iter()
                     .map(|column| {
-                        let index = vk.cs.get_any_query_index(*column, 0);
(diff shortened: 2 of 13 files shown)


src/poly.rs
```diff
@@ -200,3 +200,32 @@ impl<'a, F: Field, B: Basis> Mul<F> for Polynomial<F, B> {
         self
     }
 }
+
+/// Describes a relative location in the evaluation domain; applying a rotation
+/// by i will rotate the vector in the evaluation domain by i.
+#[derive(Copy, Clone, Debug, Hash, PartialEq, Eq, Ord, PartialOrd)]
```
ebfull, line 206: It's best not to derive these things unless and until we need them so that we can later expand the API contract; there may be reasons (that we cannot guess now) that would make it problematic for a user to be able to sort and hash these rotations. Anecdote is that in bellman being able to `Hash` a `Variable` would have allowed a user to create non-deterministic hashmaps in circuit synthesis code, which meant that circuit synthesis would be non-deterministic.

ebfull: I would support `Copy, Clone, Debug` though.

src/poly.rs
```diff
@@ -200,3 +200,32 @@ impl<'a, F: Field, B: Basis> Mul<F> for Polynomial<F, B> {
         self
     }
 }
+
+/// Describes a relative location in the evaluation domain; applying a rotation
```
ebfull, line 204: Now that we're using this in two different ways, a more generic description might be:

```
/// Describes the relative rotation of a vector. Negative numbers represent
/// reverse (leftmost) rotations and positive numbers represent forward (rightmost)
/// rotations. Zero represents no rotation.
```

src/plonk/lookup/verifier.rs
```diff
@@ -61,7 +61,7 @@ impl<C: CurveAffine> Proof<C> {
                 columns
                     .iter()
                     .map(|column| {
-                        let index = vk.cs.get_any_query_index(*column, 0);
+                        let index = vk.cs.get_any_query_index(*column, Rotation::default());
```
ebfull, line 64: Use `cur()` here to be consistent with the rest of the PR.

src/plonk/lookup/prover.rs
```diff
@@ -107,7 +107,7 @@ impl Argument {
                     };
                     (
                         &values[column.index()],
-                        &cosets[pk.vk.cs.get_any_query_index(column, 0)],
+                        &cosets[pk.vk.cs.get_any_query_index(column, Rotation::default())],
```
ebfull, line 110: Use `cur()` here to be consistent with the rest of the PR.

ebfull (changes requested):

src/poly.rs
ebfull: Also, I'm not entirely sure it makes sense to sort rotations because Rotation(a) = Rotation(b) could be true for a particular vector due to length overflow.

ebfull (changes requested):
I was wrong apparently, we need `PartialEq` derive on `Rotation`.

ebfull (dismissed):
`benches/plonk.rs` needs to be updated as well with the API changes. Run `cargo bench plonk`.

src/plonk/lookup/prover.rs
```diff
@@ -323,10 +323,7 @@ impl<'a, C: CurveAffine> Permuted<'a, C> {
         let product_blind = Blind(C::Scalar::rand());
         let product_commitment = params.commit_lagrange(&z, product_blind).to_affine();
         let z = pk.vk.domain.lagrange_to_coeff(z);
-        let product_coset = pk
-            .vk
-            .domain
-            .coeff_to_extended(z.clone(), Rotation::default());
+        let product_coset = pk.vk.domain.coeff_to_extended(z.clone(), Rotation::cur());
         let product_inv_coset = pk.vk.domain.coeff_to_extended(z.clone(), Rotation(-1));
```
str4d, line 327:
```suggestion
        let product_inv_coset = pk.vk.domain.coeff_to_extended(z.clone(), Rotation::prev());
```

src/plonk/permutation/prover.rs
```diff
@@ -122,7 +122,7 @@ impl Argument {
         let permutation_product_blind = blind;
         let z = domain.lagrange_to_coeff(z);
         let permutation_product_poly = z.clone();
-        let permutation_product_coset = domain.coeff_to_extended(z.clone(), Rotation::default());
+        let permutation_product_coset = domain.coeff_to_extended(z.clone(), Rotation::cur());
         let permutation_product_coset_inv = domain.coeff_to_extended(z, Rotation(-1));
```
str4d, line 126:
```suggestion
        let permutation_product_coset_inv = domain.coeff_to_extended(z, Rotation::prev());
```

str4d (approved):
ACK modulo comments.

str4d (changes requested):
Doc tests need updating to use `Rotation`.

str4d (approved):

ebfull (approved):

Issue #67, opened 2020-11-23
Multi-proof prover


str4d:
It's possible to structure halo2 proofs in such a way that they can share resources. We can then create multiple proofs at the same time, with most of the proof size and time in a common block, and then marginal size and time costs per proof.

PR #111, opened 2020-12-23
New Transcript API (and modified commitment scheme)

### LINKED ISSUE #66: Refactor proofs to not require length encodings
We should refactor how proofs are formed, so that they e.g. come in as byte streams, so that we don't need to encode lengths in the proofs.

ebfull:
Closes #66.

The goal of this PR is to solve multiple problems:

1. Currently, we have various `Proof` objects scattered around the codebase for each of the different subprotocols. Some of these proof objects contain vectors of (vectors of) curve points and scalars. This complicates serialization/deserialization of proofs because the lengths of these vectors depend on the configuration of the circuit. We don't want to encode the lengths of vectors inside of proofs because we'll know them at runtime and they do not change. We should instead treat proof objects as opaque byte streams into the verifier.
2. It's easy to accidentally put stuff in a `Proof` object that isn't also placed in the transcript.
3. In #67 we need to be able to create multiple PLONK proofs at the same time and they will share many different substructures when they're for the same circuit.
4. (Bonus) Now verification also accounts for the cost of deserialization, which isn't negligible due to point compression.

The idea is to refactor the `transcript` module to provide two separate concepts: a `TranscriptWrite` trait that represents something we can write to (at proving time) and a `TranscriptRead` trait that represents something we can read from (at verifying time). Crucially, implementations of these traits are responsible for simultaneously writing to some `std::io::Write` buffer at the same time that they hash things into the transcript, and similarly for `TranscriptRead/std::io::Read`.

We can then remove all of the `Proof`-like structures from the codebase.

In addition, we perform a few refactorings of the PLONK implementation to reflect the fact that the verifier needs to temporarily store values in the proof and cannot simply access a `Proof` object anymore. This makes the verifier look a lot more like the prover, which does the same thing for other reasons.

In order to make this easier I had to also change the commitment scheme to more closely match the one from the accumulation scheme paper. That involved three changes:

1. We subtract `[v] G_0` from the commitment `P` that we're opening at `x`; the argument from then forward becomes "does the polynomial committed at `P` have a root at `x`" which will be efficient inside of recursive circuits because it's just a fixed-based scalar multiplication, versus the variable base `P' = P + [v] U` thing that we do in the Halo paper.
2. The prover supplies a commitment `R` to a random polynomial that also has a root at `x`. The verifier samples random `iota` and computes `P' = P - [v] G_0 + [iota] R` and the inner product argument is run on `P'`. This makes it so that we don't need to do a Schnorr proof at the end because `P'` is uniformly distributed in the space of polynomials with a root at `x`.
3. We don't sample `U` for checking the values; we use a fixed `U` and compute a challenge `z` to compute `U' = [z] U` and use `U'` for that purpose. It was never really necessary for `U` to be sampled from the group, which I didn't realize at the time.

### DIFF
--- a/src/arithmetic/curves.rs
+++ b/src/arithmetic/curves.rs
@@ -122,6 +122,10 @@ pub trait CurveAffine:
     /// The base field over which this elliptic curve is constructed.
     type Base: FieldExt;
 
+    /// Personalization of BLAKE2b hasher used to generate the uniform
+    /// random string.
+    const BLAKE2B_PERSONALIZATION: &'static [u8; 16];
+
     /// Obtains the additive identity.
     fn zero() -> Self;
 
--- a/src/pasta/curves.rs
+++ b/src/pasta/curves.rs
@@ -11,7 +11,7 @@ use super::{Fp, Fq};
 use crate::arithmetic::{Curve, CurveAffine, FieldExt, Group};
 
 macro_rules! new_curve_impl {
-    ($name:ident, $name_affine:ident, $base:ident, $scalar:ident) => {
+    ($name:ident, $name_affine:ident, $base:ident, $scalar:ident, $blake2b_personalization:literal) => {
         /// Represents a point in the projective coordinate space.
         #[derive(Copy, Clone, Debug)]
         pub struct $name {
@@ -497,6 +497,8 @@ macro_rules! new_curve_impl {
             type Scalar = $scalar;
             type Base = $base;
 
+            const BLAKE2B_PERSONALIZATION: &'static [u8; 16] = $blake2b_personalization;
+
             fn zero() -> Self {
                 Self {
                     x: $base::zero(),
@@ -700,5 +702,5 @@ macro_rules! new_curve_impl {
     };
 }
 
-new_curve_impl!(Ep, EpAffine, Fp, Fq);
-new_curve_impl!(Eq, EqAffine, Fq, Fp);
+new_curve_impl!(Ep, EpAffine, Fp, Fq, b"halo2_____pallas");
+new_curve_impl!(Eq, EqAffine, Fq, Fp, b"halo2______vesta");
--- a/src/plonk.rs
+++ b/src/plonk.rs
@@ -107,7 +107,6 @@ fn test_proving() {
     use crate::poly::commitment::{Blind, Params};
     use crate::transcript::{DummyHashRead, DummyHashWrite, TranscriptRead, TranscriptWrite};
     use circuit::{Advice, Column, Fixed};
-    use std::io;
     use std::marker::PhantomData;
     const K: u32 = 5;
 
@@ -116,7 +115,7 @@ fn test_proving() {
     pub struct Variable(Column<Advice>, usize);
 
     // Initialize the polynomial commitment parameters
-    let params: Params<EqAffine> = Params::new::<DummyHashWrite<io::Sink, _>>(K);
+    let params: Params<EqAffine> = Params::new(K);
 
     struct PLONKConfig {
         a: Column<Advice>,
--- a/src/plonk.rs
+++ b/src/plonk.rs
@@ -105,7 +105,7 @@ fn test_proving() {
     use crate::arithmetic::{Curve, FieldExt};
     use crate::pasta::{EqAffine, Fp, Fq};
     use crate::poly::commitment::{Blind, Params};
-    use crate::transcript::{DummyHashReader, DummyHashWriter, TranscriptRead, TranscriptWrite};
+    use crate::transcript::{DummyHashRead, DummyHashWrite, TranscriptRead, TranscriptWrite};
     use circuit::{Advice, Column, Fixed};
     use std::io;
     use std::marker::PhantomData;
@@ -116,7 +116,7 @@ fn test_proving() {
     pub struct Variable(Column<Advice>, usize);
 
     // Initialize the polynomial commitment parameters
-    let params: Params<EqAffine> = Params::new::<DummyHashWriter<io::Sink, _>>(K);
+    let params: Params<EqAffine> = Params::new::<DummyHashWrite<io::Sink, _>>(K);
 
     struct PLONKConfig {
         a: Column<Advice>,
@@ -448,7 +448,7 @@ fn test_proving() {
         .to_affine();
 
     for _ in 0..100 {
-        let mut transcript = DummyHashWriter::init(vec![], Fq::one());
+        let mut transcript = DummyHashWrite::init(vec![], Fq::one());
         // Create a proof
         create_proof(
             &params,
@@ -462,7 +462,7 @@ fn test_proving() {
 
         let pubinput_slice = &[pubinput];
         let msm = params.empty_msm();
-        let mut transcript = DummyHashReader::init(&proof[..], Fq::one());
+        let mut transcript = DummyHashRead::init(&proof[..], Fq::one());
         let guard =
             verify_proof(&params, pk.get_vk(), msm, pubinput_slice, &mut transcript).unwrap();
         {
@@ -476,7 +476,7 @@ fn test_proving() {
         }
         let msm = guard.clone().use_challenges();
         assert!(msm.clone().eval());
-        let mut transcript = DummyHashReader::init(&proof[..], Fq::one());
+        let mut transcript = DummyHashRead::init(&proof[..], Fq::one());
         let guard =
             verify_proof(&params, pk.get_vk(), msm, pubinput_slice, &mut transcript).unwrap();
         {
--- a/src/plonk.rs
+++ b/src/plonk.rs
@@ -6,9 +6,7 @@
 //! [plonk]: https://eprint.iacr.org/2019/953
 
 use crate::arithmetic::CurveAffine;
-use crate::poly::{
-    multiopen, Coeff, EvaluationDomain, ExtendedLagrangeCoeff, LagrangeCoeff, Polynomial,
-};
+use crate::poly::{Coeff, EvaluationDomain, ExtendedLagrangeCoeff, LagrangeCoeff, Polynomial};
 use crate::transcript::ChallengeScalar;
 
 mod circuit;
@@ -46,20 +44,6 @@ pub struct ProvingKey<C: CurveAffine> {
     fixed_polys: Vec<Polynomial<C::Scalar, Coeff>>,
     fixed_cosets: Vec<Polynomial<C::Scalar, ExtendedLagrangeCoeff>>,
     permutations: Vec<permutation::ProvingKey<C>>,
-}
-
-/// This is an object which represents a (Turbo)PLONK proof.
-// This structure must never allow points at infinity.
-#[derive(Debug, Clone)]
-pub struct Proof<C: CurveAffine> {
-    advice_commitments: Vec<C>,
-    permutations: Vec<permutation::Proof<C>>,
-    lookups: Vec<lookup::Proof<C>>,
-    advice_evals: Vec<C::Scalar>,
-    aux_evals: Vec<C::Scalar>,
-    fixed_evals: Vec<C::Scalar>,
-    vanishing: vanishing::Proof<C>,
-    multiopening: multiopen::Proof<C>,
 }
 
 /// This is an error that could occur during proving or circuit synthesis.
@@ -121,8 +105,9 @@ fn test_proving() {
     use crate::arithmetic::{Curve, FieldExt};
     use crate::pasta::{EqAffine, Fp, Fq};
     use crate::poly::commitment::{Blind, Params};
-    use crate::transcript::DummyHash;
+    use crate::transcript::{DummyHashReader, DummyHashWriter, TranscriptRead, TranscriptWrite};
     use circuit::{Advice, Column, Fixed};
+    use std::io;
     use std::marker::PhantomData;
     const K: u32 = 5;
 
@@ -131,7 +116,7 @@ fn test_proving() {
     pub struct Variable(Column<Advice>, usize);
 
     // Initialize the polynomial commitment parameters
-    let params: Params<EqAffine> = Params::new::<DummyHash<Fq>>(K);
+    let params: Params<EqAffine> = Params::new::<DummyHashWriter<io::Sink, _>>(K);
 
     struct PLONKConfig {
         a: Column<Advice>,
@@ -463,20 +448,23 @@ fn test_proving() {
         .to_affine();
 
     for _ in 0..100 {
+        let mut transcript = DummyHashWriter::init(vec![], Fq::one());
         // Create a proof
-        let proof = Proof::create::<DummyHash<Fq>, DummyHash<Fp>, _>(
+        create_proof(
             &params,
             &pk,
             &circuit,
             &[pubinputs.clone()],
+            &mut transcript,
         )
         .expect("proof generation should not fail");
+        let proof: Vec<u8> = transcript.finalize();
 
         let pubinput_slice = &[pubinput];
         let msm = params.empty_msm();
-        let guard = proof
-            .verify::<DummyHash<Fq>, DummyHash<Fp>>(&params, pk.get_vk(), msm, pubinput_slice)
-            .unwrap();
+        let mut transcript = DummyHashReader::init(&proof[..], Fq::one());
+        let guard =
+            verify_proof(&params, pk.get_vk(), msm, pubinput_slice, &mut transcript).unwrap();
         {
             let msm = guard.clone().use_challenges();
             assert!(msm.eval());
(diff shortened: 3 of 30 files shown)


ebfull:
Highly recommend reviewing with whitespace changes hidden.

src/poly/commitment/verifier.rs
```diff
@@ -62,144 +64,104 @@ impl<'a, C: CurveAffine> Guard<'a, C> {
...
+/// Checks to see if an [`Proof`] is valid given the current `transcript`, and a
+/// point `x` that the polynomial commitment `P` opens purportedly to the value
+/// `v`. The provided `commitment_msm` should evaluate to the commitment `P`
+/// being opened.
+pub fn verify_proof<'a, C: CurveAffine, R: Read, T: TranscriptRead<R, C>>(
+    params: &'a Params<C>,
+    mut msm: MSM<'a, C>,
+    transcript: &mut T,
+    x: C::Scalar,
+    mut commitment_msm: MSM<'a, C>,
```
ebfull, line 76: This can be merged with `msm` now that we don't scale the commitment we're opening by anything.

src/poly/commitment/prover.rs
```diff
@@ -1,202 +1,183 @@
...
+        let value_r = compute_inner_product(&a[half..], &b[0..half]);
+        let mut l_randomness = C::Scalar::rand();
+        let r_randomness = C::Scalar::rand();
+        metrics::counter!("multiexp", 2, "val" => "l/r", "size" => "2");
+        let l = l + &best_multiexp(&[value_l * &z, l_randomness], &[params.u, params.h]);
+        let r = r + &best_multiexp(&[value_r * &z, r_randomness], &[params.u, params.h]);
+        let mut l = l.to_affine();
+        let r = r.to_affine();
+
+        let challenge = loop {
```
ebfull, line 113: We should be able to change the argument so that we aren't required to sample squares, to be more in line with the accumulation scheme paper. This would allow us to remove the entire `.fork()` and related APIs from the `transcript` module.

src/transcript.rs
```diff
@@ -2,125 +2,221 @@
...
+    fn squeeze_challenge(&mut self) -> C::Base;
+
+    /// Writing the point to the transcript without writing it to the proof,
+    /// treating it as a common input.
+    fn common_point(&mut self, point: C) -> io::Result<()>;
+}
+
+/// Transcript view from the perspective of a verifier that has access to an
+/// input stream of data from the prover to the verifier.
+pub trait TranscriptRead<R: Read, C: CurveAffine>: Transcript<C> {
```
ebfull, line 24: I think that the `R` type parameter is unnecessary and can be removed, especially after removing the need for transcript forking.

src/transcript.rs
```diff
@@ -2,125 +2,221 @@
...
+    /// Read a curve point from the prover.
+    fn read_point(&mut self) -> io::Result<C>;
+
+    /// Read a curve scalar from the prover.
+    fn read_scalar(&mut self) -> io::Result<C::Scalar>;
+}
+
+/// Transcript view from the perspective of a prover that has access to an
+/// output stream of messages from the prover to the verifier.
+pub trait TranscriptWrite<W: Write, C: CurveAffine>: Transcript<C> {
```
ebfull, line 37: Similarly we can remove this `W` type parameter and the `finalize()` method can be removed from the trait. It's of no interest to the prover implementation what the buffer is that it's writing to, so it shouldn't be a type parameter at all.

src/poly/commitment.rs
```diff
@@ -25,21 +27,13 @@ pub struct Params<C: CurveAffine> {
...
-    z1: C::Scalar,
-    z2: C::Scalar,
+    pub(crate) u: C,
 }
 
 impl<C: CurveAffine> Params<C> {
     /// Initializes parameters for the curve, given a random oracle to draw
     /// points from.
-    pub fn new<H: Hasher<C::Base>>(k: u32) -> Self {
+    pub fn new<T: TranscriptWrite<io::Sink, C> + Sync>(k: u32) -> Self {
```
ebfull, line 36: In order to remove the writer type parameter we'll have to abandon the (misuse) of `TranscriptWrite` in this API. I'll just bring in `blake2b` and we'll sample points for the parameters with that instead until we have a better plan.

src/transcript.rs
```diff
@@ -2,125 +2,221 @@
...
-    /// in both the base_hasher and scalar_hasher
-    pub fn new() -> Self {
-        let base_hasher = HBase::init(C::Base::one());
-        let scalar_hasher = HScalar::init(C::Scalar::one());
-        Transcript {
-            base_hasher,
-            scalar_hasher,
-            scalar_needs_squeezing: false,
-            _marker: PhantomData,
+pub struct DummyHashWriter<W: Write, C: CurveAffine> {
```
ebfull, line 145: nit: `DummyHashWriter` is nounish, whereas `TranscriptWrite` is verbish.

src/transcript.rs
```diff
@@ -2,125 +2,221 @@
...
-/// This is just a simple (and completely broken) hash function, standing in for
-/// some algebraic hash function that we'll switch to later.
+/// This is just a simple (and completely broken) transcript reader
+/// implementation, standing in for some algebraic hash function that we'll
+/// switch to later.
 #[derive(Debug, Clone)]
-pub struct DummyHash<F: FieldExt> {
-    power: F,
-    state: F,
+pub struct DummyHashReader<R: Read, C: CurveAffine> {
```
ebfull, line 63: nit: `DummyHashReader` is nounish, whereas `TranscriptRead` is verbish.

src/plonk/verifier.rs
```diff
@@ -1,216 +1,105 @@
...
-            )
-            .map_err(|_| Error::OpeningError)
+use crate::transcript::{read_n_points, read_n_scalars, TranscriptRead};
+
+/// Returns a boolean indicating whether or not the proof is valid
+pub fn verify_proof<'a, C: CurveAffine, R: Read, T: TranscriptRead<R, C>>(
+    params: &'a Params<C>,
+    vk: &'a VerifyingKey<C>,
+    msm: MSM<'a, C>,
+    aux_commitments: &'a [C],
```
ebfull, line 21: These lifetime bounds are unnecessary.

src/poly/commitment/prover.rs
ebfull: I'll do this in a separate PR to make review simpler.

str4d:
It would be great for this PR to also add a design page (probably `book/src/design/implementation.md`) describing why we do this / why proofs are read and written via the transcript 😁

daira (approved):
utACK. I have not checked the protocol changes.

src/poly/commitment/verifier.rs
```diff
@@ -62,174 +61,122 @@ impl<'a, C: CurveAffine> Guard<'a, C> {
...
+    //   + \sum(L_i * u_i^2) + \sum(R_i * u_i^-2)
+    // at x to 0, by asking the prover to supply (a, h) such that it equals
+    //   = [a] (G + [b * z] U) + [h] H
+    // except that we wish for the prover to supply G as Commit(g(X); 1) so
+    // we must substitute to get
+    //   = [a] ((G - H) + [b * z] U) + [h] H
+    //   = [a] G + [-a] H + [abz] U + [h] H
+    //   = [a] G + [abz] U + [h - a] H
+    // but subtracting to get the desired equality
+    //   ... + [-a] G + [-abz] U + [a - h] H = 0
```
daira, line 130: What's the `...` here?

src/poly/commitment/verifier.rs
```diff
@@ -62,174 +61,122 @@ impl<'a, C: CurveAffine> Guard<'a, C> {
...
     }
+
+    // Our goal is to open
+    //     msm - [v] G_0 + random_poly_commitment * iota
+    //   + \sum(L_i * u_i^2) + \sum(R_i * u_i^-2)
+    // at x to 0, by asking the prover to supply (a, h) such that it equals
+    //   = [a] (G + [b * z] U) + [h] H
+    // except that we wish for the prover to supply G as Commit(g(X); 1) so
+    // we must substitute to get
+    //   = [a] ((G - H) + [b * z] U) + [h] H
```
daira, line 126: I don't understand how this is a substitution into the previous line.

src/transcript.rs
```diff
@@ -2,125 +2,203 @@
...
-        if bool::from(tmp.is_none()) {
-            return Err(());
-        };
-        let tmp = tmp.unwrap();
-        self.base_hasher.absorb(tmp.0);
-        self.base_hasher.absorb(tmp.1);
-        Ok(())
+    /// Conclude the interaction and return the output buffer (writer).
+    pub fn finalize(self) -> W {
+        // TODO: handle outstanding scalars?
```
daira, line 151: File a ticket and link to it.

src/poly/commitment/verifier.rs
ebfull: It's the left hand side of the equality, `msm - [v] G_0 + random_poly_commitment * iota + \sum(L_i * u_i^2) + \sum(R_i * u_i^-2)`.

src/poly/commitment/verifier.rs
ebfull: It's the left hand side of the equality, `msm - [v] G_0 + random_poly_commitment * iota + \sum(L_i * u_i^2) + \sum(R_i * u_i^-2)`.

src/poly/commitment/verifier.rs
ebfull: The prover will not be giving us `G` but rather `G + H` (the blinding factor is 1), so we have to subtract `H` from what they give us. We do this because auxillary commitments always have a blinding factor of 1 in the PLONK API, ensuring that under no circumstance will points at infinity emerge inside of a recursive circuit.

src/transcript.rs
```diff
@@ -284,3 +284,25 @@ impl<C: CurveAffine, Type> Deref for ChallengeScalar<C, Type> {
         &self.inner
     }
 }
+
+pub(crate) fn read_n_points<C: CurveAffine, R: Read, T: TranscriptRead<R, C>>(
```
str4d: I think this should be a `read_points` method on `TranscriptRead` (or perhaps a new `TranscriptReadExt`) with a default implementation (and likewise for `read_scalars`).

src/transcript.rs
```diff
@@ -284,3 +284,25 @@ impl<C: CurveAffine, Type> Deref for ChallengeScalar<C, Type> {
...
+    transcript: &mut T,
+    n: usize,
+) -> io::Result<Vec<C>> {
+    let mut v = Vec::with_capacity(n);
+    for _ in 0..n {
+        v.push(transcript.read_point()?);
+    }
+    Ok(v)
```
str4d, line 278:
```suggestion
    (0..n).map(|_| transcript.read_point()).collect()
```

src/transcript.rs
```diff
@@ -284,3 +284,25 @@ impl<C: CurveAffine, Type> Deref for ChallengeScalar<C, Type> {
...
+
+pub(crate) fn read_n_scalars<C: CurveAffine, R: Read, T: TranscriptRead<R, C>>(
+    transcript: &mut T,
+    n: usize,
+) -> io::Result<Vec<C::Scalar>> {
+    let mut v = Vec::with_capacity(n);
+    for _ in 0..n {
+        v.push(transcript.read_scalar()?);
+    }
+    Ok(v)
```
str4d, line 289:
```suggestion
    (0..n).map(|_| transcript.read_scalar()).collect()
```

src/arithmetic/curves.rs
```diff
@@ -122,6 +122,10 @@ pub trait CurveAffine:
     /// The base field over which this elliptic curve is constructed.
     type Base: FieldExt;
 
+    /// Personalization of BLAKE2b hasher used to generate the uniform
+    /// random string.
+    const BLAKE2B_PERSONALIZATION: &'static [u8; 16];
```
str4d, line 127: Not sure I like this being a property of the curve implementation, but it will do for now. We will likely need to refactor this later.

src/pasta/curves.rs
```diff
@@ -700,5 +702,5 @@ macro_rules! new_curve_impl {
     };
 }
 
-new_curve_impl!(Ep, EpAffine, Fp, Fq);
-new_curve_impl!(Eq, EqAffine, Fq, Fp);
+new_curve_impl!(Ep, EpAffine, Fp, Fq, b"halo2_____pallas");
+new_curve_impl!(Eq, EqAffine, Fq, Fp, b"halo2______vesta");
```
str4d, line 706: These _definitely_ will need changing to be more descriptive as to their purpose 🙂

src/poly/commitment.rs
```diff
@@ -42,27 +42,38 @@ impl<C: CurveAffine> Params<C> {
...
-        let g = {
-            let hasher = &T::init(io::sink(), C::Base::one());
+        let try_and_increment = |hasher: &Blake2bState| {
+            let mut trial = 0u64;
+            loop {
+                let mut hasher = hasher.clone();
+                hasher.update(&(trial.to_le_bytes())[..]);
+                let mut hash = [0u8; 32];
+                hash[..].copy_from_slice(hasher.finalize().as_bytes());
+                let p = C::from_bytes(&hash);
```
str4d, line 52: This can probably be:
```suggestion
                let p = C::from_bytes(&hasher.finalize().as_bytes().try_into().unwrap());
```

str4d (commented):
I reviewed everything except the first commit from #113.

src/poly/commitment/verifier.rs
```diff
@@ -142,50 +132,51 @@ pub fn verify_proof<'a, C: CurveAffine, T: TranscriptRead<C>>(
     let a = transcript.read_scalar().map_err(|_| Error::SamplingError)?;
     let neg_a = -a;
     let h = transcript.read_scalar().map_err(|_| Error::SamplingError)?;
```
str4d, line 134: This should be named `blind` to match the prover.

src/poly/multiopen/prover.rs
```diff
@@ -23,138 +24,118 @@ struct CommitmentData<C: CurveAffine> {
...
-                    f_poly.map(|f_poly| f_poly * *x_2 + &poly)
-                }
-            })
-            .unwrap();
-
-        let mut f_blind = Blind(C::Scalar::rand());
-        let mut f_commitment = params.commit(&f_poly, f_blind).to_affine();
-
-        let (opening, q_evals) = loop {
-            let mut transcript = transcript.clone();
```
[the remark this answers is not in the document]
str4d: Answer: removing the sampling of `U` from the PC prover meant it became infallible, so the multiopen prover no longer needed to fork to handle a potential failure.

src/poly/commitment/prover.rs
```diff
@@ -1,202 +1,183 @@
...
-        transcript
-            .absorb_point(&delta)
-            .map_err(|_| Error::SamplingError)?;
-
-        // Obtain the challenge c.
-        let c = ChallengeScalar::<C::Scalar, ()>::get(transcript);
-
-        // Compute z1 and z2 as described in the Halo paper.
-        let z1 = a * &*c + &d;
-        let z2 = *c * &blind + &s;
```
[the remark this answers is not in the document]
ebfull: Yeah, we're abandoning the Schnorr proof in favor of this because it's more efficient in the circuit.

src/plonk.rs
```diff
@@ -48,20 +46,6 @@ pub struct ProvingKey<C: CurveAffine> {
     permutations: Vec<permutation::ProvingKey<C>>,
 }
 
-/// This is an object which represents a (Turbo)PLONK proof.
-// This structure must never allow points at infinity.
```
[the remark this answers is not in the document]
ebfull:
```rust
        let (x, y) = Option::from(point.get_xy()).ok_or(io::Error::new(
            io::ErrorKind::Other,
            "cannot write points at infinity to the transcript",
        ))?;
```

src/plonk/lookup/prover.rs
```diff
@@ -520,29 +517,16 @@ impl<C: CurveAffine> Evaluated<C> {
...
-                eval: self.permuted_input_eval,
+                eval: self.permuted_input_inv_eval,
             }))
             // Open lookup product commitments at x_inv
             .chain(Some(ProverQuery {
                 point: x_inv,
                 poly: &self.constructed.product_poly,
                 blind: self.constructed.product_blind,
-                eval: self.product_eval,
+                eval: self.product_inv_eval,
```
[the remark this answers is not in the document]
ebfull: Looking into this, will make an issue if necessary.

src/plonk/verifier.rs
```diff
@@ -1,216 +1,105 @@
...
+    // Hash each lookup product commitment
+    let lookups = lookups
+        .into_iter()
+        .map(|lookup| lookup.absorb_product_commitment(transcript))
+        .collect::<Result<Vec<_>, _>>()?;
+
+    // Sample y challenge, which keeps the gates linearly independent.
+    let y = ChallengeY::get(transcript);
+
+    let vanishing = vanishing::Argument::absorb_commitments(vk, transcript)?;
```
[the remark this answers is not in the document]
ebfull: Done.

src/plonk/lookup/prover.rs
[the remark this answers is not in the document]
str4d: Addressed by #139.

src/pasta/curves.rs
daira: This will change again when we switch to simplified SWU for hash-to-curve.

PR #113, opened 2020-12-24
Avoid square challenges in inner product argument

ebfull:
**The base of this branch is currently `transcript-api-2` which is [unmerged](https://github.com/zcash/halo2/pull/111).**

This modifies the scheme to be almost identical to the construction outlined in Appendix A.2 of "Proof-Carrying Data from Accumulation Schemes" (https://eprint.iacr.org/2020/499). The only remaining difference is that we do not compute `[v] U = [v] [z] U` but instead subtract `[v] G_0` from the commitment before opening which I figure will be more efficient in the circuit.

As a consequence of this, things are a little simpler and we can remove both

* transcript forking
* deterministic square root computation

as I don't forsee us needing either of these.

### DIFF
--- a/src/arithmetic/fields.rs
+++ b/src/arithmetic/fields.rs
@@ -16,15 +16,6 @@ pub trait FieldExt:
     /// Inverse of `ROOT_OF_UNITY`
     const ROOT_OF_UNITY_INV: Self;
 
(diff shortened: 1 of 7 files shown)


daira (approved):
This is very nice. ACK.

PR #119, opened 2021-01-06
Fix breakage of trait resolution in Rust 1.49.0

str4d:
Previously, `ChallengeScalar` could use the operator traits defined on
the `F: Field` type it wrapped, due to its `impl Deref<Target = F>`.
This was technically ambiguous, and Rust 1.49.0 makes that ambiguity an
error.

We could fix this by adding operator impls with `ChallengeScalar` on the
RHS, but that would conflict with zcash/halo2#111. Instead we manually
dereference every challenge scalar when used in an arithmetic operation.

### DIFF
--- a/src/plonk/lookup/prover.rs
+++ b/src/plonk/lookup/prover.rs
@@ -243,19 +243,19 @@ impl<'a, C: CurveAffine> Permuted<'a, C> {
                 // Compress unpermuted input columns
                 let mut input_term = C::Scalar::zero();
                 for unpermuted_input_column in self.unpermuted_input_columns.iter() {
-                    input_term *= &theta;
+                    input_term *= &*theta;
                     input_term += &unpermuted_input_column[i];
                 }
 
                 // Compress unpermuted table columns
                 let mut table_term = C::Scalar::zero();
                 for unpermuted_table_column in self.unpermuted_table_columns.iter() {
-                    table_term *= &theta;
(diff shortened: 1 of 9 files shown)


src/plonk/lookup/prover.rs
```diff
@@ -306,11 +306,11 @@ impl<'a, C: CurveAffine> Permuted<'a, C> {
...
-                    .fold(C::Scalar::zero(), |acc, input| acc * &theta + &input[i]);
+                    .fold(C::Scalar::zero(), |acc, input| acc * &*theta + &input[i]);
 
                 let mut table_term = self.unpermuted_table_columns
                     .iter()
-                    .fold(C::Scalar::zero(), |acc, table| acc * &theta + &table[i]);
+                    .fold(C::Scalar::zero(), |acc, table| acc * &*theta + &table[i]);
 
                 input_term += &(*beta);
                 table_term += &(*gamma);
```
daira, line 316:
```suggestion
                input_term += &*beta;
                table_term += &*gamma;
```
and similarly elsewhere.

daira (approved):
ACK. Comment is nonblocking.

src/plonk/lookup/prover.rs
str4d: We're likely to remove all this dereferencing later anyway, by implementing the necessary operator traits after @ebfull is done with refactoring.

Issue #72, opened 2020-11-29
Implement table-based variant of Sarkar's square root algorithm


daira:
A Sage prototype for the Pasta fields is here: https://github.com/zcash/pasta/blob/master/squareroottab.sage

This includes a few minor optimizations and simplifications that are not in the [paper](https://eprint.iacr.org/2020/1407). We essentially follow section 3 but specialize it to the case ℓ<sub>0..k-2</sub> = w, which _greatly_ simplifies the algorithm. We also merge Tab<sub>1</sub> and Tab<sub>2</sub> (which is mentioned in the paper) and use perfect hashing for Tab<sub>3</sub> (which is not mentioned).

daira:
Draft PR (it doesn't work yet): https://github.com/zcash/halo2/pull/120

PR #120, opened 2021-01-06
Add sqrt_ratio implementation.

daira:
Based on [[Sarkar2020]](https://eprint.iacr.org/2020/1407) and [[BDLSY2012]](https://cr.yp.to/papers.html#ed25519).
The API is also based on https://ristretto.group/formulas/invsqrt.html . fixes zcash/halo2#72

Co-authored-by: Jack Grigg <jack@electriccoin.co>

### DIFF
--- a/src/arithmetic/fields.rs
+++ b/src/arithmetic/fields.rs
@@ -54,6 +54,11 @@ pub trait FieldExt:
     /// The choice of root from sqrt is unspecified.
     fn sqrt_ratio(num: &Self, div: &Self) -> (Choice, Self);
 
+    /// Equivalent to sqrt_ratio(self, one()).
+    fn sqrt_alt(&self) -> (Choice, Self) {
+        Self::sqrt_ratio(self, &Self::one())
+    }
+
     /// This computes a random element of the field using system randomness.
     fn rand() -> Self {
         Self::random(rand::rngs::OsRng)
@@ -259,13 +264,35 @@ impl<F: FieldExt> SqrtTables<F> {
         // uv = u * v
         let uv = w * num;
 
-        self.sqrt_common(num, div, &uv, &v)
+        let res = self.sqrt_common(&uv, &v);
+
+        let sqdiv = res.square() * div;
+        let is_square = (sqdiv - num).ct_is_zero();
+        let is_nonsquare = (sqdiv - F::ROOT_OF_UNITY * num).ct_is_zero();
+        assert!(bool::from(
+            num.ct_is_zero() | div.ct_is_zero() | (is_square ^ is_nonsquare)
+        ));
+
+        (is_square, res)
     }
 
-    /// Same as sqrt_ratio but given num, div, v = u^((T-1)/2), and uv = u * v as input.
-    ///
-    /// The choice of root from sqrt is unspecified.
-    fn sqrt_common(&self, num: &F, div: &F, uv: &F, v: &F) -> (Choice, F) {
+    /// Same as sqrt_ratio(u, one()) but more efficient.
+    pub fn sqrt_alt(&self, u: &F) -> (Choice, F) {
+        let v = u.pow_by_t_minus1_over2();
+        let uv = *u * v;
+
+        let res = self.sqrt_common(&uv, &v);
+
+        let sq = res.square();
+        let is_square = (sq - u).ct_is_zero();
+        let is_nonsquare = (sq - F::ROOT_OF_UNITY * u).ct_is_zero();
+        assert!(bool::from(u.ct_is_zero() | (is_square ^ is_nonsquare)));
+
+        (is_square, res)
+    }
+
+    /// Common part of sqrt_ratio and sqrt_alt: return res given v = u^((T-1)/2) and uv = u * v.
+    fn sqrt_common(&self, uv: &F, v: &F) -> F {
         let sqr = |x: F, i: u32| (0..i).fold(x, |x, _| x.square());
         let inv = |x: F| self.inv[self.hasher.hash(&x)] as usize;
 
@@ -296,20 +323,11 @@ impl<F: FieldExt> SqrtTables<F> {
                                 // 1 == x3 * ROOT_OF_UNITY^t_
         t_ = (t_ + 1) >> 1;
         assert!(t_ <= 0x80000000);
-        let res = *uv
-            * self.g0[t_ & 0xFF]
+
+        *uv * self.g0[t_ & 0xFF]
             * self.g1[(t_ >> 8) & 0xFF]
             * self.g2[(t_ >> 16) & 0xFF]
-            * self.g3[t_ >> 24];
-
-        let sqdiv = res.square() * div;
-        let is_square = (sqdiv - num).ct_is_zero();
-        let is_nonsquare = (sqdiv - F::ROOT_OF_UNITY * num).ct_is_zero();
-        assert!(bool::from(
-            num.ct_is_zero() | div.ct_is_zero() | (is_square ^ is_nonsquare)
-        ));
-
-        (is_square, res)
+            * self.g3[t_ >> 24]
     }
 }
 
--- a/src/arithmetic/fields.rs
+++ b/src/arithmetic/fields.rs
@@ -5,6 +5,7 @@ use core::mem::size_of;
 use static_assertions::const_assert;
 use std::assert;
 use std::convert::TryInto;
+use std::marker::PhantomData;
 use subtle::{Choice, ConstantTimeEq, CtOption};
 
 use super::Group;
@@ -41,14 +42,17 @@ pub trait FieldExt:
     /// Element of multiplicative order $3$.
     const ZETA: Self;
 
-    /// XOR parameter of the perfect hash function used for SqrtTables.
-    const HASH_XOR: u32;
-
-    /// Modulus of the perfect hash function used for SqrtTables.
-    const HASH_MOD: usize;
-
-    /// Tables for square root computation.
-    fn get_tables() -> &'static SqrtTables<Self>;
+    /// Computes:
+    ///
+    /// * (true,  sqrt(num/div)),                 if num and div are nonzero and num/div is a square in the field;
+    /// * (true,  0),                             if num is zero;
+    /// * (false, 0),                             if num is nonzero and div is zero;
+    /// * (false, sqrt(ROOT_OF_UNITY * num/div)), if num and div are nonzero and num/div is a nonsquare in the field;
+    ///
+    /// where ROOT_OF_UNITY is a generator of the order 2^n subgroup (and therefore a nonsquare).
+    ///
+    /// The choice of root from sqrt is unspecified.
+    fn sqrt_ratio(num: &Self, div: &Self) -> (Choice, Self);
 
     /// This computes a random element of the field using system randomness.
     fn rand() -> Self {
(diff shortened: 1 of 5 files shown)


src/arithmetic/fields.rs
```diff
@@ -41,6 +50,15 @@ pub trait FieldExt:
...
     const ZETA: Self;
 
+    /// XOR parameter of the perfect hash function used for SqrtTables.
+    const HASH_XOR: u32;
+
+    /// Modulus of the perfect hash function used for SqrtTables.
+    const HASH_MOD: usize;
+
+    /// Tables for square root computation.
+    fn get_tables() -> &'static SqrtTables<Self>;
```
daira, line 60: Consider renaming to `get_sqrt_tables`.

src/arithmetic/fields.rs
```diff
@@ -85,6 +103,119 @@ pub trait FieldExt:
...
+        //
+        // then u and uv as above. The addition chains are given in
+        // https://github.com/zcash/pasta/blob/master/addchain_sqrt.py .
+        // The overall cost of this part is similar to a single full-width exponentiation,
+        // regardless of S.
+
+        let sqr = |x: Self, i: u32| (0..i).fold(x, |x, _| x.square());
+
+        // s = div^(2^S - 1)
+        let s = (0..5).fold(*div, |d: Self, i| sqr(d, 1 << i) * d);
```
daira, line 147:
```suggestion
        // s = div^(2^S - 1)
        let s = div.pow_by_2toS_minus1();
```
But @str4d suggests not doing this for now, because Pallas and Vesta have S = 32.

src/arithmetic/fields.rs
```diff
@@ -85,6 +103,119 @@ pub trait FieldExt:
...
+
+        // t == div^(2^(S+1) - 1)
+        let t = s.square() * div;
+
+        // TODO: replace this with an addition chain.
+        let w = ff::Field::pow_vartime(&(t * num), &Self::T_EXPONENT_MINUS1_OVER2) * s;
```
daira, line 153:
```suggestion
        let w = (t * num).pow_tminus1_over2() * s;
```

src/arithmetic/fields.rs
```diff
@@ -85,6 +103,119 @@ pub trait FieldExt:
...
+        let sqr = |x: Self, i: u32| (0..i).fold(x, |x, _| x.square());
+
+        let x3 = *uv * v;
+        let x2 = sqr(x3, 8);
+        let x1 = sqr(x2, 8);
+        let x0 = sqr(x1, 8);
+
+        // i = 0, 1
+        let mut t_: usize = tab.inv[x0.hash()] as usize; // = t >> 16
+                                                         // 1 == x0 * ROOT_OF_UNITY^(t_ << 24)
```
daira, line 178:
```suggestion
        let mut t_ = tab.inv[x0.hash()] as usize; // = t >> 16
                                                  // 1 == x0 * ROOT_OF_UNITY^(t_ << 24)
```

daira: @str4d suggests to have a method `tab.inv(x0)` here.

src/arithmetic/fields.rs
str4d: Yeah, I feel that these kinds of APIs would not belong in `ff::Field`, and thus should be discouraged in `FieldExt` (the latter is an internal proving ground, but ideally we would not need it). And that we probably shouldn't actually be providing a default implementation of `sqrt_div`, since it necessarily requires exposing implementation-specific details to downstream users.

Instead (as discussed in PR gardening today) I think it would make more sense to make the `SqrtTable` struct be `Sakhar2020Sqrt` or something, have its init function take in e.g. a perfect hashing function for the given field, and then have that struct provide this implementation (taking in any necessary addition chains). Then a field implementor who wants to use this algorithm can just plug the pieces together to implement `sqrt_div`.

daira: Done in 8350773aeaf6dd0012c0afdd9cd94b1075ddfc27 (I haven't included the addition chains yet, but they're easy to do from here).

daira:
Note that a "plain" `sqrt` can be implemented in terms of `sqrt_alt` like this:
```
    /// Return a square root of this field element in a CtOption iff it exists.
    fn sqrt(&self) -> CtOption<Self> {
        let (is_square, res) = self.sqrt_alt();
        CtOption::new(res, is_square)
    }
```
but I wanted to get feedback on this implementation before replacing the existing one.

daira:
This is likely to conflict (trivially) with #113, since it adds code near where that PR deletes code.

src/arithmetic/fields.rs
```diff
@@ -1,10 +1,17 @@
...
+use core::mem::size_of;
+use static_assertions::const_assert;
+use std::assert;
+use std::convert::TryInto;
+use std::marker::PhantomData;
 use subtle::{Choice, ConstantTimeEq, CtOption};
 
 use super::Group;
 
+const_assert!(size_of::<usize>() >= 4);
```
ebfull, line 13: Let's remove

```rust
// No goofy hardware please.
assert!(core::mem::size_of::<usize>() >= 4);
```

from `commitment.rs` since we're making this static now.

src/arithmetic/fields.rs
```diff
@@ -174,6 +211,197 @@ pub trait FieldExt:
...
+
+/// Tables used for square root computation.
+#[derive(Debug)]
+pub struct SqrtTables<F: FieldExt> {
+    hasher: SqrtHasher<F>,
+    inv: Vec<u8>,
+    g0: [F; 256],
+    g1: [F; 256],
+    g2: [F; 256],
+    g3: [F; 129],
```
ebfull, line 237: These should be at least `Box<[F; 256]>` to save on stack space; this looks like it'll consume 26 KB.

src/pasta/fields/fp.rs
```diff
@@ -753,6 +772,46 @@ impl FieldExt for Fp {
...
+        let rl = sqr(rk, 5) * r11;
+        let rm = sqr(rl, 4) * r111;
+        let rn = sqr(rm, 4) * r11;
+        let ro = sqr(rn, 6) * r1001;
+        let rp = sqr(ro, 5) * r1101;
+        let rq = sqr(rp, 4) * r11;
+        let rr = sqr(rq, 7) * r111;
+        let rs = sqr(rr, 3) * r11;
+        let rt = rs.square();
+        //assert!(rt == ff::Field::pow_vartime(&self, &Fp::T_EXPONENT_MINUS1_OVER2));
```
ebfull, line 812: Remove this or make it `debug_assert!()`.

src/pasta/fields/fq.rs
```diff
@@ -753,6 +772,46 @@ impl FieldExt for Fq {
...
+        let sl = sqr(sk, 4) * s1011;
+        let sm = sqr(sl, 4) * s1011;
+        let sn = sqr(sm, 5) * s11;
+        let so = sqr(sn, 4) * self;
+        let sp = sqr(so, 5) * s11;
+        let sq = sqr(sp, 4) * s111;
+        let sr = sqr(sq, 5) * s1011;
+        let ss = sqr(sr, 3) * self;
+        let st = sqr(ss, 4);
+        //assert!(st == ff::Field::pow_vartime(&self, &Fq::T_EXPONENT_MINUS1_OVER2));
```
ebfull, line 812: Ditto.

ebfull (changes requested):

daira:
This has been rebased on main, and does now replace the existing sqrt implementation.

src/arithmetic/fields.rs
```diff
@@ -103,6 +140,197 @@ pub trait FieldExt:
...
+    hash_xor: u32,
+    hash_mod: usize,
+    marker: PhantomData<F>,
+}
+
+impl<F: FieldExt> SqrtHasher<F> {
+    /// Returns a perfect hash of x for use with SqrtTables::inv.
+    fn hash(&self, x: &F) -> usize {
+        ((x.get_lower_32() ^ self.hash_xor) as usize) % self.hash_mod
+    }
```
str4d, line 155: Where does this come from? Add a reference or documentation.

src/arithmetic/fields.rs
```diff
@@ -103,6 +140,197 @@ pub trait FieldExt:
...
+    inv: Vec<u8>,
+    g0: Box<[F; 256]>,
+    g1: Box<[F; 256]>,
+    g2: Box<[F; 256]>,
+    g3: Box<[F; 129]>,
+}
+
+impl<F: FieldExt> SqrtTables<F> {
+    /// Build tables given parameters for the perfect hash.
+    pub fn new(hash_xor: u32, hash_mod: usize) -> Self {
```
str4d, line 171: I have not checked this function yet.

src/arithmetic/fields.rs
```diff
@@ -103,6 +140,197 @@ pub trait FieldExt:
...
+        let sq = res.square();
+        let is_square = (sq - u).ct_is_zero();
+        let is_nonsquare = (sq - F::ROOT_OF_UNITY * u).ct_is_zero();
+        assert!(bool::from(u.ct_is_zero() | (is_square ^ is_nonsquare)));
+
+        (is_square, res)
+    }
+
+    /// Common part of sqrt_ratio and sqrt_alt: return res given v = u^((T-1)/2) and uv = u * v.
+    fn sqrt_common(&self, uv: &F, v: &F) -> F {
```
str4d, line 295: I have not checked this function yet.

src/pasta/fields/fp.rs
```diff
@@ -635,6 +599,10 @@ impl ff::PrimeField for Fp {
     }
 }
 
+lazy_static! {
+    static ref FP_TABLES: SqrtTables<Fp> = SqrtTables::new(0x11BE, 1098);
```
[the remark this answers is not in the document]
daira:
```suggestion
    // The perfect hash parameters are found by `squareroottab.sage` in zcash/pasta.
    static ref FP_TABLES: SqrtTables<Fp> = SqrtTables::new(0x11BE, 1098);
```

src/pasta/fields/fq.rs
```diff
@@ -635,6 +599,10 @@ impl ff::PrimeField for Fq {
     }
 }
 
+lazy_static! {
+    static ref FQ_TABLES: SqrtTables<Fq> = SqrtTables::new(0x116A9E, 1206);
```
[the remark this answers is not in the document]
daira:
```suggestion
    // The perfect hash parameters are found by `squareroottab.sage` in zcash/pasta.
    static ref FQ_TABLES: SqrtTables<Fq> = SqrtTables::new(0x116A9E, 1206);
```

src/arithmetic/fields.rs
```diff
@@ -103,6 +140,197 @@ pub trait FieldExt:
...
+#[derive(Debug)]
+struct SqrtHasher<F: FieldExt> {
+    hash_xor: u32,
+    hash_mod: usize,
+    marker: PhantomData<F>,
+}
+
+impl<F: FieldExt> SqrtHasher<F> {
+    /// Returns a perfect hash of x for use with SqrtTables::inv.
+    fn hash(&self, x: &F) -> usize {
```
daira, line 153:
```suggestion
    fn hash(&self, x: &F) -> usize {
        // This is just the simplest perfect hash construction that could possibly work.
        // The 32 low-order bits are unique within the 2^S order subgroup, then the xor
        // acts as an injective randomization function.
```

src/arithmetic/fields.rs
str4d: I checked that this function matches `SqrtField.sarkar_sqrt_common()` in [`squareroottab.sage`](https://github.com/zcash/pasta/blob/master/squareroottab.sage).

str4d (approved):
I'm satisfied that this correctly implements `squareroottab.sage`. I have _not_ checked that either this or `squareroottab.sage` corresponds to a square-root algorithm, because there isn't enough documentation for me to link table generation or `sqrt_common` to the paper.

daira:
The last part of the implementation of `sqrt_ratio` and `sqrt_alt` check that the output squares to the input or the input times `ROOT_OF_UNITY`, so it will either output a correct answer or panic. That said, I'll expand the comments to explain the correspondence to the paper.

src/arithmetic/fields.rs
```diff
@@ -103,6 +140,201 @@ pub trait FieldExt:
...
+            inv[hash] = ((256 - j) & 0xFF) as u8;
+        }
+
+        SqrtTables::<F> {
+            hasher,
+            inv,
+            g0: Box::new(gtab[0][..].try_into().unwrap()),
+            g1: Box::new(gtab[1][..].try_into().unwrap()),
+            g2: Box::new(gtab[2][..].try_into().unwrap()),
+            g3: Box::new(gtab[3][0..129].try_into().unwrap()),
```
ebfull, line 212: This will temporarily occupy a large portion of the stack before Rust actually copies it into the heap, which is annoying but can be mitigated by doing e.g. `gtab[0].into_boxed_slice()` though you will lose static slice lengths without using `unsafe` to restore them.

I suspect that when we bring the pasta impl out into a separate crate eventually we will need to feature flag this entire sqrt implementation so that users who are trying to compile this for embedded devices can turn off these massive precomputed tables. So, for now, let's leave this code alone.

ebfull (approved):

ebfull:
**Please don't forget to change the target branch of this PR to `zcash/main` instead of `daira/sqrt_ratio`**

src/arithmetic/fields.rs
daira: I considered using `into_boxed_slice`, but it's not worth the unsafety.

I'd rather just reduce the table size than switch off this implementation for embedded devices. With 16-element tables it's still much more efficient than Tonelli-Shanks.

PR #135, opened 2021-01-12
Serialize params

ebfull:


src/arithmetic/curves.rs
```diff
@@ -153,10 +155,27 @@ pub trait CurveAffine:
     /// endian representation.
     fn from_bytes(bytes: &[u8; 32]) -> CtOption<Self>;
 
+    /// Reads a compressed element from the buffer and attempts to parse it
+    /// using `from_bytes`.
+    fn read<R: Read>(reader: &mut R) -> io::Result<Self> {
+        let mut compressed = [0u8; 32];
+        reader.read_exact(&mut compressed[..])?;
+        Option::from(Self::from_bytes(&compressed)).ok_or(io::Error::new(
```
str4d, line 163: Clippy lint fix:
```suggestion
        Option::from(Self::from_bytes(&compressed)).ok_or_else(|| io::Error::new(
```

src/arithmetic/fields.rs
```diff
@@ -50,10 +52,27 @@ pub trait FieldExt:
...
     /// Attempts to obtain a field element from its normalized, little endian
     /// byte representation.
     fn from_bytes(bytes: &[u8; 32]) -> CtOption<Self>;
 
+    /// Reads a normalized, little endian represented field element from a
+    /// buffer.
+    fn read<R: Read>(reader: &mut R) -> io::Result<Self> {
+        let mut compressed = [0u8; 32];
+        reader.read_exact(&mut compressed[..])?;
+        Option::from(Self::from_bytes(&compressed)).ok_or(io::Error::new(
```
str4d, line 70: Clippy lint fix:
```suggestion
        Option::from(Self::from_bytes(&compressed)).ok_or_else(|| io::Error::new(
```

src/poly/commitment.rs
```diff
@@ -188,6 +190,55 @@ impl<C: CurveAffine> Params<C> {
...
+    pub fn write<W: io::Write>(&self, writer: &mut W) -> io::Result<()> {
+        writer.write_all(&self.k.to_le_bytes())?;
+        writer.write_all(&self.n.to_le_bytes())?;
+        for g_element in &self.g {
+            writer.write_all(&g_element.to_bytes())?;
+        }
+        for g_lagrange_element in &self.g_lagrange {
+            writer.write_all(&g_lagrange_element.to_bytes())?;
+        }
+        writer.write(&self.h.to_bytes())?;
```
str4d, line 204:
```suggestion
        writer.write_all(&self.h.to_bytes())?;
```

src/poly/commitment.rs
```diff
@@ -188,6 +190,55 @@ impl<C: CurveAffine> Params<C> {
...
+        writer.write(&self.u.to_bytes())?;
```
str4d, line 205:
```suggestion
        writer.write_all(&self.u.to_bytes())?;
```

str4d (changes requested):

src/plonk/keygen.rs
```diff
@@ -130,48 +132,90 @@ where
...
     let permutation_helper = permutation::keygen::Assembly::build_helper(params, &cs, &domain);
 
-    let (permutation_pks, permutation_vks) = cs
+    let permutation_vks = cs
         .permutations
         .iter()
-        .zip(assembly.permutations.into_iter())
-        .map(|(p, assembly)| assembly.build_keys(params, &domain, &permutation_helper, p))
-        .unzip();
+        .zip(assembly.clone().permutations.into_iter())
```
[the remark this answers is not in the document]
therealyingtong: It wasn't necessary!

src/poly/commitment.rs
```diff
@@ -188,6 +190,48 @@ impl<C: CurveAffine> Params<C> {
     pub fn get_g(&self) -> Vec<C> {
         self.g.clone()
     }
+
+    /// Writes params to a buffer.
+    pub fn write<W: io::Write>(&self, writer: &mut W) -> io::Result<()> {
+        writer.write_all(&self.k.to_le_bytes())?;
+        writer.write_all(&self.n.to_le_bytes())?;
```
str4d, line 197: I don't think we technically need both, since we assume `n = 2^k` in various parts of the codebase, but it's fine to include both.

src/plonk/permutation/keygen.rs
```diff
@@ -14,7 +14,7 @@ pub(crate) struct AssemblyHelper<C: CurveAffine> {
     deltaomega: Vec<Vec<C::Scalar>>,
 }
 
-#[derive(Debug)]
+#[derive(Clone, Debug)]
```
str4d, line 17: Since this is no longer used, we can remove it from this diff:
```suggestion
#[derive(Debug)]
```

str4d (approved):

src/plonk/keygen.rs
```diff
@@ -114,48 +132,90 @@ where
...
+        permutations: vk
+            .cs
+            .permutations
+            .iter()
+            .map(|p| permutation::keygen::Assembly::new(params.n as usize, p))
+            .collect(),
+        _marker: std::marker::PhantomData,
+    };
+
+    // Synthesize the circuit to obtain SRS
```
daira, line 180:
```suggestion
    // Synthesize the circuit to obtain URS
```

src/plonk/permutation/keygen.rs
```diff
@@ -132,19 +132,15 @@ impl Assembly {
...
         self,
         params: &Params<C>,
         domain: &EvaluationDomain<C::Scalar>,
         helper: &AssemblyHelper<C>,
         p: &Argument,
-    ) -> (ProvingKey<C>, VerifyingKey<C>) {
-        // Compute permutation polynomials, convert to coset form and
-        // pre-compute commitments for the SRS.
+    ) -> VerifyingKey<C> {
+        // Pre-compute commitments for the SRS.
```
daira, line 142:
```suggestion
        // Pre-compute commitments for the URS.
```

src/poly/commitment.rs
daira: I disagree; I think we should just write one of `k` or `n`, and calculate the other when reading. (Otherwise we should have an assertion that they're consistent when reading, which is no easier than calculating.)

daira (approved):
utACK

src/poly/commitment.rs
therealyingtong: We now write only `k`, and calculate `n` when reading.

ebfull:
Needs another rebase.

daira (approved):
re-utACK

str4d (approved):
re-utACK ffdd739f859b89ea412ef4ad844f332baea8410b

PR #136, opened 2021-01-13
Hash to curve

daira:
Add an implementation of simplified SWU. refs zcash/zcash#4710
This builds on sqrt_ratio added in #120.

It should match the Sage implementation at https://github.com/zcash/pasta/blob/master/hashtocurve.sage .

### DIFF
--- a/src/arithmetic/curves.rs
+++ b/src/arithmetic/curves.rs
@@ -11,7 +11,7 @@ use super::{FieldExt, Group};
 use std::io::{self, Read, Write};
 
 /// This trait is a common interface for dealing with elements of an elliptic
-/// curve group in the "projective" form, where that arithmetic is usually more
+/// curve group in a "projective" form, where that arithmetic is usually more
 /// efficient.
 pub trait Curve:
     Sized
@@ -60,7 +60,7 @@ pub trait Curve:
     /// Obtains the additive identity.
     fn zero() -> Self;
 
-    /// Obtains the base point of the curve.
+    /// Obtains the base point of the curve, if defined.
     fn one() -> Self;
 
     /// Doubles this element.
@@ -76,6 +76,9 @@ pub trait Curve:
     /// Converts this element into its affine form.
     fn to_affine(&self) -> Self::Affine;
 
+    /// Return the Jacobian coordinates of this point.
+    fn jacobian_coordinates(&self) -> (Self::Base, Self::Base, Self::Base);
+
     /// Returns whether or not this element is on the curve; should
     /// always be true unless an "unchecked" API was used.
     fn is_on_curve(&self) -> Choice;
@@ -84,8 +87,15 @@ pub trait Curve:
     /// sizes of the slices are different.
     fn batch_to_affine(v: &[Self], target: &mut [Self::Affine]);
 
-    /// Returns the curve constant b
+    /// Returns the curve constant a.
+    fn a() -> Self::Base;
+
+    /// Returns the curve constant b.
     fn b() -> Self::Base;
+
+    /// Obtains a point given Jacobian coordinates $X : Y : Z$, failing
+    /// if the coordinates are not on the curve.
+    fn new_jacobian(x: Self::Base, y: Self::Base, z: Self::Base) -> CtOption<Self>;
 }
 
 /// This trait is the affine counterpart to `Curve` and is used for
@@ -128,10 +138,13 @@ pub trait CurveAffine:
     /// random string.
     const BLAKE2B_PERSONALIZATION: &'static [u8; 16];
 
+    /// CURVE_ID used for hash-to-curve.
+    const CURVE_ID: &'static str;
+
     /// Obtains the additive identity.
     fn zero() -> Self;
 
-    /// Obtains the base point of the curve.
+    /// Obtains the base point of the curve, if defined.
     fn one() -> Self;
 
     /// Returns whether or not this element is the identity.
@@ -182,6 +195,9 @@ pub trait CurveAffine:
     /// element.
     fn to_bytes_wide(&self) -> [u8; 64];
 
-    /// Returns the curve constant $b$
+    /// Returns the curve constant $a$.
+    fn a() -> Self::Base;
+
+    /// Returns the curve constant $b$.
     fn b() -> Self::Base;
 }
--- a/src/arithmetic/curves.rs
+++ b/src/arithmetic/curves.rs
@@ -80,9 +80,21 @@ pub trait Curve:
     fn jacobian_coordinates(&self) -> (Self::Base, Self::Base, Self::Base);
 
     /// Requests a hasher that accepts messages and returns near-uniformly
-    /// distributed elements in the group, given domain prefix `hasher`.
+    /// distributed elements in the group, given domain prefix `domain_prefix`.
     ///
     /// This method is suitable for use as a random oracle.
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// use halo2::arithmetic::{Curve, CurveAffine};
+    /// fn pedersen_commitment<C: CurveAffine>(x: C::Scalar, r: C::Scalar) -> C {
+    ///     let hasher = C::Projective::hash_to_curve("z.cash:example_pedersen_commitment");
+    ///     let g = hasher(b"g");
+    ///     let h = hasher(b"h");
+    ///     (g * x + h * r).to_affine()
+    /// }
+    /// ```
     fn hash_to_curve(domain_prefix: &str) -> Box<dyn Fn(&[u8]) -> Self + 'static>;
 
     /// Returns whether or not this element is on the curve; should
--- a/src/arithmetic/curves.rs
+++ b/src/arithmetic/curves.rs
@@ -92,7 +92,7 @@ pub trait Curve:
     ///     let hasher = C::Projective::hash_to_curve("z.cash:example_pedersen_commitment");
     ///     let g = hasher(b"g");
     ///     let h = hasher(b"h");
-    ///     (g * x + h * r).to_affine()
+    ///     (g * x + &(h * r)).to_affine()
     /// }
     /// ```
     fn hash_to_curve(domain_prefix: &str) -> Box<dyn Fn(&[u8]) -> Self + 'static>;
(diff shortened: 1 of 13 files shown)


str4d:
Needs a rebase now that #120 has been merged.

str4d:
Rebased on `main`.

str4d:
@daira it looks like this PR is missing the `benches/hashtocurve.rs` file from your local workspace?

daira:
> @daira it looks like this PR is missing the `benches/hashtocurve.rs` file from your local workspace?

Fixed.

src/arithmetic/hashtocurve.rs
```diff
@@ -0,0 +1,330 @@
...
+
+        // 6. gx2 = x2^3 + A * x2 + B  [optimized out; see below]
+        // 7. If is_square(gx1), set x = x1 and y = sqrt(gx1)
+        // 8. Else set x = x2 and y = sqrt(gx2)
+        let (gx1_square, y1) = F::sqrt_ratio(&num_gx1, &div_gx1);
+
+        // This magic also comes from a generalization of [WB2019, section 4.2].
+        //
+        // The Sarkar square root algorithm with input s gives us a square root of
+        // ROOT_OF_UNITY * s for free when s is not square, where h is a fixed nonsquare.
```
daira, line 271:
```suggestion
        // ROOT_OF_UNITY * s for free when s is not square, where h is a fixed nonsquare.
```

src/arithmetic/hashtocurve.rs
```diff
@@ -0,0 +1,330 @@
...
+        // We know that Z / ROOT_OF_UNITY is a square since both Z and ROOT_OF_UNITY are
```
daira, line 272:
```suggestion
        // We know that Z / h is a square since both Z and h are
```

src/arithmetic/hashtocurve.rs
```diff
@@ -0,0 +1,330 @@
...
+
+/// The simplified SWU hash-to-curve method, using an isogenous curve
+/// y^2 = x^3 + a*x + b. This currently only supports prime-order curves.
+#[derive(Debug)]
+pub struct SimplifiedSWUWithDegree3Isogeny<
+    F: FieldExt,
+    I: CurveAffine<Base = F>,
+    C: CurveAffine<Base = F>,
+> {
+    /// `Z` parameter (ξ in [WB2019]).
```
therealyingtong, line 171:
```suggestion
    /// `Z` parameter (ξ in [WB2019](https://eprint.iacr.org/2019/403)).
```

src/arithmetic/hashtocurve.rs
```diff
@@ -0,0 +1,330 @@
...
+    /// field elements.
+    fn hash_to_field(&self, message: &[u8], domain_separation_tag: &[u8], count: usize) -> Vec<F>;
+}
+
+/// A MessageHasher for SHAKE128
+/// [FIPS202](https://nvlpubs.nist.gov/nistpubs/FIPS/NIST.FIPS.202.pdf).
+/// It does not support domain separation tags longer than 128 bytes.
+#[derive(Debug, Default)]
+pub struct Shake128<F: FieldExt> {
+    marker: PhantomData<F>,
```
therealyingtong, line 108:
```suggestion
    _marker: PhantomData<F>,
```

src/arithmetic/hashtocurve.rs
```diff
@@ -0,0 +1,330 @@
...
+}
+
+impl<F: FieldExt> MessageHasher<F> for Blake2bXof<F> {
+    fn hash_name(&self) -> &str {
+        "XOF:BLAKE2b"
+    }
+
+    #[allow(unused_variables)]
+    fn hash_to_field(&self, message: &[u8], domain_separation_tag: &[u8], count: usize) -> Vec<F> {
+        todo!()
```
ebfull, line 159: The entire `Blake2bXof` should be removed from this PR, or this should be finished and the SHAKE128 thing should be removed -- we don't need multiple instances of hashers in the codebase right now.

ebfull (changes requested):

src/arithmetic/hashtocurve.rs
daira: I included Shake128 because I was originally thinking that would make it easier to compare to [the secp256k1 test vectors from the Internet Draft](https://www.ietf.org/archive/id/draft-irtf-cfrg-hash-to-curve-10.html#section-appendix.j), but I hadn't noticed that those only have XMD:SHA-256. I do want to have more confidence that we've implemented the same thing as the draft; I'll think about how to do that.

src/arithmetic/hashtocurve.rs
```diff
@@ -0,0 +1,330 @@
...
+    }
+}
+
```
daira, line 162:
```suggestion
```

daira, line 162:
```suggestion
```

src/arithmetic/hashtocurve.rs
daira: I'm going to switch the hasher to XMD:BLAKE2b.

src/pasta/hashtocurve.rs
```diff
@@ -0,0 +1,145 @@
...
+    let div_y = (((x + iso[10] * z2) * x + iso[11] * z4) * x + iso[12] * z6) * z3;
+
+    let zo = div_x * div_y;
+    let xo = num_x * div_y * zo;
+    let yo = num_y * div_x * zo.square();
+
+    C::Projective::new_jacobian(xo, yo, zo).unwrap()
+}
+
+pub fn map_to_curve<F: FieldExt, C: CurveAffine<Base = F>, I: CurveAffine<Base = F>>(
```
therealyingtong, line 65: should we rename this `map_to_curve_simple_swu` instead? as per https://www.ietf.org/archive/id/draft-irtf-cfrg-hash-to-curve-10.html#name-simplified-swu-for-ab-0-2

daira:
This now just needs the test vectors to be updated.

src/pasta/hashtocurve.rs
```diff
@@ -1,32 +1,64 @@
 //! This module implements "simplified SWU" hashing to short Weierstrass curves
 //! with a = 0.
 
+use std::convert::TryInto;
 use subtle::ConstantTimeEq;
 
 use crate::arithmetic::{Curve, CurveAffine, FieldExt};
 
 /// Hashes over a message and writes the output to all of `buf`.
 pub fn hash_to_field<F: FieldExt>(message: &[u8], domain_separation_tag: &[u8], buf: &mut [F]) {
```
ebfull, line 10: Does this function now no longer support filling the `buf` with random elements no matter its size? (My understanding is that is what the previous code did.) If now it only supports writing two elements to `buf`, then `buf` should change to `&mut [F; 2]`.

src/arithmetic/curves.rs
```diff
@@ -60,7 +60,7 @@ pub trait Curve:
     /// Obtains the additive identity.
     fn zero() -> Self;
 
-    /// Obtains the base point of the curve.
+    /// Obtains the base point of the curve, if defined.
```
ebfull, line 63: Let's revert this back to the original comment. The "if defined" contradicts the API contract in the group crate (which we aren't using yet but will) and since we're not exposing the isogeneous curve in the API it doesn't matter anyway.

src/pasta/fields/fp.rs
```diff
@@ -16,7 +16,7 @@ use crate::arithmetic::{adc, mac, sbb, FieldExt, Group, SqrtTables};
 // The internal representation of this type is four 64-bit unsigned
 // integers in little-endian order. `Fp` values are always in
 // Montgomery form; i.e., Fp(a) = aR mod p, with R = 2^256.
-#[derive(Clone, Copy, Eq)]
+#[derive(Clone, Copy, Eq, Hash)]
```
ebfull, line 19: Why is this `Hash` derive introduced? It trips up clippy and I don't think it's used in this PR.

src/pasta/fields/fq.rs
```diff
@@ -16,7 +16,7 @@ use crate::arithmetic::{adc, mac, sbb, FieldExt, Group, SqrtTables};
 // The internal representation of this type is four 64-bit unsigned
 // integers in little-endian order. `Fq` values are always in
 // Montgomery form; i.e., Fq(a) = aR mod q, with R = 2^256.
-#[derive(Clone, Copy, Eq)]
+#[derive(Clone, Copy, Eq, Hash)]
```
ebfull, line 19: Ditto

src/arithmetic/curves.rs
```diff
@@ -128,10 +156,13 @@ pub trait CurveAffine:
...
     const BLAKE2B_PERSONALIZATION: &'static [u8; 16];
 
+    /// CURVE_ID used for hash-to-curve.
+    const CURVE_ID: &'static str;
+
     /// Obtains the additive identity.
     fn zero() -> Self;
 
-    /// Obtains the base point of the curve.
+    /// Obtains the base point of the curve, if defined.
```
ebfull, line 165: Ditto

ebfull (changes requested):

src/pasta/fields/fp.rs
therealyingtong: this seems to be used in `pasta::curves`, `impl Hash for $name_affine`, but that in turn doesn't seem to be used anywhere.

daira: It is used:
```
   --> src/pasta/curves.rs:656:24
    |
656 |                   self.y.hash(state);
```

src/arithmetic/curves.rs
daira:
```suggestion
    /// Obtains the base point of the curve.
```

src/arithmetic/curves.rs
daira:
```suggestion
    /// Obtains the base point of the curve.
```

src/pasta/fields/fp.rs
daira: It was used in a previous iteration. I've removed it.

ebfull (approved):

Issue #93, opened 2020-12-21
Write a simple first-fit FloorPlanner


daira:
Input: a set of regions, each of which takes up a given set of columns `region.cols` for a given number of rows `region.rows`.
Output: a mapping from regions to their start rows.

Greedy first fit algorithm (sketch):

For each column, keep track of the set of allocated [start, end) intervals so far.
We greedily find a placement for each region in turn (there is no backtracking at the region level):
* Start with the interval of all possible rows. For each column used by this region, recursively narrow the interval in which we could potentially place the region. We do this by intersecting the current interval with the free intervals in each used column. If it won't fit, backtrack to the next free interval.

It is probably easier to explain this just by giving a Python implementation:
```python
#!/usr/bin/env python3

from math import inf

class Region:
    def __init__(self, rows, cols):
        self.rows = rows
        self.cols = cols

    def __repr__(self):
        return "Region(%d, %r)" % (self.rows, self.cols)

class Column:
    def __init__(self, index):
        self.index = index
        # allocs is an unordered list of [start, end) pairs.
        self.allocs = []

    def __repr__(self):
        return "Column(%d): %r" % (self.index, self.allocs)

def plan(regions):
    for reg in regions:
        start = first_fit_region(reg, 0, 0, inf)

        # Some placement must be possible given infinite rows.
        assert start is not None
        print("%4r: %r" % (start, reg))

def first_fit_region(reg, col_index, start, end):
    assert start + reg.rows <= end
    if col_index == len(reg.cols):
        return start

    c = reg.cols[col_index]

    # Iterate over the unallocated nonempty intervals in c that intersect [start, end).
    for (c_start, c_end) in free_intervals(c.allocs, start, end):
        # Do we have enough room for this column of reg?
        if c_start + reg.rows <= c_end:
            # Recursively check whether the placement works for all other columns of reg.
            row = first_fit_region(reg, col_index+1, c_start, c_end)
            if row is not None:
                # It works. On the way back out of the recursion, record the placement.
                assert row + reg.rows <= c_end
                c.allocs.append((row, row + reg.rows))
                return row

    # No placement worked; the caller will need to try other possibilities.
    return None

# allocs is a set of [a_start, a_end) pairs representing disjoint allocated intervals.
# Return all the *unallocated* nonempty intervals intersecting [start, end).
def free_intervals(allocs, start, end):
    row = start
    for (a_start, a_end) in sorted(allocs):
        if a_start >= end:
            # further allocated intervals don't intersect
[...]

PR #137, opened 2021-01-13
Simple layouter strategy to track column usage

therealyingtong:
We implement the simplest approach here: position the region starting at the earliest row for which none of the columns are in use.

### DIFF
--- a/src/circuit.rs
+++ b/src/circuit.rs
@@ -138,6 +138,6 @@ pub trait Layouter<C: Chip> {
     /// ```
     fn assign_region(
         &mut self,
-        assignment: impl FnOnce(Region<'_, C>) -> Result<(), Error>,
+        assignment: impl FnMut(Region<'_, C>) -> Result<(), Error>,
     ) -> Result<(), Error>;
 }
--- a/src/circuit/layouter.rs
+++ b/src/circuit/layouter.rs
@@ -138,20 +138,38 @@ impl<'a, C: Chip, CS: Assignment<C::Field> + 'a> Layouter<C> for SingleChip<'a, 
     }
 }
 
+/// The shape of a region. For a region at a certain index, we track
+/// the set of columns it uses as well as the number of rows it uses.
 #[derive(Debug)]
-struct RegionShape {
+pub struct RegionShape {
     region_index: usize,
     columns: HashSet<Column<Any>>,
     row_count: usize,
 }
 
 impl RegionShape {
-    fn new(region_index: usize) -> Self {
+    /// Create a new `RegionShape` for a region at `region_index`.
+    pub fn new(region_index: usize) -> Self {
         RegionShape {
             region_index,
             columns: HashSet::default(),
             row_count: 0,
         }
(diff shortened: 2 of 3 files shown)


src/circuit.rs
```diff
@@ -138,6 +138,6 @@ pub trait Layouter<C: Chip> {
     /// ```
     fn assign_region(
         &mut self,
-        assignment: impl FnOnce(Region<'_, C>) -> Result<(), Error>,
+        assignment: impl FnMut(Region<'_, C>) -> Result<(), Error>,
```
str4d, line 141: This becomes `FnMut` because we call it more than once. Note that we maintain `Region`'s invariant that the inner `to: FnMut`s passed into `Region::assign_advice` and `Region::assign_fixed` are only ever called once.

src/plonk/circuit.rs
```diff
@@ -11,7 +11,7 @@ use crate::poly::Rotation;
 pub trait ColumnType: 'static + Sized {}
 
 /// A column with an index and type
-#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
```
str4d, line 14: `Hash` is derived here so we can use columns with a `HashSet` or as keys in a `HashMap`.

src/circuit/layouter.rs
```diff
@@ -160,7 +238,6 @@ impl<'r, 'a, C: Chip, CS: Assignment<C::Field> + 'a> RegionLayouter<C>
             self.layouter.regions[self.region_index] + offset,
             to,
         )?;
-        self.row_count = cmp::max(self.row_count, offset);
```
str4d, line 163: Note that this PR also fixes an off-by-one error that existed here (which was causing the last row of a region to be overwritten by the next region).

src/circuit/layouter.rs
```diff
@@ -102,26 +104,122 @@ impl<'a, C: Chip, CS: Assignment<C::Field> + 'a> Layouter<C> for SingleChip<'a,
...
 
         Ok(())
     }
 }
 
+/// The shape of a region. For a region at a certain index, we track
+/// the set of columns it uses as well as the number of rows it uses.
+#[derive(Debug)]
+pub struct RegionShape {
+    region_index: usize,
```
daira, line 145: The concept I had of a region shape (e.g. for #93) didn't include the starting index, but I guess that can be changed later if needed.

daira (approved):
utACK

src/circuit/layouter.rs
str4d: This isn't the row that the region starts on; it's a "region ID" within the layouter, used to enable `Cell`s to be pointers to regions instead of to absolute circuit positions (which directly enables the kind of layouting that this PR implements). `RegionShape` stores `region_index` solely to ensure that any `Cell`s it returns are consistent across both the shape-finding pass, and the assignment pass.

PR #145, opened 2021-01-22
Circuit annotations and developer tooling

### LINKED ISSUE #124: Implement namespacing
In `bellman` we have a nice feature for debugging where circuits and gadgets can create internal namespaces to label the circuit structure. It would be really nice for `halo2` to support similar functionality; this would both help `MockProver` to provide better debugging feedback, and enable us to render nice views of chips.

str4d:
This adds support for:
- Gate naming
- User namespacing of gadgets
  - As-implemented, I think this also can just keep on namespacing through into instructions until a region boundary, which may or may not be desirable.
- Gadget name extraction (behind the `gadget-traces` feature flag)
- Region naming
- Cell annotations
- Developer tooling to:
  - render a circuit's namespace as a dot graph.
  - render a circuit as a table.

Closes #124.
Part of #157.

### DIFF
--- a/examples/circuit-layout.rs
+++ b/examples/circuit-layout.rs
@@ -1,0 +1,391 @@
+use halo2::{
+    arithmetic::FieldExt,
+    dev::circuit_layout,
+    pasta::Fp,
+    plonk::{Advice, Assignment, Circuit, Column, ConstraintSystem, Error, Fixed},
+    poly::Rotation,
+};
+use plotters::prelude::*;
+use std::marker::PhantomData;
+
+fn main() {
+    /// This represents an advice column at a certain row in the ConstraintSystem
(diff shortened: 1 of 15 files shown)


examples/circuit-layout.rs
```diff
@@ -0,0 +1,391 @@
...
+            }
+            Ok(())
+        }
+    }
+
+    impl<F: FieldExt> Circuit<F> for MyCircuit<F> {
+        type Config = PLONKConfig;
+
+        fn configure(meta: &mut ConstraintSystem<F>) -> PLONKConfig {
+            let e = meta.advice_column();
```
daira, line 250: We should turn off the "bindings with single-character names in scope" lint.

daira (approved):
utACK

PR #149, opened 2021-01-27
Replace DummyHash with BLAKE2b

### LINKED ISSUE #148: Change transcript to use BLAKE2b
Currently we use `DummyHash{Read,Write}` which is a fake algebraic hash function in `transcript.rs`. However, for orchard (which won't be using recursion _yet_) we should switch to BLAKE2b for security purposes and then later work to write a hasher implementation that uses Poseidon/Rescue/whatever.

therealyingtong:
Closes #148

### DIFF
--- a/src/transcript.rs
+++ b/src/transcript.rs
@@ -1,7 +1,9 @@
 //! This module contains utilities and traits for dealing with Fiat-Shamir
 //! transcripts.
 
+use blake2b_simd::{Params as Blake2bParams, State as Blake2bState};
 use ff::Field;
+use std::convert::TryInto;
 use std::ops::Deref;
 
 use crate::arithmetic::{CurveAffine, FieldExt};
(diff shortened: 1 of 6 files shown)


src/transcript.rs
```diff
@@ -39,30 +41,36 @@ pub trait TranscriptWrite<C: CurveAffine>: Transcript<C> {
...
+    pub fn init(reader: R) -> Self {
+        Blake2bRead {
+            base_state: Blake2bParams::new()
+                .hash_length(32)
+                .personal(C::BLAKE2B_PERSONALIZATION)
+                .to_state(),
+            scalar_state: Blake2bParams::new()
+                .hash_length(32)
+                .personal(C::BLAKE2B_PERSONALIZATION)
+                .to_state(),
```
ebfull, line 65: This impl can be simplified by having only one instances of a BLAKE2b hasher involved and hashing both the scalars and the group elements into the same hasher. The reason why there were separate hashers in the old implementation is that it was loosely describing an algebraic hash where we will indeed be forced to have separate hashers because they're separate fields.

src/transcript.rs
```diff
@@ -39,30 +41,36 @@ pub trait TranscriptWrite<C: CurveAffine>: Transcript<C> {
...
+impl<R: Read, C: CurveAffine> Blake2bRead<R, C> {
     /// Initialize a transcript given an input buffer and a key.
-    pub fn init(reader: R, key: C::Base) -> Self {
-        DummyHashRead {
-            base_state: key + &C::Base::from_u64(1013),
-            scalar_state: C::Scalar::from_u64(1013),
+    pub fn init(reader: R) -> Self {
+        Blake2bRead {
+            base_state: Blake2bParams::new()
+                .hash_length(32)
```
ebfull, line 59: This should be configured for hash_length 64.

src/transcript.rs
```diff
@@ -77,75 +85,89 @@ impl<R: Read, C: CurveAffine> TranscriptRead<C> for DummyHashRead<R, C> {
...
-        self.base_state = self.base_state.square();
-        self.base_state += &(y * &C::Base::ZETA);
-        self.base_state = self.base_state.square();
+        self.base_state.update(&x.to_bytes());
+        self.base_state.update(&y.to_bytes());
 
         Ok(())
     }
 
     fn squeeze_challenge(&mut self) -> C::Base {
```
ebfull, line 115: The way this method should work should be a bit simpler (especially after you get rid of `scalar_state`). Something like this:

```rust
let mut hasher = self.base_state.clone();
let result: [u8; 64] = hasher.finalize().as_bytes().try_into().unwrap(); 
self.base_state.update(&result[..]);
return C::Base::from_bytes_wide(&result);
```

`from_bytes_wide` allows you to avoid the loop thing while keeping the value well-distributed.

ebfull (changes requested):

PR #153, opened 2021-01-28
[WIP] Add gadgets for elliptic curve operations

str4d:
Required ECC instructions:
- [ ] inputize public point
- [ ] witness private point
- [ ] load fixed point
- [x] add
- [x] double
- [ ] fixed-base mul
- [x] var-base mul

### DIFF
--- a/src/gadget/ecc.rs
+++ b/src/gadget/ecc.rs
@@ -3,13 +3,13 @@
 use std::fmt;
 
 use crate::{
-    arithmetic::Curve,
+    arithmetic::CurveAffine,
     circuit::{Chip, Layouter},
     plonk::Error,
 };
 
 /// The set of circuit instructions required to use the ECC gadgets.
-pub trait EccInstructions<C: Curve>: Chip<Field = C::Base> {
+pub trait EccInstructions<C: CurveAffine>: Chip<Field = C::Base> {
     /// Variable representing an elliptic curve point.
     type Point: Clone + fmt::Debug;
     /// Variable representing an element of the elliptic curve's scalar field.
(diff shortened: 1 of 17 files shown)


src/gadget/ecc.rs
```diff
@@ -0,0 +1,67 @@
...
+use std::fmt;
+
+use crate::{
+    arithmetic::Curve,
+    circuit::{Chip, Layouter},
+    plonk::Error,
+};
+
+/// The set of circuit instructions required to use the ECC gadgets.
+pub trait EccInstructions<C: Curve>: Chip<Field = C::Base> {
```
str4d, line 12: I think `C` here needs to be a type parameter rather than an associated type, because it's _possible_ (though unlikely) that a chip may want to implement `EccInstructions` for multiple curves built over the same base field.

src/gadget/ecc.rs
```diff
@@ -0,0 +1,67 @@
...
+
+/// An element of the given elliptic curve's scalar field.
+#[derive(Debug)]
+pub struct Scalar<C: Curve, EccChip: EccInstructions<C>> {
+    inner: EccChip::Scalar,
+}
+
+/// An elliptic curve point over the given curve.
+#[derive(Debug)]
+pub struct Point<C: Curve, EccChip: EccInstructions<C>> {
```
str4d, line 44: Making `C` a type parameter of `EccInstructions` has the nice side-effect that the `Point` gadget is typed to a specific curve.

src/gadget/ecc.rs
daira: Actually that's not so unlikely, e.g. it's needed for simplified SWU (to implement the isogenous curve).

src/gadget/ecc.rs
```diff
@@ -0,0 +1,67 @@
...
+    ) -> Result<Self::Point, Error>;
+
+    /// Performs point doubling, returning `[2] a`.
+    fn double(layouter: &mut impl Layouter<Self>, a: &Self::Point) -> Result<Self::Point, Error>;
+
+    /// Performs variable-base scalar multiplication, returning `[scalar] base`.
+    fn mul(
+        layouter: &mut impl Layouter<Self>,
+        base: &Self::Point,
+        scalar: &Self::Scalar,
```
daira, line 32: Consider swapping these so that they're in the same order as the usual (additive) notation.

str4d: I did that to begin with, and then thought I might prefer it the other way around 😅 I'll swap it back.

PR #154, opened 2021-01-29
Arity-3 Poseidon chip

str4d:


daira:
I think we might actually need arity 4: the encoding of (ak, nk) is three field elements (ak is an affine curve point and nk is a field element), plus one for the sponge capacity.

PR #158, opened 2021-01-30
Improvements to background section

daira:
* describe groups in general, rather than via the example of F_p^*
* explain the isomorphism between group elements and scalars, and how it is useful
* corrections and pedantry :-)


book/src/background/curves.md
```diff
@@ -19,25 +23,48 @@ handling: we find the line tangent to the point, and then find the single other
...
+is a prime $q$, then the scalars can be considered as elements of a _scalar field_,
+$\mathbb{F}_q$.
 
 Elliptic curves, when properly designed, have an important security property. Given two
 random elements $G, H \in E(\mathbb{F}_p)$ finding $a$ such that $[a] G = H$, otherwise
 known as the discrete log of $H$ with respect to $G$, is considered computationally
 infeasible with classical computers. This is called the elliptic curve discrete log
 assumption.
 
+If an elliptic curve group $\mathbb{G}$ has prime order $q$ (like the ones used in Halo),
```
str4d, line 37: "used in Halo 2"? (since I assume you're referring to this crate)

book/src/background/fields.md
```diff
@@ -37,32 +37,33 @@ Important notes:
...
 
-Any non-zero element $a \in \mathbb{F}_p$ has a _multiplicative inverse_ $b = a^{-1}$,
-which is the _unique_ element $b$ such that $ab = 1$.
+## Groups
 
-A quick way of obtaining the inverse is $a^{-1} = a^{p - 2}$. The reason for this stems
-from [Fermat's little theorem][fermat-little], which states that $a^p = a \pmod p$ for any
-integer $a$. If $a$ is nonzero, we can divide by $a$ twice to get $a^{p-2} = a^{-1}.$
+Groups are simpler and more limited than fields; they have only one binary operator $\cdot$
+and fewer axioms. They also have an identity, which we'll denote as $1$.
```
str4d, line 47:
```suggestion
and fewer axioms. They also have an identity, which we'll denote as $\mathcal{O}$.
```
(since 1 collides with the isomorphism-to-integers discussion later, and this is consistent with that section)

book/src/background/fields.md
```diff
@@ -37,32 +37,33 @@ Important notes:
...
 
-[fermat-little]: https://en.wikipedia.org/wiki/Fermat%27s_little_theorem
+[group]: https://en.wikipedia.org/wiki/Group_(mathematics)
+[group-axioms]: https://en.wikipedia.org/wiki/Group_(mathematics)#Definition
+
+Any non-zero element $a$ in a group has an _inverse_ $b = a^{-1}$,
+which is the _unique_ element $b$ such that $a \cdot b = 1$.
```
str4d, line 53:
```suggestion
which is the _unique_ element $b$ such that $a \cdot b = \mathcal{O}$.
```

book/src/background/fields.md
```diff
@@ -77,20 +78,53 @@ have only _one_ operator $\cdot$ and fewer axioms.
 > scalars to negative integers by inversion, i.e. $[-k] A + [k] A = \mathcal{O}$ or
 > $a^{-k} \times a^k = 1$.
 
-The _order_ of a group element $a$ is defined (in multiplicative notation) as the smallest
-positive integer $k$ such that $a^k = 1$.
+The _order_ of an element $a$ of a finite group is defined (in multiplicative notation)
+as the smallest positive integer $k$ such that $a^k = 1$. The order _of the group_ is
```
str4d, line 82:
```suggestion
as the smallest positive integer $k$ such that $a^k = \mathcal{O}$. The order _of the group_ is
```

book/src/background/curves.md
```diff
@@ -1,6 +1,10 @@
 # Elliptic curves
 
 Elliptic curves constructed over finite fields are another important cryptographic tool.
+
+We use elliptic curves because they provide a cryptographic [group](fields.md#Inverses_and_groups),
+i.e. a group in which the [discrete logarithm problem](fields#) is hard.
```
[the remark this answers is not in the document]
daira: Huh, I thought I'd fixed that. *Checks whether everything was committed.*

book/src/background/curves.md
daira:
```suggestion
If an elliptic curve group $\mathbb{G}$ has prime order $q$ (like the ones used in Halo 2),
```

book/src/background/fields.md
daira: No this would be more confusing. The standard notation for the additive and multiplicative identities of a field are 0 and 1 respectively. We only use \mathcal{O} for the identity of a group that is written additively.

I don't agree that it collides with the isomorphism-to-integers discussion. The sides of the isomorphism are different structures and could each be written with any notation.

book/src/background/fields.md
daira: Rejected, see above.

book/src/background/fields.md
daira: Rejected, see above.

book/src/background/fields.md
daira: Also, although we *can* choose any notation, I think we *should* avoid having \mathcal{O} in an additively written group and \mathcal{O} in the scalar field that are *not* isomorphic (because the group identity corresponds to 0 in the scalar field, not 1).

book/src/background/curves.md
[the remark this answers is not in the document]
str4d: This still isn't fixed.

str4d (approved):
ACK modulo link.

book/src/background/curves.md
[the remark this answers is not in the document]
daira:
```suggestion
i.e. a group in which the discrete logarithm problem (discussed below) is hard.
```

PR #163, opened 2021-02-02
Helpers for implementing regions

str4d:


src/plonk/circuit.rs
```diff
@@ -1,10 +1,13 @@
 use core::cmp::max;
 use core::ops::{Add, Mul};
 use ff::Field;
-use std::convert::TryFrom;
+use std::{
+    convert::TryFrom,
```
ebfull, line 5: Where is this used in the code that is introduced?

src/plonk/circuit.rs
```diff
@@ -1,10 +1,13 @@
...
+    ops::{Neg, Sub},
+};
 
 use super::{lookup, permutation, Error};
-use crate::poly::Rotation;
+use crate::{arithmetic::FieldExt, poly::Rotation};
```
ebfull, line 10: Where is `FieldExt` needed?

ebfull (changes requested):

src/plonk/circuit.rs
ebfull: nvm misread the code change

src/plonk/circuit.rs
str4d: It's used in the type bounds of the `Neg` and `Sub` impls for `Expression<F>`, since we need the ability to define `-F::one()`.

str4d: @ebfull clarified that he meant this could just be `F: Field`.

ebfull (approved):

PR #164, opened 2021-02-02
Implement Sinsemilla

### LINKED ISSUE #73: Implement Sinsemilla gadget

therealyingtong:
Closes #73.

TODOs:
- [ ] Decide how to constrain `(x_q, y_q)` (the point used to initialise the hash) in the circuit. We could introduce a custom gate but that's rather a waste of a fixed column.
- [ ] Decide on personalisation input to generate `Q`.

### DIFF
--- a/src/gadget/sinsemilla.rs
+++ b/src/gadget/sinsemilla.rs
@@ -10,6 +10,9 @@ use crate::{
     plonk::Error,
 };
 
+mod chip;
+pub use chip::{SinsemillaChip, SinsemillaColumns, SinsemillaConfig};
+
 /// Domain prefix used in SWU hash-to-curve to generate S_i's.
 pub const S_DOMAIN_PREFIX: &str = "z.cash:SinsemillaS";
 
--- a/src/gadget/sinsemilla.rs
+++ b/src/gadget/sinsemilla.rs
@@ -1,0 +1,41 @@
+//! Gadget and chips for the [Sinsemilla] hash function.
+//!
+//! [Sinsemilla]: https://hackmd.io/iOw7-HpFQY6dPF1aFY8pAw
+
+use std::fmt;
+
+use crate::{
+    arithmetic::FieldExt,
+    circuit::{Chip, Layouter},
+    plonk::Error,
+};
+
+/// Domain prefix used in SWU hash-to-curve to generate S_i's.
+pub const S_DOMAIN_PREFIX: &str = "z.cash:SinsemillaS";
+
+/// Domain prefix used in SWU hash-to-curve to generate Q.
+pub const Q_DOMAIN_PREFIX: &str = "z.cash:SinsemillaQ";
+
+/// Personalization input used to generate Q
+/// TODO: Decide on personalization
+pub const Q_PERSONALIZATION: [u8; 4] = [0u8; 4];
+
+/// The set of circuit instructions required to use the [`Sinsemilla`](https://zcash.github.io/halo2/design/gadgets/sinsemilla.html) gadget.
+pub trait SinsemillaInstructions<F: FieldExt>: Chip<Field = F> {
+    /// A message of at most `kn` bits.
+    type Message: Clone + fmt::Debug;
+    /// The output of `Hash`.
+    type HashOutput: fmt::Debug;
+
+    /// Return Q
+    fn q() -> (F, F);
+
+    /// Hashes the given message.
+    ///
+    /// TODO: Since the output is always curve point, maybe this should return
+    /// `<Self as EccInstructions>::Point` instead of an associated type.
+    fn hash(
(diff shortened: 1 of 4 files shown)


src/gadget/sinsemilla.rs
```diff
@@ -0,0 +1,311 @@
...
+
+    /// This represents an advice column at a certain row in the ConstraintSystem
+    #[derive(Copy, Clone, Debug)]
+    pub struct Variable(Column<Advice>, usize);
+
+    struct MyCircuit<F: FieldExt, C: CurveAffine<Base = F>> {
+        message: Vec<bool>,
+        _marker_f: PhantomData<F>,
+        _marker_c: PhantomData<C>,
+    }
```
nuttycom, line 68:
```suggestion
    struct MyCircuit<C: CurveAffine> {
        message: Vec<bool>,
        _marker: PhantomData<C>,
    }
```
We can always just use C::Base to refer to the field element type, and don't need to constrain it here.

src/gadget/sinsemilla.rs
```diff
@@ -0,0 +1,311 @@
...
+
+    struct MyLayouter<'a, C: Chip, CS: Assignment<C::Field> + 'a> {
+        cs: &'a mut CS,
+        config: C::Config,
+        regions: Vec<usize>,
```
nuttycom, line 73: Can we newtype this `usize` to make it clearer what this represents?

src/gadget/sinsemilla/chip.rs
```diff
@@ -0,0 +1,240 @@
...
+    x_p: Column<Advice>,
+    lookup_table: GeneratorTable,
+}
+
+/// A chip implementing SinsemillaInstructions
+#[derive(Debug)]
+pub struct SinsemillaChip<F: FieldExt, C: CurveAffine<Base = F>> {
+    _marker_f: PhantomData<F>,
+    _marker_c: PhantomData<C>,
+}
```
nuttycom, line 35:
```suggestion
pub struct SinsemillaChip<C: CurveAffine> {
    _marker: PhantomData<C>,
}
```

src/gadget/sinsemilla/chip.rs
```diff
@@ -0,0 +1,240 @@
...
+
+        // y_a = (1/2) ⋅ (lambda1 + lambda2) ⋅ (x_a - (lambda1^2 - x_a - x_p))
+        let lambda1_cur = meta.query_advice(lambda1, Rotation::cur());
+        let lambda2_cur = meta.query_advice(lambda2, Rotation::cur());
+        let x_a_cur = meta.query_advice(x_a, Rotation::cur());
+        let x_p_cur = meta.query_advice(x_p, Rotation::cur());
+        let y_a_cur = (lambda1_cur.clone() + lambda2_cur.clone())
+            * (x_a_cur.clone()
+                - (lambda1_cur.clone() * lambda1_cur.clone() - x_a_cur.clone() - x_p_cur.clone()))
+            * F::from_u64(2).invert().unwrap();
```
nuttycom, line 64:
```suggestion
            * F::TWO_INV;
```

src/gadget/sinsemilla/chip.rs
```diff
@@ -0,0 +1,240 @@
...
+        let x_a_next = meta.query_advice(x_a, Rotation::next());
+        let x_p_next = meta.query_advice(x_p, Rotation::next());
+        let y_a_next = (lambda1_next.clone() + lambda2_next)
+            * (x_a_next.clone()
+                - (lambda1_next.clone() * lambda1_next - x_a_next.clone() - x_p_next))
+            * F::from_u64(2).invert().unwrap();
+
+        // Sinsemilla gate
+        meta.create_gate("Sinsemilla", |_| {
+            // λ_{2,i}^2 − x_{A,i+1} −(λ_{1,i}^2 − x_{A,i} − x_{P,i}) − x_{A,i} = 0
```
nuttycom, line 93: From pairing with @ebfull:

These need to be two separate gates, rather than combining them into one, because they need to both be equal to zero independently. Otherwise, one term from one of the expressions might be able to cancel out a nonzero result from the other.

nuttycom (changes requested):

src/gadget/sinsemilla.rs
therealyingtong: This `usize` represents the starting row of the region. Opened #203 to newtype this.

str4d: Instead of defining anothger `MyLayouter` here, use `circuit::layouter::SingleChip`. (I made the same changes to the `sha-256` chip impl.)

therealyingtong:
Superseded by https://github.com/zcash/orchard/pull/24 and https://github.com/zcash/orchard/pull/25.

PR #167, opened 2021-02-05
[book] Add Sinsemilla gadget and chip description

therealyingtong:


therealyingtong:
This spec is likely change significantly. Let's add it to the book when it's closer to being finalised.

PR #172, opened 2021-02-08
Arity-3 Poseidon

str4d:
TODO:
- [x] Round constant and MDS generation
- [ ] Implement logic to check for and reject weak MDS matrices.
- [x] Chip implementing arity-3 Poseidon permutation.
- [ ] Gadget for Poseidon hash function.


str4d:
Replaced by https://github.com/zcash/orchard/pull/28.

PR #173, opened 2021-02-09
Hash verification key into transcript

### LINKED ISSUE #171: Feed hash of verification key into the transcript
We haven't done this yet, but now that we've nailed the basic design of the internal PLONK implementation I think we're ready to start.

You _need_ to feed the verification key (or a binding commitment to it) into the transcript at the beginning of the interaction with the prover. If you don't, then challenges only depend on the prover's messages which can be computed _without_ a description of the circuit. This would allow someone creating a circuit to place a backdoor by precomputing a challenge and then adaptively constructing the circuit so that soundness is broken for that particular challenge.

Strictly speaking all common inputs should be part of the transcript, including the URS and any other public params, but there's probably no practical attacks that exist without breaking discrete log.

therealyingtong:
Closes #171


str4d (approved):
utACK

ebfull (changes requested):
We need to hash vector lengths in to prevent ambiguity.

We also need to hash in many other parts of the verification key and public parameters than this; literally any degree of freedom available to circuit builders.

This includes:

* `vk.domain.omega`, `vk.domain.k`, `vk.domain.extended_k`, `vk.quotient_poly_degree` (the other constants in the domain don't influence the proof)
* `vk.cs` -- the circuit arrangement (expressions, etc.) can contain constants and these could be backdoored as well

src/plonk.rs
```diff
@@ -74,6 +77,72 @@ impl<C: CurveAffine> VerifyingKey<C> {
             cs,
         })
     }
+
+    /// Hashes a verification key into a transcript.
+    pub fn hash<T: Transcript<C>>(&self, transcript: &mut T) -> io::Result<()> {
+        let try_and_increment = |hasher: &Blake2bState| {
```
ebfull, line 83: Ah, let's just add a `common_scalar` to the `Transcript` trait, and refactor the transcript implementations to use it. Then there's no need to sample curve points.

ebfull: Added this in #179.

src/poly/domain.rs
```diff
@@ -376,4 +379,20 @@ impl<G: Group> EvaluationDomain<G> {
     pub fn get_quotient_poly_degree(&self) -> usize {
         self.quotient_poly_degree as usize
     }
+
+    /// Hashes the constants in the domain which influence the proof into a u64
+    pub fn hash(&self) -> u64 {
```
ebfull, line 384: This isn't going to work; the `DefaultHasher` in the standard library is not guaranteed to have a stable algorithm. You need to take in a mutable reference to a BLAKE2b hasher as input and update it with the constants.

Also, we can't hash things down into `u64`s because it would lead to trivial collisions.

ebfull (changes requested):
I think the way we should do the hashing should be as follows:

make a single BLAKE2b `Hasher` instance and pass it around to all the `hash` methods, and take the result of that (64 bytes) and turn it into a scalar (via from_bytes_wide), then use the `common_scalar` API that [just merged](https://github.com/zcash/halo2/pull/179) in order to add that scalar to the transcript.

src/plonk.rs
```diff
@@ -74,6 +77,43 @@ impl<C: CurveAffine> VerifyingKey<C> {
...
+        hasher.update(b"num_permutations");
+        hasher.update(&self.permutations.len().to_le_bytes());
+        for permutation in &self.permutations {
+            permutation.hash(transcript)?;
+        }
+
+        // Hash in final Blake2bState
+        transcript.common_scalar(C::Scalar::from_bytes_wide(
+            hasher.finalize().as_bytes().try_into().unwrap(),
+        ))?;
```
ebfull, line 113: This is the thing we want to do just once at the very end. All the other `common_scalar` operations are unnecessary.

src/plonk/circuit.rs
```diff
@@ -594,4 +606,70 @@ impl<F: Field> ConstraintSystem<F> {
         self.num_aux_columns += 1;
         tmp
     }
+
+    /// Hashes the `ConstraintSystem` into a `u64`.
+    pub fn hash(&self, mut hasher: &mut Blake2bState) -> [u8; 64] {
```
ebfull, line 611: Do not return a hash, just update the hasher that was provided. Also, this binding does not need to be mutable so it can just be `hasher: &mut Blake2bState`.

src/plonk/permutation.rs
```diff
@@ -66,6 +67,14 @@ impl<C: CurveAffine> VerifyingKey<C> {
             .collect::<Result<Vec<_>, _>>()?;
         Ok(VerifyingKey { commitments })
     }
+
+    pub(crate) fn hash<T: Transcript<C>>(&self, transcript: &mut T) -> io::Result<()> {
+        for commitment in &self.commitments {
```
ebfull, line 72: Hash in a length here for the vector.

src/poly.rs
```diff
@@ -228,4 +230,10 @@ impl Rotation {
     pub fn next() -> Rotation {
         Rotation(1)
     }
+
+    /// Hash Rotation into a Blake2bState
+    pub fn hash(&self, hasher: &mut Blake2bState) -> [u8; 64] {
```
ebfull, line 235: Do not return a hash, just update the hasher.

src/plonk/circuit.rs
```diff
@@ -594,4 +606,70 @@ impl<F: Field> ConstraintSystem<F> {
...
+            for column in argument.get_columns().iter() {
+                column.hash(&mut hasher);
+            }
+        }
+
+        hasher.update(b"num_lookups");
+        hasher.update(&self.lookups.len().to_le_bytes());
+        for argument in self.lookups.iter() {
+            hasher.update(&argument.input_columns.len().to_le_bytes());
+            hasher.update(&argument.table_columns.len().to_le_bytes());
```
ebfull, line 662: These should be the same, you can just hash in one of them and put an assertion here that the lengths are the same.

src/plonk/circuit.rs
```diff
@@ -317,6 +323,12 @@ impl<F: Field> Expression<F> {
             Expression::Scaled(poly, _) => poly.degree(),
         }
     }
+
+    /// Hash an Expression into a Blake2bState
+    pub fn hash(&self, hasher: &mut Blake2bState) -> [u8; 64] {
```
ebfull, line 328: Don't return a hash, just update hasher.

src/plonk/circuit.rs
```diff
@@ -27,6 +28,11 @@ impl<C: ColumnType> Column<C> {
     pub(crate) fn column_type(&self) -> &C {
         &self.column_type
     }
+
+    pub(crate) fn hash(&self, hasher: &mut Blake2bState) -> [u8; 64] {
+        hasher.update(&format!("{:?}", self).as_bytes());
+        hasher.finalize().as_bytes().try_into().unwrap()
+    }
```
ebfull, line 35: This function shouldn't return a hash, it should just update the hasher it is given.

src/plonk/circuit.rs
```diff
@@ -594,4 +606,70 @@ impl<F: Field> ConstraintSystem<F> {
...
+        hasher.update(b"num_advice_columns");
+        hasher.update(&self.num_advice_columns.to_le_bytes());
+
+        hasher.update(b"num_aux_columns");
+        hasher.update(&self.num_aux_columns.to_le_bytes());
+
+        hasher.update(b"num_gates");
+        hasher.update(&self.gates.len().to_le_bytes());
+        for gate in self.gates.iter() {
+            hasher.update(gate.0.to_owned().as_bytes());
```
ebfull, line 624: The name of the gate in code does not affect the argument, so this can be omitted.

ebfull (changes requested):

src/plonk/circuit.rs
therealyingtong: i do mutate it when hashing in `Expression`s, `Column`s and `Rotation`s.

src/plonk.rs
```diff
@@ -74,6 +77,41 @@ impl<C: CurveAffine> VerifyingKey<C> {
...
         })
     }
+
+    /// Hashes a verification key into a transcript.
+    pub fn hash_into<T: Transcript<C>>(&self, transcript: &mut T) -> io::Result<()> {
+        let mut hasher = Blake2bParams::new()
+            .hash_length(64)
+            .personal(C::BLAKE2B_PERSONALIZATION)
+            .to_state();
+
```
daira, line 87: Hash in the field modulus.

src/plonk/circuit.rs
```diff
@@ -27,6 +28,10 @@ impl<C: ColumnType> Column<C> {
     pub(crate) fn column_type(&self) -> &C {
         &self.column_type
     }
+
+    pub(crate) fn hash_into(&self, hasher: &mut Blake2bState) {
+        hasher.update(&format!("{:?}", self).as_bytes());
```
daira, line 33: This needs a length.

src/plonk/circuit.rs
```diff
@@ -317,6 +322,11 @@ impl<F: Field> Expression<F> {
             Expression::Scaled(poly, _) => poly.degree(),
         }
     }
+
+    /// Hash an Expression into a Blake2bState
+    pub fn hash_into(&self, hasher: &mut Blake2bState) {
+        hasher.update(&format!("{:?}", self).as_bytes());
```
daira, line 328: This needs a length.

src/plonk.rs
therealyingtong: I've chosen to hash in the field modulus in `domain.rs`: https://github.com/zcash/halo2/pull/173/files#diff-d87267170e09ce6a4d0d50d2af190351492d463599379ba46c6bad6b3733685bR384-R387

and the curve parameters below.

ebfull (approved):

str4d (changes requested):

src/plonk/circuit.rs
```diff
@@ -380,6 +380,36 @@ pub struct ConstraintSystem<F> {
     pub(crate) lookups: Vec<lookup::Argument>,
 }
 
+impl<F: PrimeField> std::fmt::Display for ConstraintSystem<F> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> Result<(), std::fmt::Error> {
+        write!(
+            f,
+            "ConstraintSystem {{\
```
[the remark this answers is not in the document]
ebfull: The main reason is that the `&'static str` name of the gate is not something we actually want to encode in the transcript; it isn't involved in proof generation and is just an annotation that we should be able to change freely.

src/plonk.rs
```diff
@@ -74,6 +77,48 @@ impl<C: CurveAffine> VerifyingKey<C> {
             cs,
         })
     }
+
+    /// Hashes a verification key into a transcript.
+    pub fn hash_into<T: Transcript<C>>(&self, transcript: &mut T) -> io::Result<()> {
+        let mut hasher = Blake2bParams::new()
+            .hash_length(64)
+            .personal(C::BLAKE2B_PERSONALIZATION)
```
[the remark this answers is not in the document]
ebfull: Yes, I was punting this decision off to another time. After the hashing to curve PR lands, we're going to remove `C::BLAKE2B_PERSONALIZATION` entirely, so we'll have to address it then and @daira had a suggested personalization.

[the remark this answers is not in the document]
daira: https://github.com/zcash/halo2/issues/182#issuecomment-778762162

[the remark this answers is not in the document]
ebfull: Not sure about that personalization @daira. Specifically what we're hashing here is the verification key, and we are then taking that hash and mapping it to a scalar that we put into the transcript. So for instance "Halo2-VerifyHash" would perhaps be better, and of course I'm sure there are even better options also.

src/plonk.rs
```diff
@@ -74,6 +77,48 @@ impl<C: CurveAffine> VerifyingKey<C> {
...
+        write!(
+            f,
+            "VerificationKey {{\
+            base_modulus = {base_modulus}, \
+            scalar_modulus = {scalar_modulus}, \
+            domain = {domain}, \
+            cs = {cs}, \
+            fixed_commitments = {fixed_commitments:?}, \
+            permutations = {permutations:?}\
+        }}",
```
daira:
```suggestion
            "VerificationKey {{\n"
            "  base_modulus = {base_modulus},\n"
            "  scalar_modulus = {scalar_modulus},\n"
            "  domain = {domain},\n"
            "  cs = {cs},\n"
            "  fixed_commitments = {fixed_commitments:?},\n"
            "  permutations = {permutations:?}\n"
            "}}",
```
Why the double braces, though?

str4d: I don't think we want newlines in this, because that adds even more dependency on the precise spaces. We should stick as closely as possible to what `#[derive(Debug)]` outputs, which is not pretty-printed.

src/plonk.rs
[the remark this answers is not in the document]
daira: Oh, I see. "Halo2-Verify-Key" then.

src/plonk.rs
str4d: (Clarification: `{:?}` does not pretty-print. `{:#?}` does.)

daira: OK, but what is written now *does* include spaces, since this way of doing multiline strings does not strip indentation.

src/plonk.rs
```diff
@@ -74,6 +76,46 @@ impl<C: CurveAffine> VerifyingKey<C> {
...
+        write!(
+            f,
+            "VerificationKey {{ \
+            base_modulus = {base_modulus}, \
+            scalar_modulus = {scalar_modulus}, \
+            domain = {domain}, \
+            cs = {cs}, \
+            fixed_commitments = {fixed_commitments:?}, \
+            permutations = {permutations:?}\
+        }}",
```
daira, line 110:
```suggestion
            "VerificationKey {{"
            "base_modulus = {base_modulus}, "
            "scalar_modulus = {scalar_modulus}, "
            "domain = {domain}, "
            "cs = {cs}, "
            "fixed_commitments = {fixed_commitments:?}, "
            "permutations = {permutations:?}"
            "}}",
```

src/plonk.rs
ebfull: Well, default Debug uses spaces AFAIK, just not newlines.

str4d: It _does_ strip spaces. You can see that in the test case.

str4d: https://github.com/zcash/halo2/pull/173/files#diff-ce29ac47e4ad4e869b74fd280b52b536e4967606470ae4f64261c50c32e4404fR567

src/plonk/circuit.rs
```diff
@@ -380,6 +380,36 @@ pub struct ConstraintSystem<F> {
...
+            "ConstraintSystem {{ \
+             num_fixed_columns: {num_fixed_columns}, \
+             num_advice_columns: {num_advice_columns}, \
+             num_instance_columns: {num_instance_columns}, \
+             fixed_queries: {fixed_queries:?}, \
+             advice_queries: {advice_queries:?}, \
+             instance_queries: {instance_queries:?}, \
+             permutations: {permutations:?}, \
+             lookups: {lookups:?}, \
+             gates: [",
```
daira, line 396:
```suggestion
            "ConstraintSystem {{"
            "num_fixed_columns: {num_fixed_columns}, "
            "num_advice_columns: {num_advice_columns}, "
            "num_instance_columns: {num_instance_columns}, "
            "fixed_queries: {fixed_queries:?}, "
            "advice_queries: {advice_queries:?}, "
            "instance_queries: {instance_queries:?}, "
            "permutations: {permutations:?}, "
            "lookups: {lookups:?}, "
            "gates: [",
```

src/plonk.rs
```diff
@@ -522,6 +564,7 @@ fn test_proving() {
...
...516c0f0bf), (0x37ead9904c760201ec4734ef398f0bdb5fe5a5e6e9db19c85e6b5483bdeb0a0b, 0x1dc08c38ed713b14f7a21a891a83b52160a3ffb0dccfbd70db7c7eb235dd193e), (0x2dc3d20553691216c988ecbb596c4bda329f27d50bd8a7c2fb0d84b423da3cb4, 0x025b40e800020458e15e3a57268562e6c08c6971d71262bd67c72437cfc60b4c)], permutations = [VerifyingKey { commitments: [(0x289f468bca3471a3d240169ec65047d0c4bb5e1135e81822523c74b596139fed, 0x1a585c821c71fb49c883859f1389bcae45f17593ddb5f9fee1781b27129e1b06), (0x096ef96a7725c636e7ca645dfe539694cf2a988da1ca50a468320f419d008054, 0x1ac0b48a254f8e2311081f81aa4a8ff39e37e40aa4c003325f1ac1219d231818), (0x254c9f0088599aba37607cfd1700c653ef3ec21bfd60b98c3d725678540bc4df, 0x134d9818929589052f3cd80d50140b851db26231b2b784a6b2528a64805598dc)] }, VerifyingKey { commitments: [(0x289f468bca3471a3d240169ec65047d0c4bb5e1135e81822523c74b596139fed, 0x1a585c821c71fb49c883859f1389bcae45f17593ddb5f9fee1781b27129e1b06), (0x096ef96a7725c636e7ca645dfe539694cf2a988da1ca50a468320f419d008054, 0x1ac0b48a254f8e2311081f81aa4a8ff39e37e40aa4c003325f1ac1219d231818), (0x254c9f0088599aba37607cfd1700c653ef3ec21bfd60b98c3d725678540bc4df, 0x134d9818929589052f3cd80d50140b851db26231b2b784a6b2528a64805598dc)] }]}");
```
str4d, line 567: This needs updating now that the spacing is fixed.

str4d (changes requested):

src/poly/domain.rs
```diff
@@ -377,3 +377,19 @@ impl<G: Group> EvaluationDomain<G> {
...
+
+impl<G: Group> std::fmt::Display for EvaluationDomain<G> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> Result<(), std::fmt::Error> {
+        write!(
+            f,
+            "Domain {{ \
+            k = {k:?}, \
+            extended_k = {extended_k:?}, \
+            omega = {omega:?} \
+        }}",
```
daira, line 389:
```suggestion
            "Domain {{"
            "k = {k:?}, "
            "extended_k = {extended_k:?}, "
            "omega = {omega:?}"
            "}}",
```

src/plonk.rs
str4d: @daira https://stackoverflow.com/a/29483453

daira: Oh, so it does.

src/plonk.rs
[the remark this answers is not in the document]
ebfull: Done.

ebfull (approved):

daira (approved):
ACK

PR #174, opened 2021-02-09
Add gadgets for elliptic curve operations

str4d:


src/gadget/ecc.rs
```diff
@@ -0,0 +1,146 @@
...
+    type Scalar: Clone + fmt::Debug;
+    /// Variable representing an elliptic curve point.
+    type Point: Clone + fmt::Debug;
+    /// Variable representing a fixed elliptic curve point (constant in the circuit).
+    type FixedPoint: Clone + fmt::Debug;
+
+    /// Witnesses the given scalar as a private input to the circuit.
+    fn witness_scalar(
+        layouter: &mut impl Layouter<Self>,
+        value: impl FnMut() -> Result<C::Scalar, Error>,
```
str4d, line 23: We don't actually want this to be a function; we need to optionally map over the value if it exists, and with a function we can only learn this by calling the function, making it redundant.
```suggestion
        value: Option<C::Scalar>,
```

src/gadget/ecc.rs
```diff
@@ -0,0 +1,146 @@
...
+    ) -> Result<Self::Scalar, Error>;
+
+    /// Witnesses the given point as a private input to the circuit.
+    fn witness_point(
+        layouter: &mut impl Layouter<Self>,
+        value: impl FnMut() -> Result<C, Error>,
+    ) -> Result<Self::Point, Error>;
+
+    /// Exposes the given point as a public input to the circuit.
+    fn expose_point(layouter: &mut impl Layouter<Self>, a: &Self::Point) -> Result<(), Error>;
```
str4d, line 33: Sean says let's just remove this for now.

src/gadget/ecc.rs
```diff
@@ -0,0 +1,146 @@
...
+    /// Witnesses the given scalar as a private input to the circuit.
+    fn witness_scalar(
+        layouter: &mut impl Layouter<Self>,
+        value: impl FnMut() -> Result<C::Scalar, Error>,
+    ) -> Result<Self::Scalar, Error>;
+
+    /// Witnesses the given point as a private input to the circuit.
+    fn witness_point(
+        layouter: &mut impl Layouter<Self>,
+        value: impl FnMut() -> Result<C, Error>,
```
str4d, line 29:
```suggestion
        value: Option<C>,
```

src/gadget/ecc.rs
```diff
@@ -0,0 +1,146 @@
...
+    ) -> Result<Self::Point, Error>;
+
+    /// Exposes the given point as a public input to the circuit.
+    fn expose_point(layouter: &mut impl Layouter<Self>, a: &Self::Point) -> Result<(), Error>;
+
+    /// Loads a fixed point into the circuit.
+    fn load_fixed(
+        layouter: &mut impl Layouter<Self>,
+        value: impl FnMut() -> Result<C, Error>,
```
str4d, line 38:
```suggestion
        value: Option<C>,
```

ebfull (approved):

PR #176, opened 2021-02-10
Generalise lookup argument to work over expressions

### LINKED ISSUE #175: Generalise lookup argument to work over expressions
Currently, our lookup argument checks that the values in a given set of input columns are contained in a corresponding set of table columns. Instead of comparing columns directly, we can generalise the argument to compare expressions (i.e. linear combinations of columns).

therealyingtong:
Closes #175

### DIFF
--- a/src/plonk.rs
+++ b/src/plonk.rs
@@ -472,7 +472,7 @@ fn test_proving() {
             let sl_ = meta.query_any(sl.into(), Rotation::cur());
             let sl2_ = meta.query_any(sl2.into(), Rotation::cur());
             meta.lookup(&[a_.clone()], &[sl_.clone()]);
-            meta.lookup(&[a_, b_], &[sl_, sl2_]);
+            meta.lookup(&[a_ + b_], &[sl_ + sl2_]);
 
             meta.create_gate("Combined add-mult", |meta| {
                 let d = meta.query_advice(d, Rotation::next());
--- a/src/plonk.rs
+++ b/src/plonk.rs
@@ -909,39 +909,37 @@ fn test_proving() {
         ],
         lookups: [
             Argument {
-                input_columns: [
-                    Column {
-                        index: 1,
-                        column_type: Advice,
-                    },
+                input_expressions: [
+                    Advice(
+                        0,
+                    ),
                 ],
-                table_columns: [
-                    Column {
-                        index: 6,
-                        column_type: Fixed,
-                    },
+                table_expressions: [
+                    Fixed(
+                        0,
+                    ),
                 ],
             },
             Argument {
-                input_columns: [
-                    Column {
-                        index: 1,
-                        column_type: Advice,
-                    },
-                    Column {
-                        index: 2,
-                        column_type: Advice,
-                    },
+                input_expressions: [
+                    Product(
+                        Advice(
+                            0,
+                        ),
+                        Advice(
+                            1,
+                        ),
+                    ),
                 ],
-                table_columns: [
-                    Column {
-                        index: 6,
-                        column_type: Fixed,
-                    },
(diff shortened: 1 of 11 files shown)


src/plonk/lookup/prover.rs
```diff
@@ -80,24 +84,44 @@ impl Argument {
...
-                .unzip();
+                .collect();
+
+            let unpermuted_cosets: Vec<_> = unpermuted_columns
+                .iter()
+                .map(|column| {
+                    let poly = pk.vk.domain.lagrange_to_coeff(column.clone());
+                    pk.vk.domain.coeff_to_extended(poly, Rotation::cur())
+                })
+                .collect();
```
therealyingtong, line 124: I'm not confident that this is correct.

src/plonk/lookup/prover.rs
```diff
@@ -76,28 +80,45 @@ impl Argument {
...
```
therealyingtong, line 121: I'm not confident that this is correct.

src/plonk.rs
```diff
@@ -423,8 +423,12 @@ fn test_proving() {
...
              * ]
              */
-            meta.lookup(&[a.into()], &[sl.into()]);
-            meta.lookup(&[a.into(), b.into()], &[sl.into(), sl2.into()]);
+            let a_ = meta.query_any(a.into(), Rotation::cur());
+            let b_ = meta.query_any(b.into(), Rotation::cur());
+            let sl_ = meta.query_any(sl.into(), Rotation::cur());
+            let sl2_ = meta.query_any(sl2.into(), Rotation::cur());
+            meta.lookup(&[a_.clone()], &[sl_.clone()]);
+            meta.lookup(&[a_ + b_], &[sl_ + sl2_]);
```
therealyingtong, line 431: `meta.lookup(&[a_ * b_], &[sl_ * sl2_])` does not work currently.

src/poly.rs
```diff
@@ -168,15 +168,10 @@ impl<'a, F: Field, B: Basis> Sub<&'a Polynomial<F, B>> for Polynomial<F, B> {
...
-    type Output = Polynomial<F, ExtendedLagrangeCoeff>;
-
-    fn mul(
-        mut self,
-        rhs: &'a Polynomial<F, ExtendedLagrangeCoeff>,
-    ) -> Polynomial<F, ExtendedLagrangeCoeff> {
+impl<'a, F: Field, B: Basis> Mul<&'a Polynomial<F, B>> for Polynomial<F, B> {
+    type Output = Polynomial<F, B>;
+
+    fn mul(mut self, rhs: &'a Polynomial<F, B>) -> Polynomial<F, B> {
```
therealyingtong, line 174: I'm not confident that this is correct.

therealyingtong: (it was incorrect)

src/plonk/lookup.rs
```diff
@@ -33,6 +34,15 @@ impl Argument {
...
+        let mut input_degree = 1;
+        for expr in self.input_expressions.iter() {
+            input_degree = std::cmp::max(input_degree, expr.degree());
+        }
+        let mut table_degree = 1;
+        for expr in self.table_expressions.iter() {
+            table_degree = std::cmp::max(table_degree, expr.degree());
+        }
+
+        1 + input_degree + table_degree
```
ebfull, line 46: Technically we could have expressions which are constants, and so `input_degree` and `table_degree` should start out as zero. If they did, this would need to change to `std::cmp::max(1 + input_degree + table_degree, 2)` where the 2 comes from the other gate expressions.

However, that would be weird and for now we can leave it like this.

ebfull (approved):

PR #177, opened 2021-02-11
[book] Fill out design section about proving system

str4d:


book/src/design/proving-system.md
```diff
@@ -1 +1,69 @@
...
+4. Construct the multipoint opening argument to check that all evaluations are consistent
+   with their respective commitments.
+5. Run the inner product argument to create a polynomial commitment opening proof for the
+   multipoint opening argument polynomial.
+
+These stages are presented in turn across this section of the book.
+
+## Example
+
+To aid our explanations, we will at times refer to the following example circuit:
```
ebfull, line 26: Let's call this a "constraint system" rather than a circuit. The distinction is useful because we will (in general) be trying to _embed_ arithmetic circuits into relations that involve constraint systems.

book/src/design/proving-system/circuit-commitments.md
```diff
@@ -0,0 +1,34 @@
+# Circuit commitments
+
+## Committing to the circuit assignments
+
+At the start of proof creation, the prover has a table of cell assignments that it claims
+satisfy the circuit. The table has $n = 2^k$ rows, and is broken into advice, auxiliary,
```
ebfull, line 6: constraint system

book/src/design/proving-system/circuit-commitments.md
```diff
@@ -0,0 +1,34 @@
...
+and fixed columns. We define $F_{i,j}$ as the assignment in the $j$th row of the $i$th
+fixed column. Without loss of generality, we'll similarly define $A_{i,j}$ to represent
+the advice and auxiliary assignments.
+
+> The only difference between advice and auxiliary columns, is that the commitments to
+> auxiliary columns are not placed in the proof, and are instead computed by the verifier.
```
ebfull, line 12: Fixed columns are also not placed in the proof, but are computed in a pre-processing phase (though they appear in the transcript so that challenges depend on them).

book/src/design/proving-system/circuit-commitments.md
```diff
@@ -0,0 +1,34 @@
...
+
+To commit to these assignments, we construct Lagrange polynomials of degree $n$ for each
```
ebfull, line 14: these polynomials end up being degree `n - 1` actually; the vanishing polynomial is of degree n though.

book/src/design/proving-system/vanishing.md
```diff
@@ -0,0 +1,73 @@
...
+
+$$h(X) = \frac{\text{gate}_0(X) + y \cdot \text{gate}_1(X) + \dots + y^i \cdot \text{gate}_i(X) + \dots}{t(X)},$$
+
+as a random (the prover commits to the cell assignments before the verifier samples $y$)
+linear combination of the circuit relations.
+
+- If this random linear combination is perfectly divisible by $t(X)$, then with high
+  probability all relations are satisfied.
+- Conversely, if at least one relation is not satisfied, then with high probability the
+  random linear combination of relations is not perfectly divisible by $t(X)$.
```
ebfull, line 36: It's with high probability that h(x) t(x) would not equal the evaluation of the numerator at x since this cannot hold for most values of x in the field. The probability that the _polynomial_ (in formal indeterminate X) is not perfectly divisible would in this case be 100%.

book/src/design/proving-system/vanishing.md
```diff
@@ -0,0 +1,73 @@
...
+
+For efficiency, the prover split $h(X)$ into degree-$n$ polynomials
```
ebfull, line 38: specifically, it's because our commitment scheme only supports committing to polynomials of degree `n - 1` so we must split it up. It is an efficiency detail in the sense that it wouldn't make sense for the polynomial commitment scheme to be designed to support committing to h(X) when the rest of the entire protocol only needs commitments to `n - 1` degree polynomials.

book/src/design/proving-system/vanishing.md
```diff
@@ -0,0 +1,73 @@
...
+
+$$h_0(X) + X^n h_1(X) + \dots + X^{n(d-1)} h_{d-1}(X),$$
+
+and produces blinding commitments
+
+$$\mathbf{H} = [\text{Commit}(h_0(X)), \text{Commit}(h_1(X)), \dots, \text{Commit}(h_{d-1}(X))].$$
+
+> TODO:
+> - Check pieces are correct.
+> - Explain why $h(X)$ has degree $n \cdot d$.
```
ebfull, line 48: Dividing by the vanishing polynomial decreases the degree by n, so h(X) will be slightly less degree than this. Also, we have to be careful to avoid off-by-one errors because for example a poly

`\frac{a(X) * b(X) - c(X)}{t(X)}`

is of degree `n - 2` because a(X)b(X) is degree `2n - 2`, and t(X) is degree n.

ebfull (changes requested):

book/src/design/proving-system/circuit-commitments.md
str4d: Yep, I make this note further down. I'll see if I can reorder things to make this clearer.

book/src/design/proving-system/circuit-commitments.md
```diff
@@ -0,0 +1,80 @@
...
+
+  $$A_\text{compressed}(X) = \theta^{m-1} A_0(X) + \theta^{m-2} A_1(X) + \dots + \theta A_{m-2}(X) + A_{m-1}(X)$$
+  $$S_\text{compressed}(X) = \theta^{m-1} S_0(X) + \theta^{m-2} S_1(X) + \dots + \theta S_{m-2}(X) + S_{m-1}(X)$$
+
+- The prover then permutes $A_\text{compressed}(X)$ and $S_\text{compressed}(X)$ according
+  to the [rules of the lookup argument](lookup-argument.md), obtaining $A'(X)$ and $S'(X)$.
+
+Finally, the prover creates blinding commitments for all of the lookups
+
+$$\mathbf{L}_p = \left[ (\text{Commit}(A'(X))), \text{Commit}(S'(X))), \dots \right]$$
```
daira, line 48: Why not also Z(X)?

book/src/design/proving-system/circuit-commitments.md
```diff
@@ -0,0 +1,80 @@
...
+In addition to committing to the individual permuted lookups, the prover needs to commit
+to constraint polynomials that enforce the correct relation between the permuted lookups
+and their un-permuted forms:
+
+$$L(X) = \frac{(A_\text{compressed}(X) + \beta)(S_\text{compressed}(X) + \gamma)}{(A'(X) + \beta)(S'(X) + \gamma)}$$
+
+$\beta$ and $\gamma$ are used to combine the permutation arguments for $A'(X)$ and $S'(X)$
+while keeping them independent. We can reuse $\beta$ and $\gamma$ from the equality
+constraint permutation here because they serve the same purpose in both places, and we
+aren't trying to combine the lookup and equality constraint permutation arguments.
```
daira, line 74:
```suggestion
aren't trying to combine the lookup and equality constraint permutation arguments. The important
thing here is that the verifier samples $\beta$ and
$\gamma$ after the prover has committed to all
the cell values, including: those used in lookup
columns, $A'$, $S'$, [to do: and $Z$].
```

book/src/design/proving-system/inner-product.md
```diff
@@ -0,0 +1,11 @@
+# Inner product argument
+
+Halo 2 uses a polynomial commitment scheme for which we can create polynomial commitment
+opening proofs, based around the Inner Product Argument.
+
+> TODO: Explain Halo 2's variant of the IPA.
+>
+> It is very similar to $\text{PC}_\text{DL}.\text{Open}$ from Appendix A.2 of [BCMS20].
+> See [this comparision](comparison.md#bcms20-appendix-a2) for details.
```
daira, line 9:
```suggestion
> See [this comparison](comparison.md#bcms20-appendix-a2) for details.
```

book/src/design/proving-system/vanishing.md
```diff
@@ -0,0 +1,79 @@
+# Vanishing argument
+
+Having committed to the circuit assignments, the prover now needs to demonstrate that the
+various circuit relations are satisfied:
+
+- The custom gates, represented by polynomials $\text{gate}_i(X)$.
+- The rules of the lookup arguments.
+- The rules of the equality constraint permutations.
+
+Each of these relations is inherently represented as a polynomial of degree $d$ (the
```
daira, line 10:
```suggestion
Each of these relations is represented as a polynomial of degree $d$ (the
```

daira (approved):
ACK with comments/questions.

book/src/design/proving-system.md
```diff
@@ -1 +1,71 @@
...
+
+To aid our explanations, we will at times refer to the following example constraint
+system:
+
+- Four advice columns $a, b, c, d$.
+- One fixed column $f$.
+- Three custom gates:
+  - $a \cdot b \cdot c_{-1} - d = 0$
+  - $f_{-1}(c) = 0$
+  - $f(d \cdot a) = 0$
```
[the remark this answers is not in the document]
str4d: At, I copied this from the example Sean have me, and I think he was using this fixed column like a selector. You're right that cdot here is what is intended.

book/src/design/proving-system.md
```diff
@@ -1 +1,69 @@
 # Proving system
+
+The Halo 2 proving system can be broken down into five stages:
+
+1. Commit to polynomials encoding the main components of the circuit:
+   - Cell assignments.
+   - Permuted values for each lookup argument.
+   - Equality constraint permutations.
+   - Products for each lookup argument.
```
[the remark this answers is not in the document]
str4d: I was separating these out to reflect the ordering of the commitments, but with your other suggestion to that section I think this is fine.

book/src/design/proving-system/vanishing.md
str4d: Hmm, I was trying to distinguish between the relation itself (which has degree `d` with respect to the columns), and the relation polynomial (which has degree `d(n-1)` with respect to `X`). Suggestions on alternative wording?

str4d: Actually, I could just say that ☺️

book/src/design/proving-system/circuit-commitments.md
str4d: I think the confusion was that `L_c(X)` below _is_ `Z(X)`. The original HackMD never defined `Z(X)`, and I missed the connection while extracting `L_c(X)` from the code. Pushed a fix.

book/src/design/proving-system.md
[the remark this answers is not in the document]
str4d: Reading this again, I disagree with this suggestion: "Cell assignments" refers specifically to the cells in the main circuit columns (that users assign), not the pseudo-columns generated by the permutations.

ebfull (changes requested):

ebfull:
ACK

PR #180, opened 2021-02-13
Add Expression::Constant variant

therealyingtong:
This is based on #176 which is currently unmerged.

### DIFF
--- a/src/plonk/circuit.rs
+++ b/src/plonk/circuit.rs
@@ -224,6 +224,8 @@ pub trait Circuit<F: Field> {
 /// Low-degree expression representing an identity that must hold over the committed columns.
 #[derive(Clone, Debug)]
 pub enum Expression<F> {
+    /// This is a constant polynomial
+    Const(F),
     /// This is a fixed column queried at a certain relative location
     Fixed(usize),
     /// This is an advice (witness) column queried at a certain relative location
@@ -243,6 +245,7 @@ impl<F: Field> Expression<F> {
     /// operations.
     pub fn evaluate<T>(
         &self,
+        const_column: &impl Fn(F) -> T,
         fixed_column: &impl Fn(usize) -> T,
         advice_column: &impl Fn(usize) -> T,
         instance_column: &impl Fn(usize) -> T,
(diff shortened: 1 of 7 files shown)


src/poly/domain.rs
```diff
@@ -177,6 +177,14 @@ impl<G: Group> EvaluationDomain<G> {
         }
     }
 
+    /// Returns a constant polynomial in the Lagrange coefficient basis
+    pub fn const_lagrange(&self, scalar: G) -> Polynomial<G, LagrangeCoeff> {
```
ebfull, line 181: Rename this to `constant` because `const` is a keyword in Rust that indicates the function can be evaluated at compile time; that is not the case here.

src/poly/domain.rs
```diff
@@ -186,6 +194,15 @@ impl<G: Group> EvaluationDomain<G> {
         }
     }
 
+    /// Returns a constant polynomial in the extended Lagrange coefficient
+    /// basis
+    pub fn const_extended(&self, scalar: G) -> Polynomial<G, ExtendedLagrangeCoeff> {
```
ebfull, line 199: Rename `const` to `constant` here as well.

src/plonk/circuit.rs
```diff
@@ -224,6 +224,8 @@ pub trait Circuit<F: Field> {
 /// Low-degree expression representing an identity that must hold over the committed columns.
 #[derive(Clone, Debug)]
 pub enum Expression<F> {
+    /// This is a constant polynomial
+    Const(F),
```
ebfull, line 228: Rename to `Constant(F)`

src/plonk/circuit.rs
```diff
@@ -243,6 +245,7 @@ impl<F: Field> Expression<F> {
     /// operations.
     pub fn evaluate<T>(
         &self,
+        const_column: &impl Fn(F) -> T,
```
ebfull, line 248: Rename to `constant` since it does not refer to a column.

ebfull (changes requested):
Minor nits, but otherwise good!

str4d (approved):
utACK

ebfull (approved):

PR #185, opened 2021-02-14
Rename "auxiliary column" to "instance column" in the book and in code

### LINKED ISSUE #181: Rename "auxiliary column" to "instance column"
This was a source of confusion for Kris when reading the [UltraPLONK Arithmetization](https://zcash.github.io/halo2/concepts/arithmetization.html) section in conjunction with the Zcash protocol spec.

In the protocol spec, we use "auxiliary input" for *private* inputs. This follows some papers on proof systems, such as [[BCTV14]](https://eprint.iacr.org/2013/879). In the Halo 2 book and code, we currently use "auxiliary column" for what are usually *public* inputs. Technically, they don't have to be public, but they do have to be part of the instance that is agreed between the prover and verifier.

I suggested that we change this to "instance column", and @sean agreed. Another candidate was "supplemental", but that's a bit too long and hard to shorten.

daira:
fixes #181

### DIFF
--- a/book/src/design/proving-system/circuit-commitments.md
+++ b/book/src/design/proving-system/circuit-commitments.md
@@ -9,9 +9,9 @@ the $i$th fixed column. Without loss of generality, we'll similarly define $A_{i
 represent the advice and instance assignments.
 
 > We separate fixed columns here because they are provided by the verifier, whereas the
-> advice columns are provided by the prover. The instance columns are agreed between both
-> parties. In practice, the commitments to instance and fixed columns are computed by
-> both the prover and verifier, and only the advice commitments are stored in the proof.
+> advice and instance columns are provided by the prover. In practice, the commitments to
+> instance and fixed columns are computed by both the prover and verifier, and only the
+> advice commitments are stored in the proof.
 
 To commit to these assignments, we construct Lagrange polynomials of degree $n - 1$ for
 each column, over an evaluation domain of size $n$ (where $\omega$ is the $n$th primitive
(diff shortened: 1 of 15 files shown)


book/src/design/proving-system/circuit-commitments.md
```diff
@@ -4,14 +4,14 @@
...
 the $i$th fixed column. Without loss of generality, we'll similarly define $A_{i,j}$ to
-represent the advice and auxiliary assignments.
+represent the advice and instance assignments.
 
 > We separate fixed columns here because they are provided by the verifier, whereas the
-> advice and auxiliary columns are provided by the prover. In practice, the commitments to
-> auxiliary and fixed columns are computed by both the prover and verifier, and only the
-> advice commitments are stored in the proof.
+> advice columns are provided by the prover. The instance columns are agreed between both
+> parties. In practice, the commitments to instance and fixed columns are computed by
```
str4d, line 13: This change makes the point I was trying to convey less clear. I want the reader to understand that the reason we defined a separate `F_{i,j}` but not a separate `I_{i,j}` is because both advice and instance columns are only defined at proving time (and within the proving system are treated almost identically), whereas fixed columns are defined in advance (hence "provided by the verifier"). It does not matter that both parties agree on the instance columns; in practice, the prover provides them to the verifier out-of-band (wrt the proof), and any rejection of the instance column contents occurs before the verifier participates in the verification protocol.

str4d (approved):
utACK c769dd0fa6701613efa51cae25014aa4ae92a5e9

PR #187, opened 2021-02-15
Book fixes

str4d:


book/src/background/polynomials.md
```diff
@@ -7,7 +7,16 @@ A(X) = a_0 + a_1 X + a_2 X^2 + a_3 X^3
...
 defines a degree-$3$ polynomial. $a_0$ is referred to as the constant term. Polynomials of
-degree $n-1$ have $n$ coefficients.
+degree $n-1$ have $n$ coefficients. We will often want to compute the result of replacing
+the formal indeterminate $X$ with some concrete value $x$, which we denote by $A(x)$.
+
+> In mathematics this is commonly referred to as "evaluating $A(X)$ at a point $x$".
+> The word "point" here stems from the geometrical usage of polynomials in the form
+> $y = A(x)$, where $(x, y)$ is the coordinate of a point in two-dimensional space.
+> However, the polynomials we deal with are almost always constrained to equal zero, and
+> $x$ will be an [element of some group](fields.md#groups). This should not be confused
```
ebfull, line 17: x will be an element of a field not a group

ebfull (changes requested):

book/src/background/polynomials.md
str4d: I wasn't sure initially whether restricting `x` to a field specifically, made sense given that we actually draw our evaluations from a multiplicative subgroup of a field. But we do actually require that the evaluation domain support both addition and multiplication in order to implement polynomials, so evaluation points can't be elements of an arbitrary group with a single operation.

ebfull: In this context we aren't referring to the domain of the polynomial, just the general sense of "evaluating" the polynomial at some x. We'll never be evaluating polynomials at points in the domain not least because this would leak witness information into the proof! :)

str4d, line 17:
```suggestion
> $x$ will be an [element of some field](fields.md). This should not be confused
```

ebfull (approved):

PR #193, opened 2021-02-17
Allow permutations to be over all column types

### LINKED ISSUE #190: Allow permutations to be over all column types
Just like we did with the lookup argument.

therealyingtong:
Closes #190


ebfull (approved):

str4d (commented):
Gah, I forgot to submit this before @ebfull merged the PR.

Issue #165, opened 2021-02-02
Split plonk/circuit module up


ebfull:
This is a _sketch_ of how we might break some of these things out into.

```
plonk
- permutation
- verifier
- prover
- verifier
- keygen
- Assignment
- Expression
- ConstraintSystem
- Cell
circuit
- Column
- Region
- Layouter
- Chip
- RegionCell
- LooseCell
- Permutation
```

PR #199, opened 2021-02-18
Upstream Permutation struct into plonk::circuit

### LINKED ISSUE #189: Upstream the Permutation structure from crate::circuit into plonk module

therealyingtong:
Closes #189

### DIFF
--- a/src/circuit.rs
+++ b/src/circuit.rs
@@ -4,7 +4,7 @@ use std::{fmt, marker::PhantomData};
 
 use crate::{
     arithmetic::FieldExt,
-    plonk::{Advice, Any, Column, ConstraintSystem, Error, Fixed},
+    plonk::{Advice, Any, Column, Error, Fixed, Permutation},
 };
 
 pub mod layouter;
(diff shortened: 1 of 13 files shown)


ebfull (changes requested):
I should have been more specific in #189 about what I was desiring.

I want to change `ConstraintSystem` (in `crate::plonk::circuit`) so that it returns

```rust
pub fn permutation(&mut self, columns: &[Column<Any>]) -> Permutation {
```

and move the `Permutation` structure that currently exists in `src/circuit.rs` into `src/plonk/circuit.rs` where `ConstraintSystem` is defined.

Then, we can change the `Assignment` trait to do

```rust
    /// Assign two advice columns to have the same value
    fn copy(
        &mut self,
        permutation: Permutation,
        left_column: Column<Any>,
        left_row: usize,
        right_column: Column<Any>,
        right_row: usize,
    ) -> Result<(), Error>;
```

and so it's on the implementation of `Assignment` to figure out the index into the permutation's involved columns if it needs it. It can do that by scanning the vector inside of the `Permutation` structure. (This means that `Permutation` will need getters probably.)

Overall this simplifies the API so that downstream APIs don't need to care about "the permutation index" and the "column index inside the permutation" which are annoying things that can be screwed up easily.

src/circuit.rs
```diff
@@ -37,24 +37,6 @@ pub struct Cell {
     column: Column<Any>,
 }
 
-/// A permutation configured by a chip.
-#[derive(Clone, Debug)]
-pub struct Permutation {
```
str4d, line 42: I'm not sure that we want to move `Permutation` into `halo2::plonk::circuit`. Per #165 (where we want to get rid of `halo2::plonk::circuit`), the goal is for `halo2::circuit` to contain the types that most crate users (circuit implementors) deal with, while `halo2::plonk` contains the inner details that few people will use directly beyond `create_proof` and `verify_proof`.

str4d: Hmm, I now see that @ebfull suggested the move. I'm only weakly against it, so we can keep it as-is and then maybe move it back during #165.

str4d (approved):

ebfull (approved):

PR #202, opened 2021-02-20
Migrate to group traits

str4d:
Part of #32.

### DIFF
--- a/src/pasta/curves.rs
+++ b/src/pasta/curves.rs
@@ -57,7 +57,7 @@ macro_rules! new_curve_impl {
 
             impl_projective_curve_specific!($name, $base, $curve_type);
 
-            fn zero() -> Self {
+            fn identity() -> Self {
                 Self {
                     x: $base::zero(),
                     y: $base::zero(),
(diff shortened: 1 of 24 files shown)


str4d:
Best reviewed per-commit.

ebfull (approved):

src/pasta/curves.rs
```diff
@@ -796,14 +827,61 @@ macro_rules! impl_projective_curve_specific {
...
+}
+
+macro_rules! impl_projective_curve_ext {
+    ($name:ident, $name_affine:ident, $iso:ident, $base:ident, special_a0_b5) => {
+        fn hash_to_curve<'a>(domain_prefix: &'a str) -> Box<dyn Fn(&[u8]) -> Self + 'a> {
+            use super::hashtocurve;
+
+            Box::new(move |message| {
+                let mut us = [Field::zero(); 2];
+                hashtocurve::hash_to_field($name_affine::CURVE_ID, domain_prefix, message, &mut us);
```
daira, line 842: Does it make more sense for `CURVE_ID` to be defined on `CurveExt`? Then this macro would not require `$name_affine`.

(I think we'll also remove `BLAKE2B_PERSONALIZATION` when we switch to using SSWU for the Halo bases; I just wanted to do that separately.)

daira: Oh, I guess this has already merged. We can make that simplification when we switch the Halo bases.

daira:
Post-hoc ACK.

PR #204, opened 2021-02-22
Generate parameters using hash to curve

ebfull:
This PR also removes the `BLAKE2B_PERSONALIZATION` string.


src/arithmetic/curves.rs
```diff
@@ -87,10 +93,8 @@ pub trait CurveAffine:
...
     type Base: FieldExt;
-
-    /// Personalization of BLAKE2b hasher used to generate the uniform
-    /// random string.
-    const BLAKE2B_PERSONALIZATION: &'static [u8; 16];
+    /// The projective form of the curve
+    type CurveExt: CurveExt<AffineExt = Self, ScalarExt = <Self as CurveAffine>::ScalarExt>;
 
     /// CURVE_ID used for hash-to-curve.
     const CURVE_ID: &'static str;
```
daira, line 100: I suggested at https://github.com/zcash/halo2/pull/202#discussion_r580572834 that it might be simpler to move `CURVE_ID` to `CurveExt`.

src/poly/commitment.rs
```diff
@@ -48,51 +45,41 @@ impl<C: CurveAffine> Params<C> {
...
-            g.resize(n as usize, C::identity());
+            g.resize(n as usize, C::Curve::identity());
 
             parallelize(&mut g, move |g, start| {
-                let mut hasher = Blake2bParams::new()
-                    .hash_length(32)
-                    .personal(C::BLAKE2B_PERSONALIZATION)
-                    .to_state();
-                hasher.update(b"G vector");
+                let hasher = C::CurveExt::hash_to_curve("Halo2-Parameters");
```
daira, line 53: It should be possible to create this outside `parallelize`, I think (because it's not mutable).

ebfull: Unfortunately, `parallelize` needs the provided closure to be cloneable (because it spins up multiple threads) and so the hasher needs to impl `Clone` as well. But only `auto` traits like `Sync` and `Send` are allowed to be "additional traits" in trait objects, not traits like `Clone`. We'd have to make a new function trait just for this particular situation, which is annoying.

ebfull: In practice the cost is minor and we can revisit this in the future when GATs are added to Rust or something.

daira (approved):
utACK with comment.

str4d (approved):
utACK 12230a49e058e3aa17ff72f82e2f4ac287dd33e8

daira (commented):
Post-hoc ACK.

PR #205, opened 2021-02-23
Add documentation of Minimum Supported Rust Version.

daira:

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,6 +4,10 @@
 
 ## [Documentation](https://docs.rs/halo2)
 
+This crate currently has a Minimum Supported Rust Version of 1.49.0. When const generics
+are stabilized [in Rust 1.51.0](https://github.com/rust-lang/rust/pull/79135), we plan
+to require that version.
+
 ## License
 
 Copyright 2020 The Electric Coin Company.


README.md
```diff
@@ -4,6 +4,10 @@
 
 ## [Documentation](https://docs.rs/halo2)
 
+This crate currently has a Minimum Supported Rust Version of 1.49.0. When const generics
```
str4d, line 7: If we _are_ deciding to pin a MSRV now, then we should both make this change in CI, and add a `rust-toolchain` file.

ebfull:
Can we punt this to `1.51.0`'s release?

str4d:
I'm fine with waiting until 1.51.0, which should be released the week of March 22nd.

daira:
Currently we fail to build on Rust 1.48.0 or earlier. This wasted several hours of my time, trying to fix compiler errors that were actually compiler bugs. So I think this should be merged now for documentation; we can do the other stuff with CI later.

ebfull:
Okay, @daira. Sounds reasonable.

daira:
Thanks :-)

Issue #116, opened 2020-12-31
Implement selector combining optimization


daira:
There's an optimization we can do after floor planning: if two or more selectors are never used on the same row, then they can be combined into one selector, subject to max degree limitations.

For example, if we have polynomial constraints q<sub>A</sub> · p<sub>A</sub>(...) = 0 and q<sub>B</sub> · p<sub>B</sub>(...) = 0, where q<sub>A</sub> and q<sub>B</sub> are boolean selectors that are never used together, then we can transform that into q<sub>C</sub> · ((q<sub>C</sub> - 2) · p<sub>A</sub>(...) + (q<sub>C</sub> - 1) · p<sub>B</sub>(...)) = 0, where q<sub>C</sub> = 1 if q<sub>A</sub> is 1;  q<sub>C</sub> = 2 if q<sub>B</sub> is 1;  and q<sub>C</sub> = 0 otherwise.

in order for that to work, we need the optimizer to know which fixed columns are boolean selectors and which regions they are used by.

Alternatively we can do this optimization at the end after synthesis. But I would like to have the information to be able to do it before synthesis if we need to.

str4d:
In PR gardening, @daira said ze would like "non-boolean selectors", that have the two values `[0, non-zero]` rather than `[0, 1]`. These can't be combined arbitrarily; you can combine arbitrary boolean selectors with at most one non-boolean selector.

therealyingtong:
Closed by #347

PR #208, opened 2021-02-23
Add a Selector type

str4d:
This currently just wraps a `Column<Fixed>`, but enables us to start writing circuits that can later have their selector usage optimised.

Part of #116.

### DIFF
--- a/src/plonk/circuit.rs
+++ b/src/plonk/circuit.rs
@@ -7,8 +7,11 @@ use std::{
 };
 
 use super::{lookup, permutation, Error};
-use crate::arithmetic::FieldExt;
-use crate::poly::Rotation;
+use crate::{
+    arithmetic::FieldExt,
+    circuit::{Chip, Region},
+    poly::Rotation,
+};
 
 /// A column type
 pub trait ColumnType: 'static + Sized + std::fmt::Debug {}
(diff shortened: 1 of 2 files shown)


src/plonk/circuit.rs
```diff
@@ -126,6 +129,71 @@ impl TryFrom<Column<Any>> for Column<Instance> {
...
+///
+/// struct Config {
+///     a: Column<Advice>,
+///     b: Column<Advice>,
+///     s: Selector,
+/// }
+///
+/// fn circuit_logic<C: Chip>(mut layouter: impl Layouter<C>) -> Result<(), Error> {
+///     let config = layouter.config().clone();
+///     # let config: Config = todo!();
```
str4d, line 171: This is a trick to make the documentation look correct, while also type-checking correctly in CI.

src/plonk/circuit.rs
```diff
@@ -126,6 +129,71 @@ impl TryFrom<Column<Any>> for Column<Instance> {
...
+impl Selector {
+    /// Enable this selector at the given offset within the given region.
+    pub fn enable<C: Chip>(&self, region: &mut Region<C>, offset: usize) -> Result<(), Error> {
+        // TODO: Ensure that the default for a selector's cells is always zero, if we
+        // alter the proving system to change the global default.
+        // TODO: Add Region::enable_selector method to allow the layouter to control the
+        // selector's assignment.
+        // https://github.com/zcash/halo2/issues/116
+        region
+            .assign_fixed(|| "", self.0, offset, || Ok(C::Field::one()))
```
str4d, line 192: TODO: How should we annotate selectors being enabled? Should `Selector::enable` take an annotation, or should we compute this automatically (perhaps from a name we provide to `ConstraintSystem::selector`)?

src/plonk/circuit.rs
```diff
@@ -614,6 +687,13 @@ impl<F: Field> ConstraintSystem<F> {
         self.gates.push((name, poly));
     }
 
+    /// Allocate a new selector.
+    pub fn selector(&mut self) -> Selector {
```
str4d, line 691: We might also want to add `Selector::new(&mut ConstraintSystem)` to match `Permutation::new` as in #199.

src/plonk/circuit.rs
```diff
@@ -479,6 +547,11 @@ impl<F: Field> ConstraintSystem<F> {
         index
     }
 
+    /// Query a selector at a relative position.
+    pub fn query_selector(&mut self, selector: Selector, at: Rotation) -> Expression<F> {
```
str4d, line 551: If this querying logic was moved onto a gate-specific wrapper around `ConstraintSystem` (similar to how a `Region` will at some point wrap a `Layouter`), we could collect metadata about which columns / selectors are used in which gates (probably more easily than extracting that information from the gate expressions).

ebfull (approved):

str4d:
Rebased to fix conflicts with #199.

PR #215, opened 2021-02-25
Refactoring module structure

ebfull:


examples/sha256/table16.rs
```diff
@@ -3,8 +3,7 @@ use std::marker::PhantomData;
 use super::Sha256Instructions;
 use halo2::{
     arithmetic::FieldExt,
-    circuit::{Cell, Chip, Layouter, Region},
-    plonk::{Advice, Column, ConstraintSystem, Error, Permutation},
+    plonk::{Advice, Cell, Chip, Column, ConstraintSystem, Error, Layouter, Permutation, Region},
```
daira, line 6: Explain the motivation for this again?

Chips, regions, and layouters seem to me to be at a separate layer of abstraction than the proof system. So I worry that with this move we're getting further away from the structure that we'll eventually want.

ebfull: As @str4d mentioned, there's no need for circuit developers to need to play with the `Assignment` trait, but besides that circuit devs need to touch every single one of these types. (We can hide `Assignment`.) And they're all specific to the PLONK arithmetization. Users are going to put in a `Circuit` into the prover and get a `Layouter` on the other end during synthesis, so the layouter abstraction really has to be under the purview of the `plonk` module.

PR #216, opened 2021-02-26
add rust toolchain file

honeywest:


steven-ecc:
Hi! Thanks for this contribution and for [#217.](https://github.com/zcash/halo2/pull/217)

Could you please sign the [contributor agreement](https://github.com/zcash/halo2/blob/main/Contributor_Agreement) and send it to me at steven@electriccoin.co works? Signing the agreement allows us to redistribute the code later under another license if we need to, without having to get your permission. This is similar to how contributor agreements work in Apache software projects as well.

honeywest:
I finished sign  the contributor agreement and send to you.

str4d (approved):
I'm fine with this, because we can't build `halo2` prior to 1.49.0. We can update the toolchain file if/when we bump to MSRV 1.51.

Issue #152, opened 2021-01-27
Prove that endoscaling never produces a scalar in the 2^32 multiplicative subgroup


daira:
This eliminates a possible reason for incompleteness of proving, when using the endoscaling optimization: if we end up using a scalar in the subgroup, then the proof will not verify because (@sean help me out here :-p ).

It is possible to prove this by exhaustive search with an efficient shortest-vector algorithm in dimension 2, e.g. algorithm 3 of https://eprint.iacr.org/2020/454.pdf . We need the version with the stricter test at step 8, so that we get a short**est** vector. For each possible subgroup element, if the norm of this vector is greater than the maximum possible norm corresponding to an endoscalar, then that subgroup element cannot be an endoscalar. Some back-of-the-envelope calculations of the cost of the exhaustive search seem to suggest that it could be done in less than 24 hours, based on the performance of @str4d's PR to add Pornin's algorithm to curve25519-dalek: https://github.com/dalek-cryptography/curve25519-dalek/pull/323

This isn't strictly necessary; we could just verify the proof and redo it with new randomization (or simply abort). But it would increase confidence in correctness of the endoscaling optimization.

Issue #151, opened 2021-01-27
Add support for full-field challenge space


ebfull:
Right now we do not give the option to use the full field and only use the endoscaling, but for orchard we'll want to use the full field to simplify analysis until we end up needing PCD later on down the road. Let's make the challenge space swappable somehow.

PR #219, opened 2021-02-27
semi-final zero-knowledge changes

### LINKED ISSUE #63: Implement zero-knowledge-ness

ebfull:
Closes #63

**Change 1:**

Currently, the prover commits to the "chunks" of `h(X)` by committing to `h_0(X), h_1(X), ..., h_d(X)` such that `h(X) = h_0(X) + x^n h_1(X) + ... + x^{dn} h_d(X)`. Then after `x` is sampled the prover commits to `h_0(x)` etc. The verifier checks that `h(x)` is as expected, and opens each of the commitments at `x` to check if the prover's committed evaluations are correct.

The problem is that `h_0(x), h_1(x), ..., h_d(x)` are not always uniformly distributed. It's not clear that blinding factors in the polynomial alone can resolve this.

The alternative is to compute a linear combination of the prover's commitments (with powers of `x^n`) and open that commitment. This is actually more efficient (smaller proofs) but the reason we were not doing this originally was because it's _not_ as efficient inside the recursive circuit; x^n is a large arbitrary scalar in the current protocol. We can find a simpler option another time though.

**Change 2:**

The prover will now always commit to a random polynomial that is opened at `x` along with `h(X)`. This ensures that during multiopen the verifier will not learn `h(x_3)`.

**Change 3:**

The `Polynomial` structure now is used to delineate between "active" evaluations (in Lagrange form, these are available cells) and "inactive" evaluations which consist of: 1 extra row that is sometimes used (in permutation/lookup argument) to store stuff, and the remaining rows are just blinding factors.

All of the prover's commitments have random blinding factors added.

**Change 4:**

In order to compensate for the fewer rows available, the permutation and lookup arguments are modified slightly. All of the custom (user) gates are disabled on the "inactive" rows.

### DIFF
--- a/src/plonk/circuit.rs
+++ b/src/plonk/circuit.rs
@@ -807,6 +807,10 @@ impl<F: Field> ConstraintSystem<F> {
         // h(x) is derived by the other evaluations so it does not reveal
         // anything; in fact it does not even appear in the proof.
 
+        // h(x_3) is also not revealed; the verifier only learns a single
+        // evaluation of a polynomial in x_1 which has h(x_3) and another random
+        // polynomial evaluated at x_3 as coefficients
+
         // Add an additional blinding factor so that each polynomial becomes
         // computationally indistinguishable from random
         factors + 1
--- a/src/plonk/circuit.rs
+++ b/src/plonk/circuit.rs
@@ -473,6 +473,7 @@ pub struct ConstraintSystem<F: Field> {
     pub(crate) num_instance_columns: usize,
     pub(crate) gates: Vec<(&'static str, Expression<F>)>,
     pub(crate) advice_queries: Vec<(Column<Advice>, Rotation)>,
+    num_advice_queries: Vec<usize>,
     pub(crate) instance_queries: Vec<(Column<Instance>, Rotation)>,
     pub(crate) fixed_queries: Vec<(Column<Fixed>, Rotation)>,
 
@@ -518,6 +519,7 @@ impl<F: Field> Default for ConstraintSystem<F> {
             gates: vec![],
             fixed_queries: Vec::new(),
             advice_queries: Vec::new(),
+            num_advice_queries: Vec::new(),
             instance_queries: Vec::new(),
             permutations: Vec::new(),
             lookups: Vec::new(),
@@ -613,6 +615,7 @@ impl<F: Field> ConstraintSystem<F> {
         // Make a new query
         let index = self.advice_queries.len();
         self.advice_queries.push((column, at));
+        self.num_advice_queries[column.index] += 1;
 
         index
     }
@@ -741,6 +744,7 @@ impl<F: Field> ConstraintSystem<F> {
             column_type: Advice,
         };
         self.num_advice_columns += 1;
+        self.num_advice_queries.push(0);
         tmp
     }
 
@@ -784,5 +788,27 @@ impl<F: Field> ConstraintSystem<F> {
         }
 
         degree
+    }
+
+    /// Compute the number of blinding factors necessary to perfectly blind
+    /// each of the prover's witness polynomials.
+    pub fn blinding_factors(&self) -> usize {
+        // All of the prover's advice columns are evaluated at most
+        let factors = *self.num_advice_queries.iter().max().unwrap_or(&1);
+        // distinct points during gate checks. In the permutation and lookup
+        // argument the witness polynomials are evaluated at most 2 times:
+
+        let factors = std::cmp::max(2, factors);
+
+        // Each polynomial is evaluated at most an additional time during
+        // multiopen
+        let factors = factors + 1;
+
+        // h(x) is derived by the other evaluations so it does not reveal
+        // anything; in fact it does not even appear in the proof.
+
+        // Add an additional blinding factor so that each polynomial becomes
+        // computationally indistinguishable from random
+        factors + 1
     }
 }
--- a/src/plonk/circuit.rs
+++ b/src/plonk/circuit.rs
@@ -784,7 +784,8 @@ impl<F: Field> ConstraintSystem<F> {
         // Account for each gate to ensure our quotient polynomial is the
         // correct degree and that our extended domain is the right size.
         for (_, poly) in self.gates.iter() {
-            degree = std::cmp::max(degree, poly.degree());
+            // All gates are multiplied by (1 - l_cover(X))
+            degree = std::cmp::max(degree, poly.degree() + 1);
         }
 
         degree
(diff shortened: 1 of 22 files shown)


ebfull:
Closes #63

therealyingtong (commented):
Initial review.

str4d (changes requested):
Generally LGTM.

src/plonk/vanishing/verifier.rs
```diff
@@ -32,55 +33,41 @@ impl<C: CurveAffine> Argument<C> {
...
     pub(in crate::plonk) fn verify(
-        &self,
+        self,
         expressions: impl Iterator<Item = C::Scalar>,
         y: ChallengeY<C>,
         xn: C::Scalar,
-    ) -> Result<(), Error> {
+    ) -> Result<Evaluated<C>, Error> {
         let expected_h_eval = expressions.fold(C::Scalar::zero(), |h_eval, v| h_eval * &*y + &v);
+        let expected_h_eval = expected_h_eval * ((xn - C::Scalar::one()).invert().unwrap());
```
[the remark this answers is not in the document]
ebfull: It will happen only if x is a 2^n root of unity; this is part of the "disallowed" challenge space (see also #151,  #152, where we need to clear up what's going on with the challenges) but in practice it will be unlikely to occur. There is also another reason `x` is prohibited from being a 2^n root of unity: if the prover supplied `a(x)` in such a scenario for their secret polynomial `a(X)` then it would reveal one of the cell values!

src/poly/domain.rs
```diff
@@ -372,18 +381,71 @@ impl<G: Group> EvaluationDomain<G> {
...
+            let rotation = Rotation(rotation);
+            *result *= common;
+
+            if rotation.0 >= 0 {
+                *result *= self.get_omega().pow_vartime(&[rotation.0 as u64]);
+            } else {
+                *result *= self
+                    .get_omega_inv()
+                    .pow_vartime(&[(rotation.0 as i64).abs() as u64]);
+            }
```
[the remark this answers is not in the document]
therealyingtong: i think they already are!

src/poly/domain.rs
```diff
@@ -372,18 +372,71 @@ impl<G: Group> EvaluationDomain<G> {
...
+        // We refer to
+        // (1 / \prod_{j=0,j \neq i}^{n - 1} (omega^i - omega^j))
+        // as the barycentric weight of omega^i.
+        // We know that for i = 0
+        // (1 / \prod_{j=0,j \neq i}^{n - 1} (omega^i - omega^j))
+        // = (1 / n)
+        // If we multiply (1 / n) by (omega^{i})^-1 then we obtain
+        // (1 / \prod_{j=0,j \neq 0}^{n - 1} (omega^{i + 1} - omega^{j + 1}))
+        // = (1 / \prod_{j=0,j \neq -i}^{n - 1} (omega^i - omega^j))
+        // which is the barycentric weight of omega^{-i}
```
therealyingtong, line 420: TODO: fix this comment

src/plonk/circuit.rs
```diff
@@ -780,9 +784,36 @@ impl<F: Field> ConstraintSystem<F> {
...
+        // anything; in fact it does not even appear in the proof.
+
+        // h(x_3) is also not revealed; the verifier only learns a single
+        // evaluation of a polynomial in x_1 which has h(x_3) and another random
+        // polynomial evaluated at x_3 as coefficients
+
+        // Add an additional blinding factor so that each polynomial becomes
+        // computationally indistinguishable from random
+        factors + 1
+    }
```
therealyingtong, line 818:
```suggestion
    }

    /// Compute the number of inactive rows needed
    pub fn inactive(&self) -> usize {
        ConstraintSystem::blinding_factors(&self) + 1
    }
```

therealyingtong: this saves us from having to do `blinding_factors + 1` or `- blinding_factors - 1` in various places.

src/plonk/lookup.rs
```diff
@@ -25,15 +25,20 @@ impl<F: Field> Argument<F> {
...
+        // z'(omega X) (a'(X) + \beta) (s'(X) + \gamma)
+        // - z'(X) (\theta^{m-1} a_0(X) + ... + a_{m-1}(X) + \beta) (\theta^{m-1} s_0(X) + ... + s_{m-1}(X) + \gamma)
+        // )
         //
         // degree 2:
-        // (a′(X)−s′(X))⋅(a′(X)−a′(\omega{-1} X)) = 0
+        // l_0(X) * (a'(X) - s'(X))
+        //
+        // degree 3:
+        // (1 - (l_last(X) + l_cover(X))) * ((a′(X)−s′(X))⋅(a′(X)−a′(\omega{-1} X)))
```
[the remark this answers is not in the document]
therealyingtong: (@ebfull mentioned this in our pairing): the `\omega^{-1} X` here serves to check that each row in the input expression is either: 
- equal to the corresponding row in the table expression `(a′(X)−s′(X)`, or
- equal to the previous row in the input expression `(a′(X)−a′(\omega{-1} X))`

src/plonk/prover.rs
```diff
@@ -40,16 +40,29 @@ pub fn create_proof<C: CurveAffine, T: TranscriptWrite<C>, ConcreteCircuit: Circ
...
         .map(|instance| -> Result<InstanceSingle<C>, Error> {
-            let instance_commitments_projective: Vec<_> = instance
+            let instance_values = instance
+                .iter()
+                .map(|values| {
+                    let mut poly = domain.empty_lagrange(blinding_factors + 1);
+                    poly.clear_inactive();
+                    for (poly, value) in poly.iter_mut().zip(values.iter()) {
+                        *poly = *value;
+                    }
```
therealyingtong, line 61:
```suggestion
                    for (coeff, value) in poly.iter_mut().zip(values.iter()) {
                        *coeff = *value;
                    }
```

src/poly/commitment.rs
```diff
@@ -121,7 +121,7 @@ impl<C: CurveAffine> Params<C> {
         let mut tmp_scalars = Vec::with_capacity(poly.len() + 1);
         let mut tmp_bases = Vec::with_capacity(poly.len() + 1);
```
therealyingtong, line 122:
```suggestion
        let mut tmp_scalars = Vec::with_capacity(poly.values.len() + 1);
        let mut tmp_bases = Vec::with_capacity(poly.values.len() + 1);
```

str4d:
Replaced by #316.

PR #221, opened 2021-02-27
Fix fields sbb

honeywest:
borrow
original: 0 or 0xff...ff
now: 0 or 0x1


str4d:
The current behaviour is intentional. We want field element subtraction to be constant time, and conditionals like `if borrow == 0` are not. Having `sbb` return `0xfff...fff` on overflow gives us two properties:

- The borrow can be treated as a boolean flag (either `0x0` or `0xfff...fff`, which has the same boolean properties as `0x0` or `0x1`).
- The borrow can be used as a mask for conditional addition:
  - If `borrow = 0x0` then `a + (b & borrow) == a`.
  - If `borrow = 0xfff...fff` then `a + (b & borrow) == a + b`.

PR #222, opened 2021-02-28
Typos

kevaundray:
In some places, s was being used and in others z. This PR changes the spelling to be consistent and corrects a few typos. 

If the spelling change is not needed, I can cherry-pick these commits out.

### DIFF
--- a/book/src/SUMMARY.md
+++ b/book/src/SUMMARY.md
@@ -31,6 +31,6 @@
   - [Polynomials](background/polynomials.md)
   - [Cryptographic groups](background/groups.md)
   - [Elliptic curves](background/curves.md)
-  - [UltraPLONK arithmetisation](background/upa.md)
+  - [UltraPLONK arithmetization](background/upa.md)
   - [Polynomial commitment using inner product argument](background/pc-ipa.md)
   - [Recursion](background/recursion.md)
--- a/book/src/SUMMARY.md
+++ b/book/src/SUMMARY.md
@@ -3,7 +3,7 @@
 [halo2](README.md)
 - [Concepts](concepts.md)
   - [Proof systems](concepts/proofs.md)
-  - [UltraPLONK Arithmetization](concepts/arithmetization.md)
+  - [UltraPLONK arithmetization](concepts/arithmetization.md)
   - [Cores](concepts/cores.md)
   - [Chips](concepts/chips.md)
   - [Gadgets](concepts/gadgets.md)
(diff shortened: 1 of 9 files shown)


steven-ecc:
Hi! Thanks for this contribution.

Could you please sign the [contributor agreement](https://github.com/zcash/halo2/blob/main/Contributor_Agreement) and send it to me at steven@electriccoin.co? Signing the agreement allows us to redistribute the code later under another license if we need to, without having to get your permission. This is similar to how contributor agreements work in Apache software projects as well.

book/src/background/fields.md
```diff
@@ -3,7 +3,7 @@
 A fundamental component of many cryptographic protocols is the algebraic structure known
 as a [field]. Fields are sets of objects (usually numbers) with two associated binary
 operators $+$ and $\times$ such that various [field axioms][field-axioms] hold. The real
-numbers $\mathbb{R}$ are an example of a field with an uncountably infinite number of
+numbers $\mathbb{R}$ are an example of a field with an uncountable number of
```
therealyingtong, line 6:
```suggestion
numbers $\mathbb{R}$ are an example of a field with an uncountably infinite number of
```
We do mean to say "uncountably infinite" here: https://mathworld.wolfram.com/UncountablyInfinite.html

therealyingtong (changes requested):

book/src/background/fields.md
kevaundray: Yep, this was just a small nit for readability, as uncountability implies the set is infinite. 

I can change it back if you think the previous reading is clearer or better terminology wise.

book/src/background/fields.md
```diff
@@ -3,7 +3,7 @@
...
 elements.
```
daira, line 7:
```suggestion
numbers $\mathbb{R}$ are an example of a field with uncountably many elements.
```

book/src/SUMMARY.md
```diff
@@ -3,7 +3,7 @@
 [halo2](README.md)
 - [Concepts](concepts.md)
   - [Proof systems](concepts/proofs.md)
-  - [UltraPLONK Arithmetization](concepts/arithmetization.md)
+  - [UltraPLONK arithmetization](concepts/arithmetization.md)
```
daira, line 6: This is intentionally capitalized because it has the abbreviation UPA.

book/src/SUMMARY.md
```diff
@@ -31,6 +31,6 @@
   - [Polynomials](background/polynomials.md)
   - [Cryptographic groups](background/groups.md)
   - [Elliptic curves](background/curves.md)
-  - [UltraPLONK arithmetisation](background/upa.md)
+  - [UltraPLONK arithmetization](background/upa.md)
```
daira, line 34:
```suggestion
  - [UltraPLONK Arithmetization](background/upa.md)
```

daira:
Merged into #320. I don't think we need a contributor agreement for fixing typos.

PR #225, opened 2021-03-02
examples: Add cost model

### LINKED ISSUE #47: Performance modelling
Define a parameterized model that can predict the prover and verifier performance of a given Halo 2 circuit, so that I can understand how to optimize circuits. The model needs to be validated against measured performance.

str4d:
Usage: `cargo run --example cost-model -- --help`

Closes #47.


daira (approved):
ACK. I think this is good enough to merge and we should refine it in follow-up PRs.

str4d:
Agreed. I've opened #266 and #267 for the remaining work.

PR #226, opened 2021-03-02
Fix Group::random implementation for Pallas and Vesta

str4d:
Also removes the unused `CurveAffine::from_bytes_wide` and `CurveAffine::to_bytes_wide` methods.

### DIFF
--- a/src/pasta/curves.rs
+++ b/src/pasta/curves.rs
@@ -65,12 +65,6 @@ macro_rules! new_curve_impl {
                     let x = $base::random(&mut rng);
                     let ysign = (rng.next_u32() % 2) as u8;
 
-                    if x.is_zero() && ysign == 0 {
-                        // This would be the identity if encoded, and Group::random must
-                        // sample from the non-identity elements of the group.
-                        continue;
-                    }
-
                     let x3 = x.square() * x;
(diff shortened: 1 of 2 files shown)


ebfull:
As far as I can tell `CurveAffine::from_bytes_wide` is not broken. It's just that it was being misused in the `Group` trait implementation. If you sample random 64 bytes it will likely not be an uncompressed encoding of a curve point.

src/pasta/curves.rs
```diff
@@ -62,13 +62,27 @@ macro_rules! new_curve_impl {
...
-                        }
+                    let x = $base::random(&mut rng);
+                    let ysign = (rng.next_u32() % 2) as u8;
+
+                    if x.is_zero() && ysign == 0 {
+                        // This would be the identity if encoded, and Group::random must
+                        // sample from the non-identity elements of the group.
+                        continue;
+                    }
+
```
ebfull, line 73:
```suggestion
```

ebfull (changes requested):

src/pasta/curves.rs
ebfull: x = 0 and y = 0 are not valid curve coordinates

ebfull (approved):

PR #228, opened 2021-03-03
Migrate to the pasta_curves crate

str4d:
This enables us to decouple further improvements to the Pasta implementation (e.g. GPU backend) from development of the `halo2` crate, and also enables using `pasta_curves` in places where we don't need a `halo2` dependency, such as RedPallas.


daira (approved):
utACK

str4d:
The corresponding PR to create `pasta_curves` is https://github.com/zcash/pasta_curves/pull/3

Cargo.toml
```diff
@@ -40,16 +40,14 @@ harness = false
 
 [dependencies]
 backtrace = { version = "0.3", optional = true }
-subtle = "2.3"
 crossbeam-utils = "0.8"
 ff = "0.9"
 group = "0.9"
 metrics = "0.14.2"
 num_cpus = "1.13"
+pasta_curves = { git = "https://github.com/zcash/pasta_curves.git", branch = "main" }
```
ebfull, line 48: Can this be specific to the revision currently on `main`? We don't want changes to `pasta_curves`'s git repo to break halo2 code suddenly.

ebfull (changes requested):

Cargo.toml
str4d: Yep, I was waiting for the other PR to merge. I'll update this after https://github.com/zcash/pasta_curves/pull/2 is merged.

ebfull (approved):

PR #229, opened 2021-03-03
Remove metrics crate and inline modelling

str4d:
It is causing problems in (some) downstream crates, and we won't want
these annotations in the codebase anyway if a downstream crate is
collecting metrics. We'll be replacing this with the cost model binary
in https://github.com/zcash/halo2/pull/225 shortly.


porcuquine:
Confirmed this solves my downstream problem. Thank you.

ebfull (approved):

PR #234, opened 2021-03-29
[book] Point addition and compression background

therealyingtong:


book/src/background/curves.md
```diff
@@ -191,7 +201,36 @@ complicated, but when the curve has a prime $q$ number of points (and thus a pri
...
+As mentioned in the [Fields](./fields.md) section, we can interpret the least significant
+bit of a field element as its "sign", since its additive inverse will always have the
+opposite LSB. So we record the LSB of the $y$-coordinate as `sign`.
+
+Pallas and Vesta are defined over the $\mathbb{F}_p$ and $\mathbb{F}_q$ fields, which
+elements can be expressed in $255$ bits. This conveniently leaves one unused bit in a
+32-byte representation. We pack the $y$-coordinate `sign` bit into the highest bit in
+the representation of the $x$-coordinate:
+
+```
```
str4d, line 218:
```suggestion
```text
```

book/src/background/curves.md
```diff
@@ -191,7 +201,36 @@ complicated, but when the curve has a prime $q$ number of points (and thus a pri
...
+          ^                <------------------------------------->                 ^
+         LSB                              30 bytes                                MSB
+```
+
+### Deserialization
+When deserializing a compressed curve point, we first read the most significant bit as
+`ysign`, the sign of the $y$-coordinate. Then, we set this bit to zero to recover the
+original $x$-coordinate.
+
+If $x = 0, y = 0,$ we return the additive identity $(0, 0, 0)$. Otherwise, we proceed
```
str4d, line 230:
```suggestion
If $x = 0, y = 0,$ we return the "point at infinity" $\mathcal{O}$. Otherwise, we proceed
```

book/src/background/curves.md
```diff
@@ -191,7 +201,36 @@ complicated, but when the curve has a prime $q$ number of points (and thus a pri
...
+32-byte representation. We pack the $y$-coordinate `sign` bit into the highest bit in
+the representation of the $x$-coordinate:
+
+```
+         <----------------------------------- x --------------------------------->
+Enc(P) = [_ _ _ _ _ _ _ _] [_ _ _ _ _ _ _ _] ... [_ _ _ _ _ _ _ _] [_ _ _ _ _ _ _ sign]
+          ^                <------------------------------------->                 ^
+         LSB                              30 bytes                                MSB
+```
+
```
str4d, line 224:
```suggestion

The "point at infinity" $\mathcal{O}$ that serves as the group identity, does not have an
affine $(x, y)$ representation. However, it turns out that there are no points on either
the Pallas or Vesta curve with $x = 0$ or $y = 0$. We therefore use the "fake" affine
coordinates $(0, 0)$ to encode $\mathcal{O}$, which results in the all-zeroes 32-byte
array.

```

book/src/background/curves.md
```diff
@@ -152,17 +152,27 @@ corresponding numerators such that $X/Z = x'$ and $Y/Z = y'$. This completely av
 need to perform an inversion when doubling, and something analogous to this can be done
 when adding two distinct points.
 
-### TODO: Point addition
+### Point addition
+We now add two points with distinct $x$-coordinates, $P = (x_0, y_0)$ and $Q = (x_1, y_1),$
+where $x_0 \neq x_1$, to obtain $R = P + Q = (x_2, y_2).$ The line $\overline{PQ}$ has slope
+$\lambda = (y_1 - y_0)/(x_1 - x_0) \implies y - y_0 = \lambda \cdot (x - x_0).$
```
str4d, line 158:
```suggestion
$$\lambda = frac{y_1 - y_0}{x_1 - x_0} \implies y - y_0 = \lambda \cdot (x - x_0).$$
```

book/src/background/curves.md
```diff
@@ -152,17 +152,27 @@ corresponding numerators such that $X/Z = x'$ and $Y/Z = y'$. This completely av
 need to perform an inversion when doubling, and something analogous to this can be done
 when adding two distinct points.
 
-### TODO: Point addition
+### Point addition
+We now add two points with distinct $x$-coordinates, $P = (x_0, y_0)$ and $Q = (x_1, y_1),$
+where $x_0 \neq x_1$, to obtain $R = P + Q = (x_2, y_2).$ The line $\overline{PQ}$ has slope
```
str4d, line 157:
```suggestion
where $x_0 \neq x_1,$ to obtain $R = P + Q = (x_2, y_2).$ The line $\overline{PQ}$ has slope
```

str4d (approved):
ACK after patches.

PR #239, opened 2021-04-09
Book: rename "chip" to "Config"

daira:


therealyingtong:
Closing because we have since renamed `Chip` -> `Config` -> `Core`

PR #242, opened 2021-04-13
Add support for full-field challenge space

### LINKED ISSUE #151: Add support for full-field challenge space

therealyingtong:
Closes #151

### DIFF
--- a/src/plonk/lookup/prover.rs
+++ b/src/plonk/lookup/prover.rs
@@ -75,9 +75,8 @@ impl<F: FieldExt> Argument<F> {
     pub(in crate::plonk) fn commit_permuted<
         'a,
         C,
-        I,
-        E: EncodedChallenge<C, I>,
-        T: TranscriptWrite<C, I, E>,
+        E: EncodedChallenge<C>,
+        T: TranscriptWrite<C, E>,
     >(
         &self,
         pk: &ProvingKey<C>,
@@ -250,11 +249,7 @@ impl<C: CurveAffine> Permuted<C> {
     /// grand product polynomial over the lookup. The grand product polynomial
     /// is used to populate the Product<C> struct. The Product<C> struct is
     /// added to the Lookup and finally returned by the method.
-    pub(in crate::plonk) fn commit_product<
-        I,
-        E: EncodedChallenge<C, I>,
-        T: TranscriptWrite<C, I, E>,
-    >(
+    pub(in crate::plonk) fn commit_product<E: EncodedChallenge<C>, T: TranscriptWrite<C, E>>(
         self,
         pk: &ProvingKey<C>,
         params: &Params<C>,
@@ -498,7 +493,7 @@ impl<'a, C: CurveAffine> Committed<C> {
 }
 
 impl<C: CurveAffine> Constructed<C> {
-    pub(in crate::plonk) fn evaluate<I, E: EncodedChallenge<C, I>, T: TranscriptWrite<C, I, E>>(
+    pub(in crate::plonk) fn evaluate<E: EncodedChallenge<C>, T: TranscriptWrite<C, E>>(
         self,
         pk: &ProvingKey<C>,
         x: ChallengeX<C>,
--- a/src/plonk/lookup/prover.rs
+++ b/src/plonk/lookup/prover.rs
@@ -10,7 +10,7 @@ use crate::{
         multiopen::ProverQuery,
         Coeff, EvaluationDomain, ExtendedLagrangeCoeff, LagrangeCoeff, Polynomial, Rotation,
     },
-    transcript::{ChallengeSpace, TranscriptWrite},
+    transcript::{EncodedChallenge, TranscriptWrite},
 };
 use ff::Field;
 use group::Curve;
@@ -72,7 +72,13 @@ impl<F: FieldExt> Argument<F> {
     /// - constructs Permuted<C> struct using permuted_input_value = A', and
     ///   permuted_table_expression = S'.
     /// The Permuted<C> struct is used to update the Lookup, and is then returned.
-    pub(in crate::plonk) fn commit_permuted<'a, C, S: ChallengeSpace<C>, T: TranscriptWrite<C, S>>(
+    pub(in crate::plonk) fn commit_permuted<
+        'a,
+        C,
+        I,
+        E: EncodedChallenge<C, I>,
+        T: TranscriptWrite<C, I, E>,
+    >(
         &self,
         pk: &ProvingKey<C>,
         params: &Params<C>,
@@ -244,7 +250,11 @@ impl<C: CurveAffine> Permuted<C> {
     /// grand product polynomial over the lookup. The grand product polynomial
     /// is used to populate the Product<C> struct. The Product<C> struct is
     /// added to the Lookup and finally returned by the method.
-    pub(in crate::plonk) fn commit_product<S: ChallengeSpace<C>, T: TranscriptWrite<C, S>>(
+    pub(in crate::plonk) fn commit_product<
+        I,
+        E: EncodedChallenge<C, I>,
+        T: TranscriptWrite<C, I, E>,
+    >(
         self,
         pk: &ProvingKey<C>,
         params: &Params<C>,
@@ -488,7 +498,7 @@ impl<'a, C: CurveAffine> Committed<C> {
 }
 
 impl<C: CurveAffine> Constructed<C> {
-    pub(in crate::plonk) fn evaluate<S: ChallengeSpace<C>, T: TranscriptWrite<C, S>>(
+    pub(in crate::plonk) fn evaluate<I, E: EncodedChallenge<C, I>, T: TranscriptWrite<C, I, E>>(
         self,
         pk: &ProvingKey<C>,
         x: ChallengeX<C>,
--- a/src/plonk/lookup/prover.rs
+++ b/src/plonk/lookup/prover.rs
@@ -10,7 +10,7 @@ use crate::{
         multiopen::ProverQuery,
         Coeff, EvaluationDomain, ExtendedLagrangeCoeff, LagrangeCoeff, Polynomial, Rotation,
     },
-    transcript::TranscriptWrite,
+    transcript::{ChallengeSpace, TranscriptWrite},
 };
 use ff::Field;
 use group::Curve;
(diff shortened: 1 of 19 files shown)


src/plonk/lookup/prover.rs
```diff
@@ -72,12 +69,12 @@ impl<F: FieldExt> Argument<F> {
...
     ///   permuted_table_expression = S'.
     /// The Permuted<C> struct is used to update the Lookup, and is then returned.
-    pub(in crate::plonk) fn commit_permuted<'a, C, T: TranscriptWrite<C>>(
+    pub(in crate::plonk) fn commit_permuted<'a, C, S: ChallengeSpace<C>, T: TranscriptWrite<C, S>>(
         &self,
         pk: &ProvingKey<C>,
         params: &Params<C>,
         domain: &EvaluationDomain<C::Scalar>,
-        theta: ChallengeTheta<C>,
+        theta: ChallengeScalar<C>,
```
daira, line 77: This could be `ChallengeScalar<C, ChallengeScalarType::Theta>` where `ChallengeScalarType::Theta` is a marker type.

src/transcript.rs
```diff
@@ -233,37 +311,47 @@ impl<C: CurveAffine, Type> From<Challenge> for ChallengeScalar<C, Type> {
...
+
+impl<C: CurveAffine> ChallengeSpace<C> for ChallengeScalarFull<C> {
+    /// This algorithm applies the mapping of Algorithm 1 from the
+    /// [Halo](https://eprint.iacr.org/2019/1021) paper.
+    fn to_challenge_scalar(
+        challenge: Challenge,
+        challenge_type: ChallengeScalarType,
+    ) -> ChallengeScalar<C> {
+        ChallengeScalar {
+            inner: C::Scalar::from_u128(challenge.0),
```
daira, line 333: This doesn't look full-width. The difference between endoscaling and plain scaling is actually orthogonal to the width. For both, we probably need a width (i.e. entropy input) greater than 128 bits to achieve 128-bit security.

src/plonk/lookup/prover.rs
therealyingtong: You're right, it was unnecessary to introduce a `ChallengeScalarType` enum. I've force-pushed to undo that change and preserve the old type distinction.

src/plonk/prover.rs
```diff
@@ -241,7 +245,7 @@ pub fn create_proof<C: CurveAffine, T: TranscriptWrite<C>, ConcreteCircuit: Circ
         .collect::<Result<Vec<_>, _>>()?;
 
     // Sample theta challenge for keeping lookup columns linearly independent
-    let theta = ChallengeTheta::get(transcript);
+    let theta = transcript.squeeze_challenge_scalar();
```
daira, line 248:
```suggestion
    let theta: ChallengeTheta = transcript.squeeze_challenge_scalar();
```
Otherwise, we lose some of the advantage of the explicit typing, because the compiler could infer done other challenge type if we were to use `theta` in the wrong place. Similarly for the other challenges.

src/plonk/verifier.rs
```diff
@@ -50,7 +47,7 @@ pub fn verify_proof<'a, C: CurveAffine, T: TranscriptRead<C>>(
         .collect::<Result<Vec<_>, _>>()?;
 
     // Sample theta challenge for keeping lookup columns linearly independent
-    let theta = ChallengeTheta::get(transcript);
+    let theta = transcript.squeeze_challenge_scalar();
```
daira, line 50:
```suggestion
    let theta: ChallengeTheta = transcript.squeeze_challenge_scalar();
```

ebfull (changes requested):
My ideal API looks something like this:

```rust
/// `EncodedChallenge<C, I>` defines a challenge encoding where `I` is the input
/// that is used to derive the challenge encoding and `get_challenge` obtains
/// the _real_ `C::Scalar` that the challenge encoding represents.
pub trait EncodedChallenge<C: CurveAffine, I> {
    fn new(challenge_input: &I) -> Self;
    fn get_challenge(challenge: &Self) -> C::Scalar;
}
```

Then you can have (for blake2b or other hash functions) the challenge encoding

```rust
struct Challenge255([u8; 32]);
impl<C: CurveAffine> EncodedChallenge<C, [u8; 64]> for Challenge255 {
    fn new(challenge_input: &[u8; 64]) -> Self {
        Challenge255(C::Scalar::from_bytes_wide(challenge_input).to_bytes())
    }
    fn get_scalar(challenge: &Self) -> C::Scalar {
        C::Scalar::from_bytes(challenge.0)
    }
}
```

which takes a 64-byte input from the hash function and turns it into a 32-byte scalar directly, but in the case of Poseidon/recursion we can have a challenge encoding

```rust
struct Challenge128(u128);
impl<C: CurveAffine> EncodedChallenge<C, C::Base> for Challenge128 {
    fn new(challenge_input: &C::Base) -> Self {
        Challenge128(challenge_input.get_lower_128())
    }
    fn get_scalar(challenge: &Self) -> C::Scalar {
        let mut acc = (C::Scalar::ZETA + &C::Scalar::one()).double();

        for i in (0..64).rev() {
            let should_negate = ((challenge.0 >> ((i << 1) + 1)) & 1) == 1;
            let should_endo = ((challenge.0 >> (i << 1)) & 1) == 1;

            let q = if should_negate {
                -C::Scalar::one()
            } else {
                C::Scalar::one()
            };
            let q = if should_endo { q * &C::Scalar::ZETA } else { q };
            acc = acc + &q + &acc;
        }

        acc
    }
}
```

which takes a base field element (the output of poseidon) and turns it into a 128-bit challenge string, which can then be used to derive the actual scalar via simulated endoscaling.

src/plonk/prover.rs
therealyingtong: I think the compiler is inferring the type `theta: ChallengeTheta` here from the signature of `lookup.commit_permuted()` (on line 260) below.

daira: It is, but we can't see that in the code of this function. If we called a variable `theta` but accidentally passed it to an API expecting `beta`, for example, then the compiler wouldn't catch that. It would just infer the type as `ChallengeBeta`. I want to be able to know *when reading the declaration* that the compiler would catch any attempt to pass this variable in the wrong place.

src/transcript.rs
```diff
@@ -182,45 +212,75 @@ impl<W: Write, C: CurveAffine> Transcript<C> for Blake2bWrite<W, C> {
...
+    /// Cast an encoded challenge as a typed `ChallengeScalar`.
+    fn as_challenge_scalar<T>(&self) -> ChallengeScalar<C, T> {
+        ChallengeScalar {
+            inner: self.get_scalar(),
+            _marker: PhantomData,
+        }
+    }
+}
+
+/// A 128-bit challenge.
```
daira, line 260:
```suggestion
/// A 128-bit challenge. Note that using this challenge space may result
/// in less than 128-bit security.
```

daira (changes requested):
Changes requested as per comments.

src/poly/commitment/prover.rs
```diff
@@ -53,11 +53,11 @@ pub fn create_proof<C: CurveAffine, T: TranscriptWrite<C>>(
     // Challenge that will ensure that the prover cannot change P but can only
     // witness a random polynomial commitment that agrees with P at x, with high
     // probability.
-    let iota = *ChallengeScalar::<C, ()>::get(transcript);
+    let iota = *transcript.squeeze_challenge_scalar::<()>();
```
daira, line 56:
```suggestion
    let iota: ChallengeIota<()> = *transcript.squeeze_challenge_scalar();
```

src/transcript.rs
```diff
@@ -47,13 +58,14 @@ pub trait TranscriptWrite<C: CurveAffine>: Transcript<C> {
 
 /// We will replace BLAKE2b with an algebraic hash function in a later version.
 #[derive(Debug, Clone)]
-pub struct Blake2bRead<R: Read, C: CurveAffine> {
+pub struct Blake2bRead<R: Read, C: CurveAffine, E: EncodedChallenge<C, [u8; 64]>> {
     state: Blake2bState,
     reader: R,
     _marker: PhantomData<C>,
```
daira, line 64:
```suggestion
    _marker_c: PhantomData<C>,
```

src/transcript.rs
```diff
@@ -63,11 +75,14 @@ impl<R: Read, C: CurveAffine> Blake2bRead<R, C> {
                 .to_state(),
             reader,
             _marker: PhantomData,
```
daira, line 77:
```suggestion
            _marker_c: PhantomData,
```

src/transcript.rs
```diff
@@ -113,24 +137,18 @@ impl<R: Read, C: CurveAffine> Transcript<C> for Blake2bRead<R, C> {
...
-    }
 }
 
 /// We will replace BLAKE2b with an algebraic hash function in a later version.
 #[derive(Debug, Clone)]
-pub struct Blake2bWrite<W: Write, C: CurveAffine> {
+pub struct Blake2bWrite<W: Write, C: CurveAffine, E: EncodedChallenge<C, [u8; 64]>> {
     state: Blake2bState,
     writer: W,
     _marker: PhantomData<C>,
```
daira, line 147:
```suggestion
    _marker_c: PhantomData<C>,
```

src/transcript.rs
```diff
@@ -140,6 +158,7 @@ impl<W: Write, C: CurveAffine> Blake2bWrite<W, C> {
                 .to_state(),
             writer,
             _marker: PhantomData,
```
daira, line 160:
```suggestion
            _marker_c: PhantomData,
```

daira (approved):
utACK with non-blocking nits.

src/transcript.rs
```diff
@@ -182,88 +207,90 @@ impl<W: Write, C: CurveAffine> Transcript<C> for Blake2bWrite<W, C> {
...
-#[derive(Copy, Clone, Debug)]
-pub struct Challenge(pub(crate) u128);
-
-impl Challenge {
-    /// Obtains a new challenge from the transcript.
-    pub fn get<C: CurveAffine, T: Transcript<C>>(transcript: &mut T) -> Challenge {
-        Challenge(transcript.squeeze_challenge().get_lower_128())
-    }
+/// `Challenge` trait implemented for challenges of different lengths
+pub trait Challenge: Copy + Clone + std::fmt::Debug {
```
ebfull, line 213: Where is this trait being used? Can it be removed?

ebfull (changes requested):

ebfull (approved):

PR #253, opened 2021-04-19
Move config responsibilities from `Layouter` to `Chip`

### LINKED ISSUE #235: [Circuit] Implement Core/Chip abstraction
Related to #195 and #117.

Needed for Sinsemilla gadget https://github.com/zcash/orchard/pull/57.

therealyingtong:
Closes #235 

Review recommendations:
- the most informative example to run is `cargo run --example two-chip`;
- the `sha256` example is commented out here and fixed in #256 

`trait Chip<F: FieldExt>` is responsible for:
- configuring the underlying constraint system (i.e. creating columns, gates, permutations, lookups); 
- holding on to its own configuration: where previously we had `layouter().config()` we now have `self.config()`; and
- loading any fixed configuration needed into the circuit at the beginning of circuit synthesis.

The "top-level" chip is simply the chip with the most comprehensive config, i.e. it provides all columns and permutations needed by other chips.

`trait Layouter<F: FieldExt>` is responsible for:
- `assign_region(&mut self)` assigning values according to its own defined strategy;
- namespacing;

The layouter knows nothing about any config. This is fine because when a chip calls `layouter.assign_region()`, it has access to `self.config()` to inform the layouter. So we no longer see a `Layouter<C: Chip>` generic type, and this removal also trickles down to `RegionLayouter`, `Region`.


src/circuit.rs
```diff
@@ -1,41 +1,85 @@
...
+    /// We allow for the chip to use pre-existing columns, selectors, and
+    /// permutations in its configuration. Chips that are lower in a hierarchy
+    /// depend on other chips to pass down these objects.
+    fn configure(
+        &mut self,
+        meta: &mut ConstraintSystem<F>,
+        selectors: BTreeMap<&str, Selector>,
+        columns: BTreeMap<&str, Column<Any>>,
+        perms: BTreeMap<&str, Permutation>,
+    ) -> Self::Config;
```
str4d, line 67: Does this need to be generic on the `Chip` trait? More specifically, when would we call this function generically, rather than somewhere we know the concrete chip type? If we know the concrete type, then we can use a configuration function that takes a typed configuration (either as multiple function arguments or a single struct argument) instead of a `BTreeMap`.

src/circuit.rs
```diff
@@ -1,41 +1,85 @@
...
     /// [`Circuit::synthesize`]: crate::plonk::Circuit::synthesize
-    type Loaded: fmt::Debug;
+    type Loaded: Loaded;
+
+    /// A new chip that is not yet configured or loaded.
+    fn new() -> Self;
 
-    /// The field that the chip is defined over.
+    /// A chip constructed from a given config and loaded state.
+    fn construct(config: Self::Config, loaded: Self::Loaded) -> Self;
```
str4d, line 53: Likewise, in what context are these used generically, without knowledge of the concrete chip type?

src/circuit.rs
```diff
@@ -1,41 +1,85 @@
...
 
-    /// Load any fixed configuration for this chip into the circuit.
+    /// The chip holds its own configuration.
+    fn config(&self) -> &Self::Config;
+
+    /// Load any fixed configuration for this chip into the circuit, i.e.
+    /// assigns cells in fixed columns.
+    ///
+    /// `self.loaded()` will panic if called inside this function.
+    fn load(&mut self, layouter: &mut impl Layouter<F>) -> Result<Self::Loaded, Error>;
```
str4d, line 76: Now that this is a method on `Chip`, if it is going to take `&mut self` then the type signature should be:
```suggestion
    fn load(&mut self, layouter: &mut impl Layouter<F>) -> Result<(), Error>;
```

src/circuit.rs
```diff
@@ -1,41 +1,85 @@
...
+
+    /// Provides access to general chip state loaded at the beginning of circuit
+    /// synthesis.
     ///
-    /// `layouter.loaded()` will panic if called inside this function.
-    fn load(layouter: &mut impl Layouter<Self>) -> Result<Self::Loaded, Error>;
+    /// Panics if called inside `Chip::load`.
```
str4d, line 81: I think this is more accurately:
```suggestion
    /// Panics if called before `Chip::load`.
```

src/circuit.rs
```diff
@@ -187,12 +237,13 @@ pub trait Layouter<C: Chip> {
     ///
     /// ```ignore
     /// fn assign_region(&mut self, || "region name", |region| {
-    ///     region.assign_advice(self.config.a, offset, || { Some(value)});
+    ///     let config = chip.config().clone();
```
str4d, line 240: I don't believe the `.clone()` should be necessary, now that `chip` and `layouter` are separate variables (so we aren't trying to use an immutable reference in a method that takes a mutable reference to `self`).

src/circuit/layouter.rs
```diff
@@ -67,58 +68,43 @@ pub trait RegionLayouter<C: Chip>: fmt::Debug {
...
-        let mut ret = SingleChip {
+    pub fn new(cs: &'a mut CS) -> Result<Self, Error> {
+        let ret = SingleChipLayouter {
             cs,
-            config,
-            loaded: None,
             regions: vec![],
             columns: HashMap::default(),
-            _marker: PhantomData,
+            marker: PhantomData,
```
str4d, line 96: Undo this change (markers are never used, so to prevent lint errors):
```suggestion
            _marker: PhantomData,
```

str4d (commented):
Partial review.

src/circuit.rs
therealyingtong: i've moved `new()`, `construct()`, `configure()` and `load()` out of the `Chip` trait. they are now implemented for each concrete chip type instead.

examples/simple-example.rs
```diff
@@ -126,46 +176,31 @@ impl<F: FieldExt> FieldChip<F> {
...
-        value: Option<Self::Field>,
+        &self,
+        mut layouter: impl Layouter<F>,
+        value: Option<F>,
     ) -> Result<Self::Num, Error> {
-        let config = layouter.config().clone();
+        let config = match self.config() {
+            FieldConfigEnum::Config(config) => config.clone(),
+            _ => unreachable!(),
+        };
```
str4d, line 202: After the removal of `FieldConfigEnum`, this can just be:
```suggestion
        let config = self.config();
```
or
```suggestion
        let config = &self.config;
```
(and ditto below).

examples/simple-example.rs
```diff
@@ -275,7 +319,7 @@ struct MyCircuit<F: FieldExt> {
 
 impl<F: FieldExt> Circuit<F> for MyCircuit<F> {
     // Since we are using a single chip for everything, we can just reuse its config.
-    type Config = FieldConfig;
+    type Config = FieldConfigEnum;
```
str4d, line 322:
```suggestion
    type Config = FieldConfig;
```

I think this could also maybe instead be `type Chip = FieldChip`, since I _think_ now we'll always just construct the chip from its config immediately inside `Circuit::synthesize`. But that's a higher-level refactor (and would probably also require making the layouter choice an associated type), so let's leave that out of this PR.

examples/two-chip.rs
```diff
@@ -0,0 +1,723 @@
...
+
+    fn configure(meta: &mut ConstraintSystem<F>) -> Self::Config {
+        // We create the two advice columns that FieldChip uses for I/O.
+        let advice = [meta.advice_column(), meta.advice_column()];
+
+        // We also need an instance column to store public inputs.
+        let instance = meta.instance_column();
+
+        let mut field_chip = FieldChip::new();
+        field_chip.configure(meta, advice, instance)
```
str4d, line 659:
```suggestion
        FieldChip::configure(meta, advice, instance)
```
(and associated changes here, similar to `simple-example.rs`)

str4d (changes requested):

src/circuit.rs
```diff
@@ -85,6 +103,22 @@ pub struct Cell {
...
+    pub value: Option<T>,
+}
+
+impl<T> CellValue<T> {
+    /// Construct a `CellValue`.
+    pub fn new(cell: Cell, value: Option<T>) -> Self {
+        CellValue { cell, value }
+    }
+}
+
```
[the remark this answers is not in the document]
therealyingtong: Anecdotally this comes in handy in downstream gadgets, e.g. in Poseidon https://github.com/zcash/orchard/pull/28/files#diff-60e7cd15724aefcc52b9091b71328e342d92757d020aadc4824e94acafffda18R412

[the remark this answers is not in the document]
daira: I don't see `CellValue` used there?

examples/sha256/main.rs
```diff
@@ -1,162 +1,162 @@
...
-    pub fn digest(
-        mut layouter: impl Layouter<Sha256Chip>,
-        data: &[Sha256Chip::BlockWord],
-    ) -> Result<Sha256Digest<Sha256Chip::BlockWord>, Error> {
-        let mut hasher = Self::new(layouter.namespace(|| "init"))?;
-        hasher.update(layouter.namespace(|| "update"), data)?;
-        hasher.finalize(layouter.namespace(|| "finalize"))
-    }
-}
+// //! Gadget and chips for the [SHA-256] hash function.
```
daira, line 1: Use a block comment (`/* ... */`) so that when #256 fixes the example, we can see the actual changes in the diff.

examples/two-chip.rs
```diff
@@ -0,0 +1,636 @@
...
+// ANCHOR: add-instructions-impl
+impl<F: FieldExt> AddInstructions<F> for FieldChip<F> {
+    type Num = Number<F>;
+    fn add(
+        &self,
+        layouter: impl Layouter<F>,
+        a: Self::Num,
+        b: Self::Num,
+    ) -> Result<Self::Num, Error> {
+        let config = self.config().add_config.clone();
```
daira, line 204: Is it possible to avoid these clones? I tried but the lifetime requirements defeated me.

daira (approved):
utACK

examples/two-chip.rs
str4d: These ones, I think not (at least if we want the sub-cores to be usable on their own). We still have a large reduction in config cloning from this PR.

str4d (approved):
utACK

PR #256, opened 2021-04-23
Fix `sha256` example for new config abstraction

therealyingtong:
Split out from #253 to reduce the diff.

To test: `cargo test --example sha256 --all-features`

### DIFF
--- a/examples/sha256/table16.rs
+++ b/examples/sha256/table16.rs
@@ -187,13 +187,8 @@ impl<F: FieldExt> Table16Chip<F> {
             perm.clone(),
         );
 
-        let message_schedule = MessageScheduleConfig::configure(
-            meta,
-            lookup_inputs,
-            message_schedule,
-            extras,
-            perm,
-        );
+        let message_schedule =
+            MessageScheduleConfig::configure(meta, lookup_inputs, message_schedule, extras, perm);
 
         Table16Config {
             lookup,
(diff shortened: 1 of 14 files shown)


examples/sha256/table16.rs
```diff
@@ -158,82 +179,79 @@ impl<F: FieldExt> Table16Chip<F> {
...
             message_schedule,
             extras,
             perm.clone(),
         );
 
-        let message_schedule =
-            MessageSchedule::configure(meta, lookup_inputs, message_schedule, extras, perm);
+        let message_schedule = MessageScheduleConfig::configure(
+            meta,
+            lookup_inputs.clone(),
```
str4d, line 192: I don't believe this clone is necessary:
```suggestion
            lookup_inputs,
```

examples/sha256/table16.rs
```diff
@@ -158,82 +179,79 @@ impl<F: FieldExt> Table16Chip<F> {
...
     // message block and return the final state.
     fn compress(
-        layouter: &mut impl Layouter<Self>,
+        &self,
+        layouter: &mut impl Layouter<F>,
         initialized_state: &Self::State,
         input: [Self::BlockWord; super::BLOCK_SIZE],
     ) -> Result<Self::State, Error> {
-        let config = layouter.config().clone();
+        let config = self.config().clone();
```
str4d, line 240:
```suggestion
        let config = self.config();
```

str4d (approved):

PR #259, opened 2021-05-03
CI: Reduce noise of clippy lints

str4d:
We move Clippy lints into separate workflows that only run either directly on branches, or on trial-merges of PRs, to deduplicate the
lint annotations. We will still see stable lints duplicated in beta, but that's fine.


str4d:
We unintuitively get **extra** duplication here, because CI is being run against both the new workflows in this PR and the original workflows in `main`.

PR #261, opened 2021-05-12
[MockProver] Handle query rotations in lookups

therealyingtong:
The `MockProver` should handle query rotations for lookups, like the real prover does (https://github.com/zcash/halo2/blob/main/src/plonk/lookup/prover.rs#L100)

### DIFF
--- a/src/dev.rs
+++ b/src/dev.rs
@@ -339,21 +339,21 @@ impl<F: FieldExt> MockProver<F> {
                         &|index| {
                             let query = self.cs.fixed_queries[index];
                             let column_index = query.0.index();
-                            let rotation = query.1 .0;
+                            let rotation = query.1.0;
                             self.fixed[column_index]
                                 [(row as i32 + n + rotation) as usize % n as usize]
                         },
                         &|index| {
                             let query = self.cs.advice_queries[index];
                             let column_index = query.0.index();
-                            let rotation = query.1 .0;
+                            let rotation = query.1.0;
                             self.advice[column_index]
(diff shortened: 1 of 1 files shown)


src/dev.rs
```diff
@@ -337,16 +337,25 @@ impl<F: FieldExt> MockProver<F> {
...
+                            let rotation = query.1 .0;
+                            self.fixed[column_index]
+                                [(row as i32 + n + rotation) as usize % n as usize]
                         },
                         &|index| {
-                            let column_index = self.cs.advice_queries[index].0.index();
-                            self.advice[column_index][row as usize]
+                            let query = self.cs.advice_queries[index];
+                            let column_index = query.0.index();
+                            let rotation = query.1 .0;
```
nuttycom, line 349: Weird spacing here.

str4d, line 349:
```suggestion
                            let rotation = query.1.0;
```

src/dev.rs
```diff
@@ -337,16 +337,25 @@ impl<F: FieldExt> MockProver<F> {
...
+                            self.advice[column_index]
+                                [(row as i32 + n + rotation) as usize % n as usize]
                         },
                         &|index| {
-                            let column_index = self.cs.instance_queries[index].0.index();
-                            self.instance[column_index][row as usize]
+                            let query = self.cs.instance_queries[index];
+                            let column_index = query.0.index();
+                            let rotation = query.1 .0;
```
str4d, line 356:
```suggestion
                            let rotation = query.1.0;
```

str4d (approved):
utACK

PR #269, opened 2021-05-18
CI fixes

str4d:
Some of these slipped through while GitHub Actions was down.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions-rs/toolchain@v1
         with:
-          toolchain: stable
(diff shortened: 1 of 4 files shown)


src/dev.rs
```diff
@@ -339,21 +339,21 @@ impl<F: FieldExt> MockProver<F> {
                         &|index| {
                             let query = self.cs.fixed_queries[index];
                             let column_index = query.0.index();
-                            let rotation = query.1.0;
+                            let rotation = query.1 .0;
```
str4d, line 342: IDK what's going on here, but this is exactly what `rustfmt` wants.

str4d: Aaand after a rebase, now it's gone 🤷

PR #272, opened 2021-05-27
`Circuit` API changes

str4d:


src/plonk/circuit.rs
```diff
@@ -775,10 +771,72 @@ impl<F: Field> ConstraintSystem<F> {
...
+
+    /// Query a selector at a relative position.
+    pub fn query_selector(&mut self, selector: Selector, at: Rotation) -> Expression<F> {
+        self.queried_selectors.push(selector);
+        Expression::Fixed(self.meta.query_fixed_index(selector.0, at))
+    }
+
+    /// Query a fixed column at a relative position
+    pub fn query_fixed(&mut self, column: Column<Fixed>, at: Rotation) -> Expression<F> {
+        self.queried_columns.push(column.into());
```
str4d, line 808: Currently this only tracks the queried columns, but I'm not sure if that's sufficient for the kinds of debug checks we want to perform. I think we might actually want to track queried "virtual registers", being the tuple `(Column<Any>, Rotation)`.

str4d: Mmm, now that I think about it more, I think this is indeed what we'd want to track. Then when a selector is enabled on a row, we can figure out all of the relative cells that need to be assigned for this gate to be useful.

src/plonk/circuit.rs
```diff
@@ -460,14 +460,51 @@ impl<F> Mul<F> for Expression<F> {
...
+}
+
+#[derive(Clone, Debug)]
+pub(crate) struct Gate<F: Field> {
+    name: &'static str,
+    poly: Expression<F>,
+    /// We track queried selectors separately from other registers, so that we can use
+    /// them to trigger debug checks on gates.
+    queried_selectors: Vec<VirtualRegister>,
+    queried_registers: Vec<VirtualRegister>,
```
str4d, line 487: This might not be exactly how we need to store the selectors. What I want to do is:

- Track virtual registers (so we "know" the shape of the gate's tetris piece).
- Track which selectors are queried, at what relative offsets.
- Invert the `gate -> selectors` mapping into a `selector -> gates` mapping.
  - This might just be how we store it? Or maybe it's easier to just compute that in the dev tooling once we have all the gates.
- When a selector is enabled, look up the gates it was queried in. For each gate:
  - Use the selector's rotation to figure out where this instance of the gate is positioned.
  - Use the queried virtual registers to "mark" the corresponding absolute cells (or maybe the relative cells within a region) as "must be assigned to".
  - Return an error if not all marked cells are assigned to, with an error message indicating which gates are requiring the cell to be assigned.

therealyingtong: Another consideration: a gate could be toggled on `(1 - selector)`, i.e. only activated when the selector is NOT enable.

src/dev.rs
```diff
@@ -295,47 +295,47 @@ impl<F: FieldExt> MockProver<F> {
...
+                        move |index| {
+                            let (column, at) = &queries[index];
+                            let resolved_row = (row + at.0) % n;
+                            cells[column.index()][resolved_row as usize]
                         }
-                    })
-                });
+                    }
+
+                    if gate.poly().evaluate(
```
daira, line 319: [In the halo2 book](https://zcash.github.io/halo2/concepts/arithmetization.html) it says:
> Typically, a configuration will define polynomial constraints that are switched off and on by selectors defined in fixed columns. For example, a constraint q<sub>i</sub>⋅p(...) = 0 can be switched off for a particular row i by setting q<sub>i</sub> = 0. In this case we sometimes refer to a set of constraints controlled by a set of selector columns that are designed to be used together, as a ***gate***.

That definition of a gate is selector-centric and covers all polynomial constraints that are enabled by a set of selectors. Even if we simplified it to the constraints covered by a single selector, I think it wouldn't be the same concept as what is being called a `Gate` in the code here, which only refers to a single polynomial constraint.

Perhaps just rename `Gate` to `PolyConstraint`?

src/plonk/circuit.rs
```diff
@@ -460,14 +460,51 @@ impl<F> Mul<F> for Expression<F> {
 #[derive(Copy, Clone, Debug)]
 pub(crate) struct PointIndex(pub usize);
 
+/// A "virtual register" is a PLONK cell that has been queried at a particular relative
+/// offset within a custom gate.
```
daira, line 464: Why not call it a register? It's no more or less virtual than anything else 😝

src/plonk/circuit.rs
```diff
@@ -460,14 +460,51 @@ impl<F> Mul<F> for Expression<F> {
...
+
 /// This is a description of the circuit environment, such as the gate, column and
 /// permutation arrangements.
 #[derive(Debug, Clone)]
 pub struct ConstraintSystem<F: Field> {
     pub(crate) num_fixed_columns: usize,
     pub(crate) num_advice_columns: usize,
     pub(crate) num_instance_columns: usize,
-    pub(crate) gates: Vec<(&'static str, Expression<F>)>,
+    pub(crate) gates: Vec<Gate<F>>,
```
daira, line 507: The field name should probably change here, if the type name does.

src/plonk/circuit.rs
```diff
@@ -708,9 +710,22 @@ impl<F: Field> ConstraintSystem<F> {
     }
 
     /// Create a new gate
-    pub fn create_gate(&mut self, name: &'static str, f: impl FnOnce(&mut Self) -> Expression<F>) {
-        let poly = f(self);
-        self.gates.push((name, poly));
+    pub fn create_gate(
```
daira, line 713: Rename to `create_poly_constraint`.

src/plonk/circuit.rs
therealyingtong: Or how about `VirtualCell`? A cell that has yet to be allocated an absolute position.

src/plonk/circuit.rs
str4d: I think we could better track that if instead of using `1 - selector` we used `!selector` somewhere? In the expression would be nice, but that would require `Expression` itself to do the tracking, which could get messy. And that doesn't cover more complex cases like `(2 - selector)`.

src/plonk/circuit.rs
str4d: I'll go with `VirtualCell` for now, because I'd rather not use "register" on its own yet while I haven't figured out the downstream parts of the API change (currently I have concepts of "virtual register" and "concrete register" that I want to separate, but this might change).

src/dev.rs
str4d: That interpretation doesn't match the next sentence:

> Typically there will be a standard gate that supports generic operations like field multiplication and division

The standard PLONK gate is a single polynomial constraint containing a set of sub-constraints controlled by a set of selector columns. We currently refer to that as a `Gate`, and it is what we can create with `meta.create_gate`. That is a different interpretation than "a gate is the set of all polynomial constraints involving the same selector", which is what I think you are thinking of.

src/plonk/circuit.rs
str4d: By "could better track" I mean that `impl Neg for Selector` could update some map contained within `Selector` (either directly or by reference). That would be harder to do with `impl Sub<Selector> for Fp` which would need to interpret what that subtraction means. And in any case, the type of `selector` at this point is `Expression`.

src/dev.rs
daira: The standard PLONK gate does satisfy that definition. Its selectors are `sa`, `sb`, `sc`, and `sm`, and it has a (single-element) set of polynomial constraints controlled by those selectors.

str4d: Addressed this in #273.

src/plonk/circuit.rs
str4d: I've added the dev tooling for this in #274. As discussed in PR gardening, we'll leave this as-is for now (with `Selector` being a boolean selector that positively activates gates), and if we find we need negative-activation of gates or non-boolean selectors, we can add them.

therealyingtong (approved):
utACK

src/plonk/circuit.rs
```diff
@@ -175,19 +175,12 @@ impl TryFrom<Column<Any>> for Column<Instance> {
 /// }
 /// ```
 #[derive(Clone, Copy, Debug)]
```
therealyingtong, line 177: These additional derivations are not logically part of this PR, but they're needed downstream and could be conveniently added here.
```suggestion
#[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
```

therealyingtong (changes requested):

ebfull (approved):

PR #274, opened 2021-05-27
MockProver: Check that cells for active gates are assigned to

str4d:


src/dev.rs
```diff
@@ -18,9 +22,27 @@ mod graph;
...
+        column: Column<Any>,
+        /// The row in which this cell is located.
+        row: usize,
+        /// The index of the active gate. These indices are assigned in the order in which
+        /// `ConstraintSystem::create_gate` is called during `Circuit::configure`.
+        gate_index: usize,
+        /// The name of the active gate. These are specified by the gate creator (such as
+        /// a chip implementation), and may not be unique.
+        gate_name: &'static str,
+    },
```
therealyingtong, line 45: Could we have this be a Warning instead of a Failure? If the assignment passes all the constraints, it technically is still _a_ satisfying assignment (just maybe not the intended one)

str4d: I'm inclined to keep this as a failure for now; if someone needs this case to occur, they can explicitly handle that in their tests. We don't currently have a way to distinguish between warnings and errors without moving away from the `Result` type, so I'd rather think this through in a separate issue.

therealyingtong (approved):
utACK

ebfull (approved):

src/dev.rs
daira: I think it's completely fine for this to be an error. If the value doesn't matter then the circuit should explicitly assign 0, IMHO. Detecting potential circuit errors is more important than the minor convenience of being able to leave some cells (that are used in a gate) unassigned.

PR #276, opened 2021-05-27
Use Horner's rule in `eval_polynomial()`

therealyingtong:


str4d (approved):
utACK. I like how much simpler this ends up being!

str4d (changes requested):
Needs a `cargo fmt`

src/arithmetic.rs
```diff
@@ -304,13 +304,9 @@ fn parallel_fft<G: Group>(a: &mut [G], omega: G::Scalar, log_n: u32, log_cpus: u
 /// This evaluates a provided polynomial (in coefficient form) at `point`.
 pub fn eval_polynomial<F: Field>(poly: &[F], point: F) -> F {
     // TODO: parallelize?
```
daira, line 306: https://en.wikipedia.org/wiki/Horner%27s_method#Parallel_evaluation

PR #273, opened 2021-05-27
Allow multiple polynomial constraints per gate

str4d:
This enables a gate such as ECC complete addition to define its constraints in terms of a common set of queried columns.

PR #281, opened 2021-05-29
Fix sha-256 example to use new `create_gate()` API

therealyingtong:
This PR adjusts the SHA256 example to use the new `create_gate()` API which returns a vector of expressions (#273).


therealyingtong:
Superseded by https://github.com/zcash/halo2/pull/337.

PR #284, opened 2021-06-03
Document that batch inversion in arithmetic.rs leaves zero as zero

daira:
Document that batch inversion leaves zero as zero.


str4d (approved):
ACK. I confirmed this behaviour.

PR #285, opened 2021-06-04
dev: MockProver fixes and improvements

str4d:


src/dev.rs
```diff
@@ -43,16 +44,21 @@ pub enum VerifyFailure {
...
+    /// A constraint was not satisfied for a particular row.
+    Constraint {
+        /// The index of the gate containing the unsatisfied constraint. These indices are
+        /// assigned in the order in which `ConstraintSystem::create_gate` is called
+        /// during `Circuit::configure`.
         gate_index: usize,
-        /// The name of the gate that is not satisfied. These are specified by the gate
-        /// creator (such as a chip implementation), and may not be unique.
+        /// The name of the gate containing the unsatisfied constraint. This is specified
+        /// by the gate creator (such as a chip implementation), and may not be unique.
```
daira, line 54:
```suggestion
        /// by the gate creator (such as a chip implementation), and is not enforced to
        /// be unique.
```

src/dev.rs
```diff
@@ -379,7 +431,8 @@ impl<F: FieldExt> MockProver<F> {
                                     Any::Fixed => {
```
daira, line 431: (applies to lines 418-420 above)

I didn't know that selectors could be on a different row to the gate's logical row! That seems confusing. Is it necessary at all, or should we enforce that selectors have a row offset of 0?

str4d: Now that the `Selector` type consistently means a positive boolean selector, _maybe_ rotations are unnecessary for them? I'm fine with restricting this if desired.

daira: It is technically possible that there could be optimizations where you share selectors that are on different rows between gates, but the more I think about how confusing that would be, the more I think it should be prohibited. It also potentially interferes with visualizations where you click on a row and see the gates enabled for that row.

(If you *really* want to do those optimizations, you can still do them using fixed columns.)

str4d: I'll open a separate PR for this.

str4d: https://github.com/zcash/halo2/pull/287

ebfull (approved):

daira (commented):
Post-hoc ACK 483c872cbeb8956f87ec2cb33f71a4a3622b919c

PR #286, opened 2021-06-04
Enable annotating individual constraints within gates

str4d:
The closure passed to `ConstraintSystem::create_gate` can now return:

- Tuples of `(&'static str, Expression<F>)`
- Anything implementing `IntoIterator` (e.g. `Some(Expression<F>)`)

### DIFF
--- a/examples/simple-example.rs
+++ b/examples/simple-example.rs
@@ -364,6 +364,7 @@ fn main() {
             gate_index: 1,
             gate_name: "public input",
             constraint_index: 0,
+            constraint_name: "",
             row: 6,
         }])
     );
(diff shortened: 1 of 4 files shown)


str4d:
Some potential questions:
- With this design, all constraints within a gate need to be returned in the same way (with annotations on all of them, or none of them). I think this is fine?
- Gates require known-at-compile-time annotations, and previously these were 1:1 with constraints. So I think it's fine for the constraints themselves to also require static strings as annotations? If we wanted to have runtime-editable annotations then we'd need to return closures so we could keep string formatting out of the hot path.
- Is a tuple the nicest API for this? I think so.

str4d:
Pushed an extra commit because I realised that since #273 merged, it has been possible for `ConstraintSystem::create_gate` to create gates with no constraints, which would just be confusing.

daira (approved):
utACK

daira (approved):
ACK 1dff93a8ad3d958412eb6f821cb557988f128eef

PR #287, opened 2021-06-04
Remove rotations from Selectors

str4d:
Enabling selectors to be used in gates at non-zero rotations leads to
confusing gates, and inhibits our ability to create visualizations of
circuits. In most cases, a gate can be rearranged so that the selectors
have no rotation; in cases where cross-gate selector optimisations are
required, these can still be implemented using fixed columns.


daira (approved):
utACK

PR #289, opened 2021-06-04
CI: Build book with latest mdbook

str4d:
`mdbook-katex` is installed from crates.io, and if it doesn't use the same version of `mdbook` it can cause build issues.


str4d:
Fixes this CI failure: https://github.com/zcash/halo2/runs/2741958537

daira (approved):
utACK

PR #290, opened 2021-06-07
V1 layouter

str4d:
The V1 layouter is a dual-pass layouter, that has visibility into the entire `Circuit::synthesize` step.

This layouter improves on the single-pass strategy in two ways:

- Regions are layed out with a "first fit" strategy (using the algorithm described in #93) instead of a "last fit" strategy. This enables gaps in the circuit to be filled.

- Regions are sorted by their "advice area" (columns * rows), and are then layed out biggest-first. This takes advantage of the regularity of circuits, and that the advice columns have the most contention. It also leads to larger gaps between early layed-out regions that are more easily filled by subsequent smaller regions.

This layouter is named "V1" because its behaviour will be frozen when the Orchard circuit implementation is frozen. Future layouter optimizations could be implemented in a subsequent "V2" layouter.

Part of #93.


src/plonk/circuit.rs
```diff
@@ -34,6 +37,32 @@ impl<C: ColumnType> Column<C> {
     }
 }
 
+impl<C: ColumnType> Ord for Column<C> {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        // This ordering is consensus-critical! The layouters rely on deterministic column
```
ebfull, line 42: Wish we had a different phrase than "consensus critical" because this is halo2, there might not be any consensus involved in a downstream application. We need a word for "if the behavior of this changes for the prover, the verifier may reject the proof." Deterministic isn't enough.

ebfull (approved):

src/plonk/circuit.rs
str4d: "Circuit-critical"? We can figure this out later.

porcuquine: Maybe something like 'semantically binding'?

PR #291, opened 2021-06-08
dev: CircuitLayout builder

str4d:
This makes it easier to add configuration to the renderer.

Best reviewed with whitespace changes hidden.


daira (approved):
This is purely a dev tool and doesn't require detailed review. I skimmed it and it looks fine, utACK.

PR #293, opened 2021-06-11
Add support for batched inversion of cell assignments

str4d:
Value closures passed to `assign_advice` and `assign_fixed` can now return a `(numerator, denominator)` tuple, in order to defer inversions until after assignment.

### DIFF
--- a/src/poly.rs
+++ b/src/poly.rs
@@ -147,21 +147,26 @@ pub(crate) fn batch_invert_assigned<F: FieldExt>(
         .iter_mut()
         .flat_map(|f| {
             f.iter_mut()
-                // If the denominator is zero or one, we can skip it, reducing the
+                // If the denominator is trivial, we can skip it, reducing the
                 // size of the batch inversion.
-                .filter(|d| !(d.is_zero() || **d == F::one()))
+                .filter_map(|d| d.as_mut())
         })
         .batch_invert();
 
     assigned
         .iter()
         .zip(assigned_denominators.into_iter())
-        .map(|(poly, inv_denoms)| poly.invert(inv_denoms))
+        .map(|(poly, inv_denoms)| {
+            poly.invert(inv_denoms.into_iter().map(|d| d.unwrap_or_else(F::one)))
+        })
         .collect()
 }
 
 impl<F: Field> Polynomial<Assigned<F>, LagrangeCoeff> {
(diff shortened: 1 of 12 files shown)


str4d:
This PR conflicts with #290; that PR should be merged first.

src/poly.rs
```diff
@@ -128,6 +131,50 @@ impl<F, B> Polynomial<F, B> {
...
+
+    assigned
+        .iter()
+        .zip(assigned_denominators.into_iter())
+        .map(|(poly, inv_denoms)| poly.invert(inv_denoms))
+        .collect()
+}
+
+impl<F: Field> Polynomial<Assigned<F>, LagrangeCoeff> {
+    pub(crate) fn invert(&self, inv_denoms: Vec<F>) -> Polynomial<F, LagrangeCoeff> {
```
daira, line 164: This is actually doing a Hadamard product, which could be useful for other things, I guess.

daira (approved):
This is great, I love it! utACK; will rereview when it's rebased.

src/poly.rs
str4d: I've kept it internal for now, as I want to compute a single batch inversion with all advice or fixed columns, instead of per-column, which means the API is not nicely encapsulated.

str4d:
Rebased.

src/poly.rs
```diff
@@ -128,6 +131,50 @@ impl<F, B> Polynomial<F, B> {
...
+                .collect::<Vec<_>>()
+        })
+        .collect();
+
+    assigned_denominators
+        .iter_mut()
+        .flat_map(|f| {
+            f.iter_mut()
+                // If the denominator is zero or one, we can skip it, reducing the
+                // size of the batch inversion.
```
daira, line 151:
```suggestion
                // size (and for zeros, ensuring correctness) of the batch inversion.
```

daira (approved):
utACK with one minor suggestion to clarify a comment.

src/poly.rs
str4d: I believe this comment is unnecessary, because `.batch_invert()` leaves zeros alone.

daira: Yes, I had misremembered the contract of `batch_invert`.

ebfull (approved):

Issue #297, opened 2021-06-18
MockProver: Detect when a region activates a gate but does not assign every cell used by it


str4d:
#296 documents a problem with current layouters, where their shapes are under-measured when a region mistakenly forgets to assign a cell.

In #274 we added a `MockProver` check that all active gates have their cells assigned. This check happens globally across the circuit; we should instead have it check each region individually, by tracking region shapes and restricting the "cell assigned for gate" check to cells assigned within the region in which the gate's selector is enabled.

PR #298, opened 2021-06-18
MockProver: Check cell assignment per-region instead of globally

str4d:
Current layouters measure the shape of regions by looking at the cells
they assign. If a chip developer forgets to assign a cell in a region,
it is possible for active gates to "stick out" past the edge of a
measured region shape.

`MockProver` checks that all active gates have their cells assigned,
but this was previously checked globally. If a layouter happened to
position the buggy region such that the dangling gate overlapped an
assigned cell in in adjacent region, this check would pass.

In this commit, we extend the check to be per-region. We map enabled
selectors and assigned cells to a specific region, and then if a gate
is active within a region, we require that its cells be assigned within
that same region.

Closes zcash/halo2#297.

### DIFF
--- a/src/dev.rs
+++ b/src/dev.rs
@@ -32,6 +32,14 @@ fn cell_value<F: Field>(cell: Option<F>) -> F {
 pub enum VerifyFailure {
     /// A cell used in an active gate was not assigned to.
     Cell {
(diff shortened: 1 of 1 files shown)


ebfull:
ACK

PR #303, opened 2021-06-21
Migrate PLONK-gate-using circuits to use SingleChipLayouter

str4d:
These were all early examples that used the `Circuit` trait without any layouter, which is incompatible with upcoming changes.

The minimal change to the `PinnedVerificationKey` is because the lookup table is being allocated at the top of its columns instead of in-place.


ebfull (approved):

daira (commented):
Post-hoc ACK

PR #304, opened 2021-06-21
Reduce various trait requirements from `FieldExt` to `Field`

str4d:
We aren't using anything in `FieldExt` at these layers.


ebfull (approved):

daira (commented):
Post-hoc ACK

PR #306, opened 2021-06-22
Final transcript changes

ebfull:


daira (approved):
utACK

str4d (approved):

src/transcript.rs
```diff
@@ -108,13 +115,14 @@ impl<R: Read, C: CurveAffine> Transcript<C, Challenge255<C>>
     for Blake2bRead<R, C, Challenge255<C>>
 {
     fn squeeze_challenge(&mut self) -> Challenge255<C> {
+        self.state.update(&[BLAKE2B_PREFIX_CHALLENGE]);
         let hasher = self.state.clone();
         let result: [u8; 64] = hasher.finalize().as_bytes().try_into().unwrap();
-        self.state.update(&result[..]);
```
str4d, line 113: This change makes sense to me. Previously we needed to re-digest the challenge in order to advance the transcript (so squeezing multiple challenges in a row would return unique challenges). Now that the action of squeezing is marked in the transcript, I agree this is unnecessary.

PR #311, opened 2021-06-24
Use a rayon ThreadPool

daira:

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -37,10 +37,9 @@ harness = false
 
 [dependencies]
 backtrace = { version = "0.3", optional = true }
-crossbeam-utils = "0.8"
+rayon = "1.5.1"
 ff = "0.10"
 group = "0.10"
-num_cpus = "1.13"
 pasta_curves = "0.1"
 rand = "0.8"
 blake2b_simd = "0.5"
--- a/src/arithmetic.rs
+++ b/src/arithmetic.rs
@@ -167,13 +167,13 @@ pub fn small_multiexp<C: CurveAffine>(coeffs: &[C::Scalar], bases: &[C]) -> C::C
 pub fn best_multiexp<C: CurveAffine>(coeffs: &[C::Scalar], bases: &[C]) -> C::Curve {
     assert_eq!(coeffs.len(), bases.len());
 
-    let num_cpus = multicore::current_num_threads();
-    if coeffs.len() > num_cpus {
-        let chunk = coeffs.len() / num_cpus;
+    let num_threads = multicore::current_num_threads();
+    if coeffs.len() > num_threads {
+        let chunk = coeffs.len() / num_threads;
         let num_chunks = coeffs.chunks(chunk).len();
         let mut results = vec![C::Curve::identity(); num_chunks];
         multicore::scope(|scope| {
-            let chunk = coeffs.len() / num_cpus;
+            let chunk = coeffs.len() / num_threads;
 
             for ((coeffs, bases), acc) in coeffs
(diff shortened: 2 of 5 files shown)


daira:
This gives a significant improvement on the synthetic benchmarks but not on the real Orchard circuit, due to synchronization overhead in `rayon::ThreadPool`. We'll put off merging it until that has been investigated. It may be that the overhead of the rayon threadpool can be reduced, or we may need an alternative threadpool implementation.

daira:
I simplified the PR, but that won't have any effect on performance.

daira:
The behaviour on @str4d's machine seems to be specific to WSL2. There is a 44% improvement with the rayon threadpool when running on native Windows 10.

Cargo.toml
```diff
@@ -54,6 +56,7 @@ criterion = "0.3"
 gumdrop = "0.8"
 
 [features]
+no-threadpool = ["crossbeam-utils"]
```
str4d, line 59: Rust features should be additive (see [here](https://doc.rust-lang.org/cargo/reference/features.html#feature-unification) for details), so I think the correct way to implement this is to have two additive features, and either require that only one of them can be active at a time, or define which one "wins". Maybe also a third "best" feature flag that uses a compile-time or runtime strategy for choosing the best option.

daira: I considered that, but using the rayon threadpool is going to be the best strategy in the vast majority of cases, so this is imposing complexity on the crate users for little gain. I don't see any practical problem *in this situation* with features being additive: if any dependent crate enables no-threadpool, then you don't get the threadpool, which seems fine to me. (You can think of it as enabling the use of crossbeam if you prefer.)

If we could detect WSL at compile time, I'd do so. But what we really want to know is whether the target executable is going to be run under WSL.

daira: Prioritizing other things right now, but it looks like detecting WSL dynamically can be made to work.

daira: I removed the `no-threadpool` feature; using rayon is unconditional for now. I've filed #315 to address the regression on WSL2.

src/arithmetic.rs
```diff
@@ -172,7 +173,7 @@ pub fn best_multiexp<C: CurveAffine>(coeffs: &[C::Scalar], bases: &[C]) -> C::Cu
         let chunk = coeffs.len() / num_cpus;
         let num_chunks = coeffs.chunks(chunk).len();
         let mut results = vec![C::Curve::identity(); num_chunks];
-        thread::scope(|scope| {
+        multicore::scope(|scope| {
             let chunk = coeffs.len() / num_cpus;
```
str4d, line 177: This code chunks uniformly on `num_cpus` regardless of the size of the threadpool. The simple fix would be to make `NUM_CPUS` a `pub(crate)` static, and then use it here (and everywhere else in the crate we call `num_cpus::get()`).

src/lib.rs
```diff
@@ -25,6 +25,7 @@
 pub mod arithmetic;
 pub mod circuit;
 pub use pasta_curves as pasta;
+pub mod multicore;
```
str4d, line 28: This doesn't need to be public:
```suggestion
mod multicore;
```

str4d (dismissed):

src/arithmetic.rs
daira: This has been addressed. I made it `pub` rather than `pub(crate)`, since there's no reason to hide it, and external code might need to use it.

ebfull (approved):

str4d:
Post-hoc utACK.

PR #314, opened 2021-06-30
[book] Fix Sarkar square root explanation

therealyingtong:
As noted in the NCC audit:

Some formulas in the explanation of Sarkar's square root referenced the value g<sup>2<sup>-24</sup></sup>, where g is a primitive 2<sup>32</sup>-th root of unity modulo p (or q). This value is not defined, since g has order 2<sup>32</sup>, and 2<sup>24</sup> is not invertible modulo 2<sup>32</sup>. This expression has been corrected to g<sup>-2<sup>24</sup></sup>.

### DIFF
--- a/book/src/design/implementation/fields.md
+++ b/book/src/design/implementation/fields.md
@@ -29,7 +29,7 @@ $$
 
 $$
 invtab = \begin{bmatrix}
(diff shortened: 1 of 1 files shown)


daira (approved):
ACK

PR #316, opened 2021-07-01
Final ZK changes

ebfull:
Replaces #219.

### DIFF
--- a/benches/plonk.rs
+++ b/benches/plonk.rs
@@ -31,8 +31,6 @@ fn bench_with_k(name: &str, k: u32, c: &mut Criterion) {
         sb: Column<Fixed>,
         sc: Column<Fixed>,
         sm: Column<Fixed>,
-
-        perm: Permutation,
     }
 
     trait StandardCs<FF: FieldExt> {
@@ -164,10 +162,7 @@ fn bench_with_k(name: &str, k: u32, c: &mut Criterion) {
             left: Cell,
             right: Cell,
         ) -> Result<(), Error> {
-            layouter.assign_region(
-                || "copy",
-                |mut region| region.constrain_equal(&self.config.perm, left, right),
-            )
+            layouter.assign_region(|| "copy", |mut region| region.constrain_equal(left, right))
         }
     }
 
@@ -184,7 +179,9 @@ fn bench_with_k(name: &str, k: u32, c: &mut Criterion) {
             let b = meta.advice_column();
             let c = meta.advice_column();
 
-            let perm = meta.permutation(&[a.into(), b.into(), c.into()]);
+            meta.enable_equality(a.into());
+            meta.enable_equality(b.into());
+            meta.enable_equality(c.into());
 
             let sm = meta.fixed_column();
             let sa = meta.fixed_column();
@@ -212,7 +209,6 @@ fn bench_with_k(name: &str, k: u32, c: &mut Criterion) {
                 sb,
                 sc,
                 sm,
-                perm,
             }
         }
 
@@ -223,7 +219,7 @@ fn bench_with_k(name: &str, k: u32, c: &mut Criterion) {
         ) -> Result<(), Error> {
             let cs = StandardPlonk::new(config);
 
-            for _ in 0..(1 << (self.k - 1)) {
+            for _ in 0..((1 << (self.k - 1)) - 5) {
                 let mut a_squared = None;
                 let (a0, _, c0) = cs.raw_multiply(&mut layouter, || {
                     a_squared = self.a.map(|a| a.square());
--- a/benches/plonk.rs
+++ b/benches/plonk.rs
@@ -221,7 +221,7 @@ fn bench_with_k(name: &str, k: u32, c: &mut Criterion) {
         ) -> Result<(), Error> {
             let cs = StandardPlonk::new(config);
 
-            for _ in 0..((1 << (self.k - 1)) - 5) {
+            for _ in 0..((1 << (self.k - 1)) - 3) {
                 let mut a_squared = None;
                 let (a0, _, c0) = cs.raw_multiply(&mut layouter, || {
                     a_squared = self.a.map(|a| a.square());
--- a/benches/plonk.rs
+++ b/benches/plonk.rs
@@ -175,6 +175,8 @@ fn bench_with_k(name: &str, k: u32, c: &mut Criterion) {
         }
 
         fn configure(meta: &mut ConstraintSystem<F>) -> PlonkConfig {
+            meta.set_minimum_degree(5);
+
             let a = meta.advice_column();
             let b = meta.advice_column();
             let c = meta.advice_column();
--- a/examples/simple-example.rs
+++ b/examples/simple-example.rs
@@ -26,7 +26,12 @@ trait NumericInstructions<F: FieldExt>: Chip<F> {
     ) -> Result<Self::Num, Error>;
 
     /// Exposes a number as a public input to the circuit.
-    fn expose_public(&self, layouter: impl Layouter<F>, num: Self::Num) -> Result<(), Error>;
+    fn expose_public(
+        &self,
+        layouter: impl Layouter<F>,
+        num: Self::Num,
+        row: usize,
+    ) -> Result<(), Error>;
 }
 // ANCHOR_END: instructions
 
@@ -222,10 +227,15 @@ impl<F: FieldExt> NumericInstructions<F> for FieldChip<F> {
         Ok(out.unwrap())
     }
 
-    fn expose_public(&self, mut layouter: impl Layouter<F>, num: Self::Num) -> Result<(), Error> {
+    fn expose_public(
+        &self,
+        mut layouter: impl Layouter<F>,
+        num: Self::Num,
+        row: usize,
+    ) -> Result<(), Error> {
         let config = self.config();
 
-        layouter.constrain_instance(num.cell, config.instance, 0)
+        layouter.constrain_instance(num.cell, config.instance, row)
     }
 }
 // ANCHOR_END: instructions-impl
@@ -285,7 +295,7 @@ impl<F: FieldExt> Circuit<F> for MyCircuit<F> {
         let c = field_chip.mul(layouter.namespace(|| "ab * ab"), ab.clone(), ab)?;
 
         // Expose the result as a public input to the circuit.
-        field_chip.expose_public(layouter.namespace(|| "expose c"), c)
+        field_chip.expose_public(layouter.namespace(|| "expose c"), c, 0)
     }
 }
 // ANCHOR_END: circuit
--- a/examples/simple-example.rs
+++ b/examples/simple-example.rs
@@ -321,8 +321,7 @@ fn main() {
 
     // Arrange the public input. We expose the multiplication result in row 0
     // of the instance column, so we position it there in our public inputs.
-    let mut public_inputs = vec![Fp::zero(); 1 << k];
-    public_inputs[0] = c;
+    let mut public_inputs = vec![c];
 
     // Given the correct public input, our circuit will verify.
     let prover = MockProver::run(k, &circuit, vec![public_inputs.clone()]).unwrap();
--- a/examples/simple-example.rs
+++ b/examples/simple-example.rs
@@ -6,7 +6,7 @@ use halo2::{
     arithmetic::FieldExt,
     circuit::{Cell, Chip, Layouter, Region, SimpleFloorPlanner},
     dev::VerifyFailure,
-    plonk::{Advice, Circuit, Column, ConstraintSystem, Error, Instance, Permutation, Selector},
+    plonk::{Advice, Circuit, Column, ConstraintSystem, Error, Instance, Selector},
     poly::Rotation,
 };
 
@@ -50,11 +50,6 @@ struct FieldConfig {
     /// the circuit.
     advice: [Column<Advice>; 2],
 
-    // We need to create a permutation between our advice columns. This allows us to
-    // copy numbers within these columns from arbitrary rows, which we can use to load
-    // inputs into our instruction regions.
-    perm: Permutation,
-
     // We need a selector to enable the multiplication gate, so that we aren't placing
     // any constraints on cells where `NumericInstructions::mul` is not being used.
     // This is important when building larger circuits, where columns are used by
@@ -78,13 +73,9 @@ impl<F: FieldExt> FieldChip<F> {
         advice: [Column<Advice>; 2],
         instance: Column<Instance>,
     ) -> <Self as Chip<F>>::Config {
-        let perm = Permutation::new(
-            meta,
-            &advice
-                .iter()
-                .map(|column| (*column).into())
-                .collect::<Vec<_>>(),
-        );
+        for column in &advice {
+            meta.enable_equality((*column).into());
+        }
         let s_mul = meta.selector();
         let s_pub = meta.selector();
 
@@ -133,7 +124,6 @@ impl<F: FieldExt> FieldChip<F> {
 
         FieldConfig {
             advice,
-            perm,
             s_mul,
             s_pub,
         }
@@ -224,8 +214,8 @@ impl<F: FieldExt> NumericInstructions<F> for FieldChip<F> {
                     0,
                     || b.value.ok_or(Error::SynthesisError),
                 )?;
-                region.constrain_equal(&config.perm, a.cell, lhs)?;
-                region.constrain_equal(&config.perm, b.cell, rhs)?;
+                region.constrain_equal(a.cell, lhs)?;
+                region.constrain_equal(b.cell, rhs)?;
 
                 // Now we can assign the multiplication result into the output position.
                 let value = a.value.and_then(|a| b.value.map(|b| a * b));
@@ -262,7 +252,7 @@ impl<F: FieldExt> NumericInstructions<F> for FieldChip<F> {
                     0,
                     || num.value.ok_or(Error::SynthesisError),
                 )?;
-                region.constrain_equal(&config.perm, num.cell, out)?;
+                region.constrain_equal(num.cell, out)?;
 
                 // We don't assign to the instance column inside the circuit;
                 // the mapping of public inputs to cells is provided to the prover.
@@ -339,7 +329,7 @@ fn main() {
     // ANCHOR: test-circuit
     // The number of rows in our circuit cannot exceed 2^k. Since our example
     // circuit is very small, we can pick a very small value here.
-    let k = 3;
+    let k = 4;
 
     // Prepare the private and public inputs to the circuit!
     let a = Fp::from(2);
--- a/examples/simple-example.rs
+++ b/examples/simple-example.rs
@@ -5,7 +5,6 @@ use std::marker::PhantomData;
 use halo2::{
     arithmetic::FieldExt,
     circuit::{Cell, Chip, Layouter, Region, SimpleFloorPlanner},
-    dev::VerifyFailure,
     plonk::{Advice, Circuit, Column, ConstraintSystem, Error, Instance, Selector},
     poly::Rotation,
 };
@@ -50,14 +49,14 @@ struct FieldConfig {
     /// the circuit.
     advice: [Column<Advice>; 2],
 
+    /// This is the public input (instance) column.
+    instance: Column<Instance>,
+
     // We need a selector to enable the multiplication gate, so that we aren't placing
     // any constraints on cells where `NumericInstructions::mul` is not being used.
     // This is important when building larger circuits, where columns are used by
     // multiple sets of instructions.
     s_mul: Selector,
-
-    // The selector for the public-input gate, which uses one of the advice columns.
-    s_pub: Selector,
 }
 
 impl<F: FieldExt> FieldChip<F> {
@@ -73,11 +72,11 @@ impl<F: FieldExt> FieldChip<F> {
         advice: [Column<Advice>; 2],
         instance: Column<Instance>,
     ) -> <Self as Chip<F>>::Config {
+        meta.enable_equality(instance.into());
         for column in &advice {
             meta.enable_equality((*column).into());
         }
         let s_mul = meta.selector();
-        let s_pub = meta.selector();
 
         // Define our multiplication gate!
         meta.create_gate("mul", |meta| {
@@ -109,23 +108,10 @@ impl<F: FieldExt> FieldChip<F> {
             vec![s_mul * (lhs * rhs + out * -F::one())]
         });
 
-        // Define our public-input gate!
-        meta.create_gate("public input", |meta| {
-            // We choose somewhat-arbitrarily that we will use the second advice
-            // column for exposing numbers as public inputs.
-            let a = meta.query_advice(advice[1], Rotation::cur());
-            let p = meta.query_instance(instance, Rotation::cur());
-            let s = meta.query_selector(s_pub);
-
-            // We simply constrain the advice cell to be equal to the instance cell,
-            // when the selector is enabled.
-            vec![s * (p + a * -F::one())]
-        });
-
         FieldConfig {
             advice,
+            instance,
             s_mul,
-            s_pub,
         }
     }
 }
@@ -239,26 +225,7 @@ impl<F: FieldExt> NumericInstructions<F> for FieldChip<F> {
     fn expose_public(&self, mut layouter: impl Layouter<F>, num: Self::Num) -> Result<(), Error> {
         let config = self.config();
 
-        layouter.assign_region(
-            || "expose public",
-            |mut region: Region<'_, F>| {
-                // Enable the public-input gate.
-                config.s_pub.enable(&mut region, 0)?;
-
-                // Load the output into the correct advice column.
-                let out = region.assign_advice(
-                    || "public advice",
-                    config.advice[1],
-                    0,
-                    || num.value.ok_or(Error::SynthesisError),
-                )?;
-                region.constrain_equal(num.cell, out)?;
-
-                // We don't assign to the instance column inside the circuit;
-                // the mapping of public inputs to cells is provided to the prover.
-                Ok(())
-            },
-        )
+        layouter.constrain_instance(num.cell, config.instance, 0)
     }
 }
 // ANCHOR_END: instructions-impl
@@ -342,24 +309,18 @@ fn main() {
         b: Some(b),
     };
 
-    // Arrange the public input. We expose the multiplication result in row 6
+    // Arrange the public input. We expose the multiplication result in row 0
     // of the instance column, so we position it there in our public inputs.
     let mut public_inputs = vec![Fp::zero(); 1 << k];
-    public_inputs[6] = c;
+    public_inputs[0] = c;
 
     // Given the correct public input, our circuit will verify.
     let prover = MockProver::run(k, &circuit, vec![public_inputs.clone()]).unwrap();
     assert_eq!(prover.verify(), Ok(()));
 
     // If we try some other public input, the proof will fail!
-    public_inputs[6] += Fp::one();
+    public_inputs[0] += Fp::one();
     let prover = MockProver::run(k, &circuit, vec![public_inputs]).unwrap();
-    assert_eq!(
-        prover.verify(),
-        Err(vec![VerifyFailure::Constraint {
-            constraint: ((1, "public input").into(), 0, "").into(),
-            row: 6,
-        }])
-    );
+    assert!(prover.verify().is_err());
     // ANCHOR_END: test-circuit
 }
--- a/examples/two-chip.rs
+++ b/examples/two-chip.rs
@@ -568,8 +568,7 @@ fn main() {
 
     // Arrange the public input. We expose the multiplication result in row 0
     // of the instance column, so we position it there in our public inputs.
-    let mut public_inputs = vec![Fp::zero(); 1 << k];
-    public_inputs[0] = d;
+    let mut public_inputs = vec![d];
 
     // Given the correct public input, our circuit will verify.
     let prover = MockProver::run(k, &circuit, vec![public_inputs.clone()]).unwrap();
--- a/examples/two-chip.rs
+++ b/examples/two-chip.rs
@@ -5,7 +5,6 @@ use std::marker::PhantomData;
 use halo2::{
     arithmetic::FieldExt,
     circuit::{Cell, Chip, Layouter, Region, SimpleFloorPlanner},
-    dev::VerifyFailure,
     plonk::{Advice, Circuit, Column, ConstraintSystem, Error, Instance, Selector},
     poly::Rotation,
 };
@@ -43,6 +42,7 @@ trait FieldInstructions<F: FieldExt>: AddInstructions<F> + MulInstructions<F> {
         &self,
         layouter: impl Layouter<F>,
         num: <Self as FieldInstructions<F>>::Num,
+        row: usize,
     ) -> Result<(), Error>;
 }
 // ANCHOR_END: field-instructions
@@ -87,8 +87,8 @@ struct FieldConfig {
     /// the circuit.
     advice: [Column<Advice>; 2],
 
-    // The selector for the public-input gate, which uses one of the advice columns.
-    s_pub: Selector,
+    /// Public inputs
+    instance: Column<Instance>,
 
     add_config: AddConfig,
     mul_config: MulConfig,
@@ -173,10 +173,7 @@ impl<F: FieldExt> AddChip<F> {
             vec![s_add * (lhs + rhs + out * -F::one())]
         });
 
-        AddConfig {
-            advice,
-            s_add,
-        }
+        AddConfig { advice, s_add }
     }
 }
 // ANCHOR END: add-chip-impl
@@ -317,10 +314,7 @@ impl<F: FieldExt> MulChip<F> {
             vec![s_mul * (lhs * rhs + out * -F::one())]
         });
 
-        MulConfig {
-            advice,
-            s_mul,
-        }
+        MulConfig { advice, s_mul }
     }
 }
 // ANCHOR_END: mul-chip-impl
@@ -429,27 +423,14 @@ impl<F: FieldExt> FieldChip<F> {
         advice: [Column<Advice>; 2],
         instance: Column<Instance>,
     ) -> <Self as Chip<F>>::Config {
-        let s_pub = meta.selector();
-
-        // Define our public-input gate!
-        meta.create_gate("public input", |meta| {
-            // We choose somewhat-arbitrarily that we will use the second advice
-            // column for exposing numbers as public inputs.
-            let a = meta.query_advice(advice[1], Rotation::cur());
-            let p = meta.query_instance(instance, Rotation::cur());
-            let s = meta.query_selector(s_pub);
-
-            // We simply constrain the advice cell to be equal to the instance cell,
-            // when the selector is enabled.
-            vec![s * (p + a * -F::one())]
-        });
-
         let add_config = AddChip::configure(meta, advice);
         let mul_config = MulChip::configure(meta, advice);
 
+        meta.enable_equality(instance.into());
+
         FieldConfig {
             advice,
-            s_pub,
+            instance,
             add_config,
             mul_config,
         }
@@ -501,29 +482,11 @@ impl<F: FieldExt> FieldInstructions<F> for FieldChip<F> {
         &self,
         mut layouter: impl Layouter<F>,
         num: <Self as FieldInstructions<F>>::Num,
+        row: usize,
     ) -> Result<(), Error> {
         let config = self.config();
 
-        layouter.assign_region(
-            || "expose public",
-            |mut region: Region<'_, F>| {
-                // Enable the public-input gate.
-                config.s_pub.enable(&mut region, 0)?;
-
-                // Load the output into the correct advice column.
-                let out = region.assign_advice(
-                    || "public advice",
-                    config.advice[1],
-                    0,
-                    || num.value.ok_or(Error::SynthesisError),
-                )?;
-                region.constrain_equal(num.cell, out)?;
-
-                // We don't assign to the instance column inside the circuit;
-                // the mapping of public inputs to cells is provided to the prover.
-                Ok(())
-            },
-        )
+        layouter.constrain_instance(num.cell, config.instance, row)
     }
 }
 // ANCHOR_END: field-instructions-impl
@@ -576,7 +539,7 @@ impl<F: FieldExt> Circuit<F> for MyCircuit<F> {
         let d = field_chip.add_and_mul(&mut layouter, a, b, c)?;
 
         // Expose the result as a public input to the circuit.
-        field_chip.expose_public(layouter.namespace(|| "expose d"), d)
+        field_chip.expose_public(layouter.namespace(|| "expose d"), d, 0)
     }
 }
 // ANCHOR_END: circuit
@@ -603,24 +566,18 @@ fn main() {
         c: Some(c),
     };
 
-    // Arrange the public input. We expose the multiplication result in row 6
+    // Arrange the public input. We expose the multiplication result in row 0
     // of the instance column, so we position it there in our public inputs.
     let mut public_inputs = vec![Fp::zero(); 1 << k];
-    public_inputs[7] = d;
+    public_inputs[0] = d;
 
     // Given the correct public input, our circuit will verify.
     let prover = MockProver::run(k, &circuit, vec![public_inputs.clone()]).unwrap();
     assert_eq!(prover.verify(), Ok(()));
 
     // If we try some other public input, the proof will fail!
-    public_inputs[7] += Fp::one();
+    public_inputs[0] += Fp::one();
     let prover = MockProver::run(k, &circuit, vec![public_inputs]).unwrap();
-    assert_eq!(
-        prover.verify(),
-        Err(vec![VerifyFailure::Constraint {
-            constraint: ((0, "public input").into(), 0, "").into(),
-            row: 7,
-        }])
-    );
+    assert!(prover.verify().is_err());
     // ANCHOR_END: test-circuit
 }
--- a/examples/two-chip.rs
+++ b/examples/two-chip.rs
@@ -6,7 +6,7 @@ use halo2::{
     arithmetic::FieldExt,
     circuit::{Cell, Chip, Layouter, Region, SimpleFloorPlanner},
     dev::VerifyFailure,
-    plonk::{Advice, Circuit, Column, ConstraintSystem, Error, Instance, Permutation, Selector},
+    plonk::{Advice, Circuit, Column, ConstraintSystem, Error, Instance, Selector},
     poly::Rotation,
 };
 
@@ -87,11 +87,6 @@ struct FieldConfig {
     /// the circuit.
     advice: [Column<Advice>; 2],
 
-    // We need to create a permutation between our advice columns. This allows us to
-    // copy numbers within these columns from arbitrary rows, which we can use to load
-    // inputs into our instruction regions.
-    perm: Permutation,
-
     // The selector for the public-input gate, which uses one of the advice columns.
     s_pub: Selector,
 
@@ -104,7 +99,6 @@ struct FieldConfig {
 #[derive(Clone, Debug)]
 struct AddConfig {
     advice: [Column<Advice>; 2],
-    perm: Permutation,
     s_add: Selector,
 }
 // ANCHOR_END: add-config
@@ -113,7 +107,6 @@ struct AddConfig {
 #[derive(Clone, Debug)]
 struct MulConfig {
     advice: [Column<Advice>; 2],
-    perm: Permutation,
     s_mul: Selector,
 }
 // ANCHOR END: mul-config
@@ -167,7 +160,6 @@ impl<F: FieldExt> AddChip<F> {
     fn configure(
         meta: &mut ConstraintSystem<F>,
         advice: [Column<Advice>; 2],
-        perm: Permutation,
     ) -> <Self as Chip<F>>::Config {
         let s_add = meta.selector();
 
@@ -183,7 +175,6 @@ impl<F: FieldExt> AddChip<F> {
 
         AddConfig {
             advice,
-            perm,
             s_add,
         }
     }
@@ -242,8 +233,8 @@ impl<F: FieldExt> AddInstructions<F> for AddChip<F> {
                     0,
                     || b.value.ok_or(Error::SynthesisError),
                 )?;
-                region.constrain_equal(&config.perm, a.cell, lhs)?;
-                region.constrain_equal(&config.perm, b.cell, rhs)?;
+                region.constrain_equal(a.cell, lhs)?;
+                region.constrain_equal(b.cell, rhs)?;
 
                 // Now we can assign the multiplication result into the output position.
                 let value = a.value.and_then(|a| b.value.map(|b| a + b));
@@ -293,8 +284,10 @@ impl<F: FieldExt> MulChip<F> {
     fn configure(
         meta: &mut ConstraintSystem<F>,
         advice: [Column<Advice>; 2],
-        perm: Permutation,
     ) -> <Self as Chip<F>>::Config {
+        for column in &advice {
+            meta.enable_equality((*column).into());
+        }
         let s_mul = meta.selector();
 
         // Define our multiplication gate!
@@ -326,7 +319,6 @@ impl<F: FieldExt> MulChip<F> {
 
         MulConfig {
             advice,
-            perm,
             s_mul,
         }
     }
@@ -384,8 +376,8 @@ impl<F: FieldExt> MulInstructions<F> for MulChip<F> {
                     0,
                     || b.value.ok_or(Error::SynthesisError),
                 )?;
-                region.constrain_equal(&config.perm, a.cell, lhs)?;
-                region.constrain_equal(&config.perm, b.cell, rhs)?;
+                region.constrain_equal(a.cell, lhs)?;
+                region.constrain_equal(b.cell, rhs)?;
 
                 // Now we can assign the multiplication result into the output position.
                 let value = a.value.and_then(|a| b.value.map(|b| a * b));
@@ -437,13 +429,6 @@ impl<F: FieldExt> FieldChip<F> {
         advice: [Column<Advice>; 2],
         instance: Column<Instance>,
     ) -> <Self as Chip<F>>::Config {
-        let perm = Permutation::new(
-            meta,
-            &advice
-                .iter()
-                .map(|column| (*column).into())
-                .collect::<Vec<_>>(),
-        );
         let s_pub = meta.selector();
 
         // Define our public-input gate!
@@ -459,12 +444,11 @@ impl<F: FieldExt> FieldChip<F> {
             vec![s * (p + a * -F::one())]
         });
 
-        let add_config = AddChip::configure(meta, advice, perm.clone());
-        let mul_config = MulChip::configure(meta, advice, perm.clone());
+        let add_config = AddChip::configure(meta, advice);
+        let mul_config = MulChip::configure(meta, advice);
 
         FieldConfig {
             advice,
-            perm,
             s_pub,
             add_config,
             mul_config,
@@ -533,7 +517,7 @@ impl<F: FieldExt> FieldInstructions<F> for FieldChip<F> {
                     0,
                     || num.value.ok_or(Error::SynthesisError),
                 )?;
-                region.constrain_equal(&config.perm, num.cell, out)?;
+                region.constrain_equal(num.cell, out)?;
 
                 // We don't assign to the instance column inside the circuit;
                 // the mapping of public inputs to cells is provided to the prover.
@@ -604,7 +588,7 @@ fn main() {
     // ANCHOR: test-circuit
     // The number of rows in our circuit cannot exceed 2^k. Since our example
     // circuit is very small, we can pick a very small value here.
-    let k = 3;
+    let k = 4;
 
     // Prepare the private and public inputs to the circuit!
     let a = Fp::rand();
--- a/src/circuit.rs
+++ b/src/circuit.rs
@@ -155,8 +155,8 @@ impl<'r, F: Field> Region<'r, F> {
     /// Assign the value of the instance column's cell at absolute location
     /// `row` to the column `advice` at `offset` within this region, and return
     /// a `Cell`.
-    pub fn assign_advice_from_instance<'v, A, AR>(
-        &'v mut self,
+    pub fn assign_advice_from_instance<A, AR>(
+        &mut self,
         annotation: A,
         instance: Column<Instance>,
         row: usize,
--- a/src/circuit.rs
+++ b/src/circuit.rs
@@ -6,7 +6,7 @@ use ff::Field;
 
 use crate::{
     arithmetic::FieldExt,
-    plonk::{Advice, Any, Assigned, Column, Error, Fixed, Selector},
+    plonk::{Advice, Any, Assigned, Column, Error, Fixed, Instance, Selector},
 };
 
 pub mod floor_planner;
@@ -152,6 +152,30 @@ impl<'r, F: Field> Region<'r, F> {
             })
     }
 
+    /// Assign the value of the instance column's cell at absolute location
+    /// `row` to the column `advice` at `offset` within this region, and return
+    /// a `Cell`.
+    pub fn assign_advice_from_instance<'v, A, AR>(
+        &'v mut self,
+        annotation: A,
+        instance: Column<Instance>,
+        row: usize,
+        advice: Column<Advice>,
+        offset: usize,
+    ) -> Result<(Cell, Option<F>), Error>
+    where
+        A: Fn() -> AR,
+        AR: Into<String>,
+    {
+        self.region.assign_advice_from_instance(
+            &|| annotation().into(),
+            instance,
+            row,
+            advice,
+            offset,
+        )
+    }
+
     /// Assign a fixed value.
     ///
     /// Even though `to` has `FnMut` bounds, it is guaranteed to be called at most once.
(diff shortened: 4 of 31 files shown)


src/plonk/circuit.rs
```diff
@@ -792,6 +792,7 @@ pub struct ConstraintSystem<F: Field> {
     pub(crate) num_instance_columns: usize,
     pub(crate) gates: Vec<Gate<F>>,
     pub(crate) advice_queries: Vec<(Column<Advice>, Rotation)>,
+    num_advice_queries: Vec<usize>,
```
daira, line 795: Document the invariant relating `num_advice_queries` and `advice_queries`.

src/plonk/circuit.rs
```diff
@@ -1104,6 +1108,32 @@ impl<F: Field> ConstraintSystem<F> {
...
     }
+
+    /// Compute the number of blinding factors necessary to perfectly blind
+    /// each of the prover's witness polynomials.
+    pub fn blinding_factors(&self) -> usize {
+        // All of the prover's advice columns are evaluated at most
+        let factors = *self.num_advice_queries.iter().max().unwrap_or(&1);
+        // distinct points during gate checks. In the permutation and lookup
+        // argument the witness polynomials are evaluated at most 2 times:
+
```
daira, line 1119:
```suggestion
        // The maximum number of distinct points on which any of the prover's
        // advice columns are evaluated during gate checks is at most:
        let factors = *self.num_advice_queries.iter().max().unwrap_or(&1);

        // In the permutation and lookup arguments the witness polynomials
        // are evaluated at most 2 times:
```

src/plonk/permutation.rs
```diff
@@ -35,8 +35,8 @@ impl Argument {
         // l_0(X) * (1 - z(X)) = 0
         //
         // degree columns + 1
-        // z(X) \prod (p(X) + \beta s_i(X) + \gamma)
-        // - z(omega^{-1} X) \prod (p(X) + \delta^i \beta X + \gamma)
+        // z(omega X) \prod (p(X) + \beta s_i(X) + \gamma)
+        // - z(X) \prod (p(X) + \delta^i \beta X + \gamma)
```
daira, line 39:
```suggestion
        // - z(X) \prod (p(X) + \beta \delta^i X + \gamma)
```
This is how it's written in the halo2 book.

src/plonk/permutation/prover.rs
```diff
@@ -171,9 +171,9 @@ impl<C: CurveAffine> Committed<C> {
             .chain(Some(
                 Polynomial::one_minus(self.permutation_product_coset.clone()) * &pk.l0,
             ))
-            // z(X) \prod (p(X) + \beta s_i(X) + \gamma) - z(omega^{-1} X) \prod (p(X) + \delta^i \beta X + \gamma)
+            // z(omega X) \prod (p(X) + \beta s_i(X) + \gamma) - z(X) \prod (p(X) + \delta^i \beta X + \gamma)
```
daira, line 174:
```suggestion
            // z(omega X) \prod (p(X) + \beta s_i(X) + \gamma) - z(X) \prod (p(X) + \beta \delta^i X + \gamma)
```

src/plonk/permutation/verifier.rs
```diff
@@ -88,10 +88,10 @@ impl<C: CurveAffine> Evaluated<C> {
             .chain(Some(
                 l_0 * &(C::Scalar::one() - &self.permutation_product_eval),
             ))
-            // z(X) \prod (p(X) + \beta s_i(X) + \gamma)
-            // - z(omega^{-1} X) \prod (p(X) + \delta^i \beta X + \gamma)
+            // z(omega X) \prod (p(X) + \beta s_i(X) + \gamma)
+            // - z(X) \prod (p(X) + \delta^i \beta X + \gamma)
```
daira, line 92:
```suggestion
            // - z(X) \prod (p(X) + \beta \delta^i X + \gamma)
```

src/plonk/permutation.rs
ebfull: Thanks. Not sure that I like this TBH, because it's important not to lose sight of powers of `\delta` here which are really doing something; the \beta term is a challenge that basically behaves like a constant at this point in the protocol.

src/circuit.rs
```diff
@@ -152,6 +152,30 @@ impl<'r, F: Field> Region<'r, F> {
...
+    /// `row` to the column `advice` at `offset` within this region, and return
+    /// a `Cell`.
+    pub fn assign_advice_from_instance<'v, A, AR>(
+        &'v mut self,
+        annotation: A,
+        instance: Column<Instance>,
+        row: usize,
+        advice: Column<Advice>,
+        offset: usize,
+    ) -> Result<(Cell, Option<F>), Error>
```
daira, line 165: Fix the clippy needless-lifetimes lint.

src/circuit.rs
```diff
@@ -177,13 +201,8 @@ impl<'r, F: Field> Region<'r, F> {
     /// Constraint two cells to have the same value.
```
daira, line 201:
```suggestion
    /// Constrain two cells to have the same value.
```

src/circuit.rs
```diff
@@ -177,13 +201,8 @@ impl<'r, F: Field> Region<'r, F> {
...
     ///
     /// Returns an error if either of the cells is not within the given permutation.
```
daira, line 203:
```suggestion
    /// Constrain two cells to have the same value.
```

src/dev.rs
```diff
@@ -64,12 +64,8 @@ pub enum VerifyFailure {
     },
     /// A permutation did not preserve the original value of a cell.
     Permutation {
```
daira, line 66: Maybe rename this to `Equality`?

src/dev.rs
```diff
@@ -250,7 +242,9 @@ pub struct MockProver<F: Group + Field> {
     // The instance cells in the circuit, arranged as [column][row].
     instance: Vec<Vec<F>>,
 
-    permutations: Vec<permutation::keygen::Assembly>,
+    permutation: permutation::keygen::Assembly,
+
+    upper_bound_cell_index: usize,
```
daira, line 247: Document what this is.

src/dev.rs
```diff
@@ -407,24 +406,27 @@ impl<F: FieldExt> MockProver<F> {
...
-        let fixed = vec![vec![None; n as usize]; cs.num_fixed_columns];
-        let advice = vec![vec![None; n as usize]; cs.num_advice_columns];
-        let permutations = cs
-            .permutations
-            .iter()
-            .map(|p| permutation::keygen::Assembly::new(n as usize, p))
-            .collect();
+        if n < cs.blinding_factors() + 3 {
+            panic!("not enough rows available to synthesize");
+        }
```
daira, line 413: Return an Error rather than panicking.

src/plonk/circuit.rs
```diff
@@ -1104,6 +1069,32 @@ impl<F: Field> ConstraintSystem<F> {
 
         degree
     }
+
+    /// Compute the number of blinding factors necessary to perfectly blind
+    /// each of the prover's witness polynomials.
+    pub fn blinding_factors(&self) -> usize {
+        // All of the prover's advice columns are evaluated at most
```
daira, line 1076:
```suggestion
        // All of the prover's advice columns are evaluated at no more than
```

src/plonk/circuit.rs
```diff
@@ -1104,6 +1069,32 @@ impl<F: Field> ConstraintSystem<F> {
...
+        let factors = *self.num_advice_queries.iter().max().unwrap_or(&1);
+        // distinct points during gate checks. In the permutation and lookup
+        // argument the witness polynomials are evaluated at most 3 times:
+
```
daira, line 1080:
```suggestion
        // distinct points during gate checks.

        // In the permutation and lookup argument the witness polynomials are
        // evaluated at most 3 times:
```

src/plonk/circuit.rs
```diff
@@ -1104,6 +1069,32 @@ impl<F: Field> ConstraintSystem<F> {
...
+        let factors = std::cmp::max(3, factors);
+
+        // Each polynomial is evaluated at most an additional time during
+        // multiopen
```
daira, line 1084:
```suggestion
        // multiopen:
```

src/plonk/circuit.rs
```diff
@@ -1104,6 +1069,32 @@ impl<F: Field> ConstraintSystem<F> {
...
+        let factors = factors + 1;
+
+        // h(x) is derived by the other evaluations so it does not reveal
+        // anything; in fact it does not even appear in the proof.
+
+        // h(x_3) is also not revealed; the verifier only learns a single
+        // evaluation of a polynomial in x_1 which has h(x_3) and another random
+        // polynomial evaluated at x_3 as coefficients
```
daira, line 1092:
```suggestion
        // polynomial evaluated at x_3 as coefficients.
```

src/plonk/circuit.rs
```diff
@@ -1104,6 +1069,32 @@ impl<F: Field> ConstraintSystem<F> {
...
+
+        // Add an additional blinding factor as a slight defense against
+        // off-by-one errors.
+        factors + 1
```
daira, line 1096: :+1:

src/plonk/keygen.rs
```diff
@@ -218,21 +222,27 @@ where
...
 
+    let cs = cs;
+    // There needs to be enough room for at least one row.
+    assert!(
+        cs.blinding_factors() // m blinding factors
+        + 1 // for l_{-(m + 1)}
+        + 1 // for l_0
+        + 1 // for at least one row
+        <= (params.n as usize)
+    );
```
daira, line 233: Duplicate code. Consider adding a `cs.minimum_rows()` method. (This could also be used in MockProver where `cs.blinding_factors() + 3` [is hard-coded](https://github.com/zcash/halo2/pull/316/files?diff=unified&w=1#diff-da17e7db207e483590bb92572e40b583f57b8737739a902946b46bc0d59b0465R411).)

src/plonk/keygen.rs
```diff
@@ -267,12 +270,34 @@ where
     let l0 = vk.domain.lagrange_to_coeff(l0);
     let l0 = vk.domain.coeff_to_extended(l0, Rotation::cur());
 
+    // Compute l_cover(X) which evaluates to 1 for each blinding factor row
+    // and 0 otherwise over the domain.
+    let mut l_cover = vk.domain.empty_lagrange();
```
daira, line 275: Consider renaming to `l_blind`; `l_cover` doesn't seem a descriptive name to me.

src/plonk/keygen.rs
```diff
@@ -267,12 +270,34 @@ where
...
+    let l_cover = vk.domain.coeff_to_extended(l_cover, Rotation::cur());
+
+    // Compute l_last(X) which evaluates to 1 on the first inactive row (just
+    // before the blinding factors) and 0 otherwise over the domain
+    let mut l_last = vk.domain.empty_lagrange();
+    *(l_last[..]
+        .iter_mut()
+        .rev()
+        .nth(cs.blinding_factors())
+        .unwrap()) = C::Scalar::one();
```
daira, line 289: Can this be expressed more simply as something like `l_last[params.n - cs.blinding_factors() - 1] = C::Scalar::one();`?

src/plonk/lookup.rs
```diff
@@ -25,18 +25,29 @@ impl<F: Field> Argument<F> {
...
 
+        // The first value in the permutation poly should be one.
         // degree 2:
-        // l_0(X) * (1 - z'(X)) = 0
+        // l_0(X) * (1 - z(X)) = 0
         //
-        // degree (1 + input_degree + table_degree):
-        // z'(X) (a'(X) + \beta) (s'(X) + \gamma)
-        // - z'(\omega^{-1} X) (\theta^{m-1} a_0(X) + ... + a_{m-1}(X) + \beta) (\theta^{m-1} s_0(X) + ... + s_{m-1}(X) + \gamma)
+        // The "last" value in the permutation poly should be a boolean, for completeness.
```
daira, line 32:
```suggestion
        // The "last" value in the permutation poly should be a boolean, for completeness and soundness.
```

src/plonk/permutation.rs
daira: OK, we can change the book instead.

src/dev.rs
ebfull: @str4d and I discussed that in a pairing, apparently there are many changes we'd like to make to the mock prover and I don't want this PR to get bogged down in those improvements.

src/dev.rs
ebfull: Adding documentation for it. Note that it's documented elsewhere in `plonk::keygen`, this is just the developer tooling, but it can't hurt to have additional docs.

daira: I'm not sure why it's called `something_cell_index`, though. Also "upper bound" is ambiguous as to whether it's the last usable row or the last plus one (whereas `usable_rows` is unambiguous, at least if you think about it for a bit).

src/plonk/keygen.rs
```diff
@@ -168,29 +170,25 @@ where
 {
     let (domain, cs, config) = create_domain::<C, ConcreteCircuit>(params);
 
+    if (params.n as usize) < cs.minimum_rows() {
```
daira, line 173: There seem to be a lot of `.n as usize` (at least 32 instances). Should it instead be declared as `usize`? Not for this PR.

src/plonk/lookup.rs
```diff
@@ -25,18 +25,30 @@ impl<F: Field> Argument<F> {
     pub(crate) fn required_degree(&self) -> usize {
         assert_eq!(self.input_expressions.len(), self.table_expressions.len());
 
+        // The first value in the permutation poly should be one.
         // degree 2:
-        // l_0(X) * (1 - z'(X)) = 0
+        // l_0(X) * (1 - z(X)) = 0
```
daira, line 30:
```suggestion
        // l_0(X) * (z(X) - 1) = 0
```
following https://github.com/zcash/orchard/issues/143.

src/plonk/keygen.rs
ebfull: Yeah, it's a general issue. I want the code to work on both 32-bit and 64-bit platforms, so I chose to use `u32` for some of the lengths to essentially force me to address the fact that some larger lengths are impossible by adding errors and explicit checks. Once the checks were written, though, it started to just get annoying to cast as `usize` all over the place. We can fix this later.

src/plonk/lookup.rs
ebfull: Oh, this will require changing the constraint system as it's not just an innocent notational quirk. I prefer `1 - X` tbh, it's more common in our other SNARK code, and we've already implemented everything like this using `Polynomial::one_minus`.

src/plonk/lookup/prover.rs
```diff
@@ -327,35 +330,44 @@ impl<C: CurveAffine> Permuted<C> {
...
+        assert_eq!(z.len(), params.n as usize);
         let z = pk.vk.domain.lagrange_from_vec(z);
 
         #[cfg(feature = "sanity-checks")]
         // This test works only with intermediate representations in this method.
         // It can be used for debugging purposes.
         {
             // While in Lagrange basis, check that product is correctly constructed
-            let n = params.n as usize;
+            let n = (params.n as usize) - (blinding_factors + 1);
```
daira, line 352: We call this u in https://github.com/zcash/halo2/blob/750ab469d0d71090ce56834fc78e46db25f5250b/book/src/design/proving-system/lookup.md#zero-knowledge-adjustment .

src/plonk/lookup/prover.rs
```diff
@@ -327,35 +330,44 @@ impl<C: CurveAffine> Permuted<C> {
...
 
-            // z'(X) (a'(X) + \beta) (s'(X) + \gamma)
-            // - z'(\omega^{-1} X) (\theta^{m-1} a_0(X) + ... + a_{m-1}(X) + \beta) (\theta^{m-1} s_0(X) + ... + s_{m-1}(X) + \gamma)
+            // l_0(X) * (1 - z(X)) = 0
```
daira, line 354:
```suggestion
            // l_0(X) * (z(X) - 1) = 0
```

src/plonk/lookup/prover.rs
```diff
@@ -327,35 +330,44 @@ impl<C: CurveAffine> Permuted<C> {
...
+            assert_eq!(z[0], C::Scalar::one());
+
+            // z(\omega X) (a'(X) + \beta) (s'(X) + \gamma)
+            // - z(X) (\theta^{m-1} a_0(X) + ... + a_{m-1}(X) + \beta) (\theta^{m-1} s_0(X) + ... + s_{m-1}(X) + \gamma)
             for i in 0..n {
-                let prev_idx = (n + i - 1) % n;
+                let next_idx = (i + 1) % n;
```
daira, line 360:
```suggestion
                let next_idx = i + 1;
```

src/plonk/lookup/prover.rs
```diff
@@ -327,35 +330,44 @@ impl<C: CurveAffine> Permuted<C> {
...
 
-                let mut left = z[i];
+                let mut left = z[next_idx];
```
daira, line 362:
```suggestion
                let mut left = z[i+1];
```

src/plonk/lookup/prover.rs
```diff
@@ -327,35 +330,44 @@ impl<C: CurveAffine> Permuted<C> {
...
-            // z'(X) (a'(X) + \beta) (s'(X) + \gamma)
-            // - z'(\omega^{-1} X) (\theta^{m-1} a_0(X) + ... + a_{m-1}(X) + \beta) (\theta^{m-1} s_0(X) + ... + s_{m-1}(X) + \gamma)
+            // l_0(X) * (1 - z(X)) = 0
+            assert_eq!(z[0], C::Scalar::one());
+
+            // z(\omega X) (a'(X) + \beta) (s'(X) + \gamma)
+            // - z(X) (\theta^{m-1} a_0(X) + ... + a_{m-1}(X) + \beta) (\theta^{m-1} s_0(X) + ... + s_{m-1}(X) + \gamma)
             for i in 0..n {
-                let prev_idx = (n + i - 1) % n;
+                let next_idx = (i + 1) % n;
```
daira, line 361:
```suggestion
```

src/plonk/lookup/prover.rs
```diff
@@ -414,16 +434,25 @@ impl<'a, C: CurveAffine> Committed<C> {
...
-            // l_0(X) * (1 - z'(X)) = 0
+            // l_0(X) * (1 - z(X)) = 0
             .chain(Some(
                 Polynomial::one_minus(self.product_coset.clone()) * &pk.l0,
             ))
-            // z'(X) (a'(X) + \beta) (s'(X) + \gamma)
-            // - z'(\omega^{-1} X) (\theta^{m-1} a_0(X) + ... + a_{m-1}(X) + \beta) (\theta^{m-1} s_0(X) + ... + s_{m-1}(X) + \gamma)
+            // l_last(X) * (z(X)^2 - z(X)) = 0
+            .chain(Some(
+                (self.product_coset.clone() * &self.product_coset - &self.product_coset)
```
daira, line 446:
```suggestion
                (self.product_coset.square() - &self.product_coset)
```

src/plonk/lookup/prover.rs
```diff
@@ -414,16 +434,25 @@ impl<'a, C: CurveAffine> Committed<C> {
...
+            // (1 - (l_last + l_blind)) * (
+            //   z(\omega X) (a'(X) + \beta) (s'(X) + \gamma)
+            //   - z(X) (\theta^{m-1} a_0(X) + ... + a_{m-1}(X) + \beta) (\theta^{m-1} s_0(X) + ... + s_{m-1}(X) + \gamma)
+            // ) = 0
             .chain({
-                // z'(X) (a'(X) + \beta) (s'(X) + \gamma)
-                let mut left = self.product_coset.clone();
+                // z(\omega X) (a'(X) + \beta) (s'(X) + \gamma)
+                let mut left = self.product_next_coset.clone();
                 parallelize(&mut left, |left, start| {
```
daira, line 456: This is only constructing the expression, right, not doing any actual computation? If that's correct then there is probably no advantage to parallelising it.

src/plonk/lookup/prover.rs
```diff
@@ -472,10 +501,11 @@ impl<'a, C: CurveAffine> Committed<C> {
             // Check that each value in the permuted lookup input expression is either
             // equal to the value above it, or the value at the same index in the
             // permuted table expression.
-            // (a′(X)−s′(X))⋅(a′(X)−a′(\omega{-1} X)) = 0
+            // (1 - (l_last + l_blind)) * (a′(X)−s′(X))⋅(a′(X)−a′(\omega{-1} X)) = 0
```
daira, line 504:
```suggestion
            // (1 - (l_last + l_blind)) * (a′(X) − s′(X))⋅(a′(X) − a′(\omega^{-1} X)) = 0
```

src/plonk/lookup/prover.rs
```diff
@@ -576,24 +608,29 @@ type ExpressionPair<F> = (Polynomial<F, LagrangeCoeff>, Polynomial<F, LagrangeCo
...
+    pk: &ProvingKey<C>,
+    params: &Params<C>,
     domain: &EvaluationDomain<C::Scalar>,
     input_expression: &Polynomial<C::Scalar, LagrangeCoeff>,
     table_expression: &Polynomial<C::Scalar, LagrangeCoeff>,
 ) -> Result<ExpressionPair<C::Scalar>, Error> {
-    let mut permuted_input_expression = input_expression.clone();
+    let blinding_factors = pk.vk.cs.blinding_factors();
+    let mut permuted_input_expression: Vec<C::Scalar> = input_expression.to_vec();
+    permuted_input_expression.truncate(params.n as usize - (blinding_factors + 1));
```
daira, line 619: We use `params.n as usize - (blinding_factors + 1)` three times; consider defining a local variable for it.

src/plonk/lookup/prover.rs
```diff
@@ -627,18 +664,28 @@ fn permute_expression_pair<C: CurveAffine>(
...
-            {
-                *permuted_table_value += permuted_table_coeff;
+    permuted_input_expression.extend((0..(blinding_factors + 1)).map(|_| C::Scalar::rand()));
+    permuted_table_coeffs.extend((0..(blinding_factors + 1)).map(|_| C::Scalar::rand()));
+    assert_eq!(permuted_input_expression.len(), params.n as usize);
+    assert_eq!(permuted_table_coeffs.len(), params.n as usize);
+
+    #[cfg(feature = "sanity-checks")]
+    {
+        let mut last = C::Scalar::rand();
```
daira, line 674: This is confusing.
```suggestion
        let mut last = None;
```

src/plonk/lookup/prover.rs
```diff
@@ -627,18 +664,28 @@ fn permute_expression_pair<C: CurveAffine>(
...
+        for (a, b) in permuted_input_expression
+            .iter()
+            .zip(permuted_table_coeffs.iter())
+            .take(params.n as usize - (blinding_factors + 1))
+        {
+            if *a != *b {
+                assert_eq!(*a, last);
             }
```
daira, line 682:
```suggestion
            if let Some(last) = last {
                if *a != *b {
                    assert_eq!(*a, last);
                }
            }
```

src/dev.rs
ebfull: Okay, I've [changed it](https://github.com/zcash/halo2/pull/316/commits/bd29ebdb057768234104846453b4545cb3e41a60) to `usable_rows` and made it a `RangeTo<usize>` (per @str4d's suggestion).

src/plonk/lookup/prover.rs
ebfull: This uncovered a bug in the sanity check actually, thanks.

src/plonk/lookup/prover.rs
```diff
@@ -627,18 +664,28 @@ fn permute_expression_pair<C: CurveAffine>(
...
+    #[cfg(feature = "sanity-checks")]
+    {
+        let mut last = C::Scalar::rand();
+        for (a, b) in permuted_input_expression
+            .iter()
+            .zip(permuted_table_coeffs.iter())
+            .take(params.n as usize - (blinding_factors + 1))
+        {
+            if *a != *b {
+                assert_eq!(*a, last);
```
daira, line 681:
```suggestion
                assert_eq!(*a, last.unwrap());
```

src/plonk/lookup/prover.rs
```diff
@@ -627,18 +664,28 @@ fn permute_expression_pair<C: CurveAffine>(
...
             }
-        },
-    );
+            last = *a;
```
daira, line 683:
```suggestion
            last = Some(*a);
```

src/dev.rs
```diff
@@ -250,7 +243,10 @@ pub struct MockProver<F: Group + Field> {
     // The instance cells in the circuit, arranged as [column][row].
     instance: Vec<Vec<F>>,
 
-    permutations: Vec<permutation::keygen::Assembly>,
+    permutation: permutation::keygen::Assembly,
+
+    // A range of available rows for assignment and copies.
+    usable_rows: RangeTo<usize>,
```
daira, line 249: Much better! :-)

src/plonk/lookup/prover.rs
ebfull: There isn't a square method for Polynomial, let's address this another time.

src/plonk/lookup.rs
daira: Yeah, I deleted these comments (that were asking for this to be changed to `z(X) - 1`).

src/plonk/lookup/prover.rs
ebfull: Sorry :) wrote it while debugging and thought I'd delete this anyway, but it's not really a bad sanity check to leave in.

ebfull (commented):
github's review UX is horribly broken for me

src/plonk/permutation.rs
```diff
@@ -20,24 +18,49 @@ pub(crate) struct Argument {
...
+        //
+        // degree 3:
+        // l_last(X) * (z'(X)^2 - z'(X)) = 0
+        //
+        // which will allow the last value to be zero to
+        // ensure the argument is perfectly complete.
+
+        // There are constraints of degree 3 regardless of the
+        // number of columns involved.
+        3
```
daira, line 57: I'm confused, I thought the idea was to pack as many columns per column-set as would fit in the degree bound.

str4d: It is. However, if the number of columns is few enough that they contribute lower than degree 3, this sets the floor due to the `l_last` constraint.

What is confusing here is that this API previously returned the actually-used degree, whereas now it returns the "required" degree, and the actual degree will fluctuate based on the degree of the rest of the circuit. It would be helpful to document this API or rename it (to distinguish it from other `required_degree` APIs that behave differently).

src/plonk/permutation/prover.rs
```diff
@@ -50,103 +59,145 @@ impl Argument {
...
-                }
-            });
-        }
-
-        // Invert to obtain the denominator for the permutation product polynomial
-        modified_values.batch_invert();
+        // How many columns can be included in a single permutation polynomial?
+        // We need to multiply by z(X) and (1 - (l_last(X) + l_blind(X))). This
+        // will never underflow because of the requirement of at least a degree
+        // 3 circuit for the permutation argument.
```
daira, line 65: Add an assertion to that effect.

src/plonk/circuit.rs
```diff
@@ -1104,6 +1072,42 @@ impl<F: Field> ConstraintSystem<F> {
 
         degree
     }
+
+    /// Compute the number of blinding factors necessary to perfectly blind
+    /// each of the prover's witness polynomials.
+    pub fn blinding_factors(&self) -> usize {
```
str4d, line 1078: Note to self: check these.

src/plonk/permutation/prover.rs
```diff
@@ -166,76 +217,125 @@ impl<C: CurveAffine> Committed<C> {
...
-                            .zip(values[start..].iter())
-                            .zip(permutation[start..].iter())
+                    .collect::<Vec<_>>()
+                    .into_iter()
+                    .map(move |(coset, coset_last)| (coset - &coset_last) * &pk.l0),
+            )
+            // And for all the sets we enforce:
+            // (1 - (l_last + l_blind)) * (
+            //   z_i(\omega X) \prod (p(X) + \beta s_i(X) + \gamma)
+            // - z_i(X) \prod (p(X) + \delta^i \beta X + \gamma)
```
daira, line 266: \prod over what range? (See #320 for what I think it is.)

src/plonk/permutation/prover.rs
```diff
@@ -166,76 +217,125 @@ impl<C: CurveAffine> Committed<C> {
...
-                        }
-                        Any::Instance => {
-                            &instance_cosets[pk.vk.cs.get_any_query_index(column, Rotation::cur())]
-                        }
+                    .skip(1)
+                    .zip(self.sets.iter())
+                    .map(|(set, last_set)| {
+                        (
+                            set.permutation_product_coset.clone(),
+                            last_set.permutation_product_coset_last.clone().unwrap(),
```
daira, line 256: Isn't it simpler to do `(set.permutation_product_coset.clone() - last_set.permutation_product_coset_last.clone().unwrap()) * &pk.l0` here, rather than two maps?

src/dev.rs
```diff
@@ -407,24 +408,27 @@ impl<F: FieldExt> MockProver<F> {
...
+            n: n as u32,
             cs,
             regions: vec![],
             current_region: None,
             fixed,
             advice,
             instance,
-            permutations,
+            permutation,
+            usable_rows: ..n - (blinding_factors + 1),
```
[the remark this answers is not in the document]
str4d: I think there's an off-by-one here. Let's say we need 5 blinding factors, and we set `n = 8`. Then this will be `..(8 - (5 + 1))` = `..2`, but after looking at `ConstraintSystem::minimum_rows` I'd expect the usable range to be `..1`. Perhaps this should be:
```suggestion
            usable_rows: ..n - unusable_rows,
```

src/dev.rs
```diff
@@ -407,24 +408,27 @@ impl<F: FieldExt> MockProver<F> {
...
-            .collect();
+        if n < cs.minimum_rows() {
+            return Err(Error::NotEnoughRowsAvailable);
+        }
+
+        let fixed = vec![vec![None; n]; cs.num_fixed_columns];
+        let advice = vec![vec![None; n]; cs.num_advice_columns];
+        let permutation = permutation::keygen::Assembly::new(n, &cs.permutation);
 
+        let blinding_factors = cs.blinding_factors();
```
str4d, line 421: And then:
```suggestion
        let unusable_rows = cs.minimum_rows() - 1;
```

src/plonk/circuit.rs
```diff
@@ -1104,6 +1072,42 @@ impl<F: Field> ConstraintSystem<F> {
...
+    }
+
+    /// Returns the minimum necessary rows that need to exist in order to
+    /// account for e.g. blinding factors.
+    pub fn minimum_rows(&self) -> usize {
+        self.blinding_factors() // m blinding factors
+            + 1 // for l_{-(m + 1)} (l_last)
+            + 1 // for l_0
+            + 1 // for at least one row
+    }
```
str4d, line 1110: Reviewed up to here.

str4d (commented):
Flushing comments.

src/plonk/permutation/verifier.rs
```diff
@@ -76,87 +104,145 @@ impl<C: CurveAffine> Evaluated<C> {
...
-            .chain(Some(
-                l_0 * &(C::Scalar::one() - &self.permutation_product_eval),
-            ))
-            // z(X) \prod (p(X) + \beta s_i(X) + \gamma)
-            // - z(omega^{-1} X) \prod (p(X) + \delta^i \beta X + \gamma)
-            .chain(Some({
-                let mut left = self.permutation_product_eval;
-                for (eval, permutation_eval) in p
-                    .columns
+            // Enforce only for the first set.
```
daira, line 118: Is it possible to reduce the duplication with the prover? (Does not block.)

src/plonk/verifier.rs
```diff
@@ -70,13 +70,9 @@ pub fn verify_proof<'a, C: CurveAffine, E: EncodedChallenge<C>, T: TranscriptRea
...
     let permutations_committed = (0..num_proofs)
-        .map(|_| -> Result<Vec<_>, _> {
+        .map(|_| {
             // Hash each permutation product commitment
-            vk.cs
-                .permutations
-                .iter()
-                .map(|argument| argument.read_product_commitment(transcript))
-                .collect::<Result<Vec<_>, _>>()
+            vk.cs.permutation.read_product_commitments(&vk, transcript)
```
daira, line 75: Fix the clippy lint.

daira (approved):
utACK

src/plonk/permutation/prover.rs
ebfull: I meant for this to be an informal/implicit product but I did omit the correct indicies, I'll fix that to clarify.

src/plonk/permutation/prover.rs
ebfull: Yeah, I was trying to avoid returning an iterator that needed ownership or a reference to `self.sets` which is being dropped at the end of this method. It's more efficient to just map like you're saying and just collect it, not sure why I didn't just do that.

src/dev.rs
```diff
@@ -296,6 +296,16 @@ impl<F: Field + Group> Assignment<F> for MockProver<F> {
         self.assign_fixed(annotation, selector.0, row, || Ok(F::one()))
     }
 
+    /// Query the value of the cell of an instance column at a particular
+    /// absolute row, if known.
+    fn query_instance(&self, column: Column<Instance>, row: usize) -> Result<Option<F>, Error> {
```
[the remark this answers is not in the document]
ebfull: Good question. _Technically_ the instance columns have every row usable. However, so do fixed columns and we prohibit assignment to those outside the range right now. I'll add the range check to the implementations of `query_instance` and we can think about this later. (I can imagine at least one reason why we might want to remove this limitation but that's for another time.)

src/dev.rs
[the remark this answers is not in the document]
ebfull: The off-by-one isn't here but in `minimum_rows()`, where I'm being strict (read: cautious) about leaving enough breathing room for the permutation argument. For example, I thought it would be weird if someone set up a circuit such that the permutation polynomial ends up looking like `[1, {0,1}, blind, ...]` over the evaluation domain, or even `[1, blind, ...]`. I didn't want to have to analyze these cases so I essentially force all circuits to have permutation polynomials that are at _least_ like `[1, something, {0,1}, blind, ...]` so there's clear separation for analysis purposes.

As an example, if `l_last = l_0` because we're not strict enough in this separation, then the proof could end up repeating the evaluations of the permutation polynomial in the proof. Could lead to bugs and it's weird.

I'll document this in `minimum_rows()` but I'm going to keep the existing (strict) behavior.

src/dev.rs
ebfull: I don't want to use `minimum_rows()` like this. We may need to change `minimum_rows()` later for a reason that has nothing to do with how many rows are usable within ordinary advice columns, for instance.

ebfull: See also [my comment](https://github.com/zcash/halo2/pull/316#discussion_r667347262) about the role of `minimum_rows()`.

src/plonk/permutation/verifier.rs
ebfull: It's too challenging for this PR, but at least theoretically possible for some of the expressions. We should explore this.

src/plonk/prover.rs
```diff
@@ -35,7 +37,7 @@ pub fn create_proof<
     params: &Params<C>,
     pk: &ProvingKey<C>,
     circuits: &[ConcreteCircuit],
-    instances: &[&[Polynomial<C::Scalar, LagrangeCoeff>]],
+    instances: &[&[&[C::Scalar]]],
```
str4d, line 40: Document that the inner columns (the innermost slice) does not need to be a full column; `create_proof` zero-extends all columns (by virtue of copying the values into an initialized `Polynomial`).

src/plonk/prover.rs
```diff
@@ -53,16 +55,26 @@ pub fn create_proof<
...
 
     let instance: Vec<InstanceSingle<C>> = instances
         .iter()
         .map(|instance| -> Result<InstanceSingle<C>, Error> {
-            let instance_commitments_projective: Vec<_> = instance
+            let instance_values = instance
+                .iter()
+                .map(|values| {
+                    let mut poly = domain.empty_lagrange();
+                    for (poly, value) in poly.iter_mut().zip(values.iter()) {
```
str4d, line 71: We should check that `values` is not longer than `poly`, or else this will silently drop the tail of instance columns that are over-long.

src/plonk/verifier.rs
```diff
@@ -13,13 +13,13 @@ use crate::poly::{
...
 
 /// Returns a boolean indicating whether or not the proof is valid
-pub fn verify_proof<'a, C: CurveAffine, E: EncodedChallenge<C>, T: TranscriptRead<C, E>>(
-    params: &'a Params<C>,
+pub fn verify_proof<'params, C: CurveAffine, E: EncodedChallenge<C>, T: TranscriptRead<C, E>>(
+    params: &'params Params<C>,
     vk: &VerifyingKey<C>,
-    msm: MSM<'a, C>,
+    msm: MSM<'params, C>,
     instance_commitments: &[&[C]],
```
str4d, line 20: Having changed `create_proof` to no longer pass in a `Polynomial` (instead creating it internally and copying in the , we should change this as well to pass in the instance columns directly and then commit to them inside `verify_proof`. Alternatively, if we want to preserve the ability to pass in instance commitments, then we should provide a helper for this conversion, and document its indended usage in the `verify_proof` doc comment.

src/dev.rs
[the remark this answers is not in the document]
str4d: In that case, document here that this `+ 1` is for `l_{-(m + 1)} (l_last)`, and that we do not need to prevent circuits from assigning to the `l_0` breathing-room row.

benches/plonk.rs
```diff
@@ -223,7 +221,7 @@ fn bench_with_k(name: &str, k: u32, c: &mut Criterion) {
         ) -> Result<(), Error> {
             let cs = StandardPlonk::new(config);
 
-            for _ in 0..(1 << (self.k - 1)) {
+            for _ in 0..((1 << (self.k - 1)) - 5) {
```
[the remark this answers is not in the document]
ebfull: Changed it to 3, thanks!

str4d (commented):
Flushing comments from partial review of 0dc4447ad8abe8f2deb5b2e9877074a00111da2e.

src/plonk/lookup/prover.rs
```diff
@@ -327,35 +330,42 @@ impl<C: CurveAffine> Permuted<C> {
         // Compute the evaluations of the lookup product polynomial
         // over our domain, starting with z[0] = 1
         let z = iter::once(C::Scalar::one())
-            .chain(lookup_product.into_iter().skip(1))
+            .chain(lookup_product)
```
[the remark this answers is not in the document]
ebfull: We talked about this in a pairing.

This witness generation code needs to produce `u` values for `z`; given that z[0] = 1, we therefore need u - 1 lookup products to finish populating `z`. We don't need to produce the `u + 1` value because we don't need it.

src/plonk/lookup/prover.rs
```diff
@@ -327,35 +330,42 @@ impl<C: CurveAffine> Permuted<C> {
...
             .scan(C::Scalar::one(), |state, cur| {
                 *state *= &cur;
                 Some(*state)
             })
+            // Take all rows including the "last" row which should
+            // be a boolean (and ideally 1, else soundness is broken)
+            .take(params.n as usize - blinding_factors)
```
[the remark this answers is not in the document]
ebfull: We could and it's a minor optimization. https://github.com/zcash/halo2/issues/323

src/plonk/lookup/prover.rs
```diff
@@ -414,16 +432,25 @@ impl<'a, C: CurveAffine> Committed<C> {
...
+            // l_0(X) * (1 - z(X)) = 0
             .chain(Some(
                 Polynomial::one_minus(self.product_coset.clone()) * &pk.l0,
             ))
-            // z'(X) (a'(X) + \beta) (s'(X) + \gamma)
-            // - z'(\omega^{-1} X) (\theta^{m-1} a_0(X) + ... + a_{m-1}(X) + \beta) (\theta^{m-1} s_0(X) + ... + s_{m-1}(X) + \gamma)
+            // l_last(X) * (z(X)^2 - z(X)) = 0
+            .chain(Some(
+                (self.product_coset.clone() * &self.product_coset - &self.product_coset)
+                    * &pk.l_last,
```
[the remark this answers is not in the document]
ebfull: These expressions are not encoded into the `PinnedVerificationKey` because there aren't many degrees of freedom for them. We could reorder these to be consistent with the book; the reason it's not done here is because it would require cloning `pk.l_last` in the current API.

src/plonk/vanishing/prover.rs
```diff
@@ -64,30 +104,48 @@ impl<C: CurveAffine> Argument<C> {
...
 impl<C: CurveAffine> Constructed<C> {
     pub(in crate::plonk) fn evaluate<E: EncodedChallenge<C>, T: TranscriptWrite<C, E>>(
         self,
         x: ChallengeX<C>,
+        xn: C::Scalar,
+        domain: &EvaluationDomain<C::Scalar>,
         transcript: &mut T,
     ) -> Result<Evaluated<C>, Error> {
-        let h_evals: Vec<_> = self
+        let h_poly = self
```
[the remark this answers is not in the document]
str4d: Understanding confirmed. Book still needs to be updated.

src/poly/multiopen.rs
```diff
@@ -279,21 +321,9 @@ fn test_roundtrip() {
...
-                    commitment: &b,
-                    eval: avx, // NB: wrong!
-                }))
-                .chain(Some(VerifierQuery {
-                    point: y,
-                    commitment: &c,
-                    eval: cvy,
-                })),
+                .chain(Some(VerifierQuery::new_commitment(&a, x, avx)))
+                .chain(Some(VerifierQuery::new_commitment(&b, x, avx))) // NB: wrong!
```
therealyingtong, line 325:
```suggestion
                .chain(Some(VerifierQuery::new_commitment(&b, x, avx)))
```

str4d (commented):
Flushing more comments.

src/poly/domain.rs
```diff
@@ -387,18 +387,71 @@ impl<G: Group> EvaluationDomain<G> {
...
+        // Then l_i(x) for some x is evaluated as
+        // ((x^n - 1) / (x - omega^i)) * (1 / \prod_{j=0,j \neq i}^{n - 1} (omega^i - omega^j))
+        // We refer to
+        // (1 / \prod_{j=0,j \neq i}^{n - 1} (omega^i - omega^j))
+        // as the barycentric weight of omega^i.
+        // We know that for i = 0
+        // (1 / \prod_{j=0,j \neq i}^{n - 1} (omega^i - omega^j))
+        // = (1 / n)
+        // If we multiply (1 / n) by (omega^{i})^-1 then we obtain
+        // (1 / \prod_{j=0,j \neq 0}^{n - 1} (omega^{i + 1} - omega^{j + 1}))
```
therealyingtong, line 424:
```suggestion
        // (1 / \prod_{j=0,j \neq 0}^{n - 1} (omega^{i} - omega^{j + i}))
```

src/plonk/lookup/prover.rs
```diff
@@ -372,13 +382,21 @@ impl<C: CurveAffine> Permuted<C> {
 
                 assert_eq!(left, right);
             }
+
+            // l_last(X) * (z(X)^2 - z(X)) = 0
+            // Assertion will fail only when soundness is broken, in which
+            // case this z[0] value will be zero. (bad!)
```
therealyingtong, line 388:
```suggestion
            // case this z[u] value will be zero. (bad!)
```

src/plonk/permutation.rs
```diff
@@ -20,24 +18,52 @@ pub(crate) struct Argument {
...
+        // following will not affect the required degree of
+        // this middleware.
+        //
+        // (1 - (l_last(X) + l_blind(X))) * (
+        //   z(\omega X) \prod (p(X) + \beta s_i(X) + \gamma)
+        // - z(X) \prod (p(X) + \delta^i \beta X + \gamma)
+        // )
+        //
+        // On the first sets of columns, except the first
+        // column, we will do
```
therealyingtong, line 43:
```suggestion
        // On all sets of columns except the first set,
```

src/plonk/lookup/prover.rs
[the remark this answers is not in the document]
therealyingtong: Also, where we were doing Z(X) previously, we are now doing Z(ωX).

src/plonk/permutation/prover.rs
```diff
@@ -50,103 +59,146 @@ impl Argument {
...
-        }
-
-        // Invert to obtain the denominator for the permutation product polynomial
-        modified_values.batch_invert();
+        // How many columns can be included in a single permutation polynomial?
+        // We need to multiply by z(X) and (1 - (l_last(X) + l_blind(X))). This
+        // will never underflow because of the requirement of at least a degree
+        // 3 circuit for the permutation argument.
+        assert!(pk.vk.cs.degree() >= 3);
+        let chunk_len = pk.vk.cs.degree() - 2;
```
therealyingtong, line 67: `chunk_len = pk.vk.cs.degree() - 1` still works for me. Is this expected?

therealyingtong, line 67: `chunk_len = pk.vk.cs.degree()` still works for me. Is this expected?

src/plonk.rs
```diff
@@ -155,6 +150,10 @@ pub enum Error {
     OpeningError,
     /// Transcript error
     TranscriptError,
+    /// Insufficient number of rows available
```
therealyingtong, line 153:
```suggestion
    /// Instance provided has fewer than the minimum rows required.
```

src/plonk/prover.rs
```diff
@@ -212,15 +251,30 @@ pub fn create_proof<
                 }
             }
 
+            let unusable_rows_start = params.n as usize - (meta.blinding_factors() + 1);
```
therealyingtong, line 254: This still works for me when `unusable_rows_start = params.n as usize - meta.blinding_factors()`. Is that expected?

src/plonk/vanishing/prover.rs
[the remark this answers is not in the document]
therealyingtong: Should the book be updated to include `random_poly_committed` as well?

src/plonk/circuit.rs
therealyingtong: Off-by-one errors such as?

examples/two-chip.rs
```diff
@@ -604,7 +551,7 @@ fn main() {
     // ANCHOR: test-circuit
     // The number of rows in our circuit cannot exceed 2^k. Since our example
     // circuit is very small, we can pick a very small value here.
-    let k = 3;
+    let k = 4;
```
therealyingtong, line 554: `cargo run --example two-chip` is currently failing for me (`InstanceTooLarge`).

examples/simple-example.rs
```diff
@@ -339,7 +306,7 @@ fn main() {
     // ANCHOR: test-circuit
     // The number of rows in our circuit cannot exceed 2^k. Since our example
     // circuit is very small, we can pick a very small value here.
-    let k = 3;
+    let k = 4;
```
therealyingtong, line 309: `cargo run --example simple-example` is currently failing for me (`InstanceTooLarge`).

src/plonk/permutation/prover.rs
```diff
@@ -297,20 +419,43 @@ impl<C: CurveAffine> Evaluated<C> {
...
-                poly: &self.constructed.permutation_product_poly,
-                blind: self.constructed.permutation_product_blind,
-            }))
-            .chain(Some(ProverQuery {
-                point: x_inv,
-                poly: &self.constructed.permutation_product_poly,
-                blind: self.constructed.permutation_product_blind,
+            .chain(self.constructed.sets.iter().flat_map(move |set| {
+                iter::empty()
+                    // Open permutation product commitments at x and \omega^{-1} x
```
therealyingtong, line 432:
```suggestion
                    // Open permutation product commitments at x and \omega x
```

src/plonk/permutation/prover.rs
str4d: That suggests there aren't any tests that have enough columns to trigger the chunking.

examples/simple-example.rs
```diff
@@ -352,24 +319,18 @@ fn main() {
         b: Some(b),
     };
 
-    // Arrange the public input. We expose the multiplication result in row 6
+    // Arrange the public input. We expose the multiplication result in row 0
     // of the instance column, so we position it there in our public inputs.
     let mut public_inputs = vec![Fp::zero(); 1 << k];
```
str4d, line 324: This currently triggers an `InstanceTooLarge` error, because we are now limiting instance columns to `usable_rows` (in response to my earlier review).
```suggestion
    let mut public_inputs = vec![c];
```

examples/simple-example.rs
```diff
@@ -352,24 +319,18 @@ fn main() {
...
-    public_inputs[6] = c;
+    public_inputs[0] = c;
```
str4d, line 325:
```suggestion
```

examples/two-chip.rs
```diff
@@ -619,24 +566,18 @@ fn main() {
         c: Some(c),
     };
 
-    // Arrange the public input. We expose the multiplication result in row 6
+    // Arrange the public input. We expose the multiplication result in row 0
     // of the instance column, so we position it there in our public inputs.
     let mut public_inputs = vec![Fp::zero(); 1 << k];
```
str4d, line 571:
```suggestion
    let mut public_inputs = vec![d];
```

examples/two-chip.rs
```diff
@@ -619,24 +566,18 @@ fn main() {
...
-    public_inputs[7] = d;
+    public_inputs[0] = d;
```
str4d, line 572:
```suggestion
```

src/plonk.rs
ebfull: Hm, while the comment as-is is confusing I don't quite think this is the correct fix. I will change to

"Instance provided has more rows than supported by circuit"

src/plonk/permutation/prover.rs
ebfull: Updating the test to force there to be multiple sets.

src/plonk/vanishing/prover.rs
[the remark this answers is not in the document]
str4d: Yes, the vanishing argument book page needs to be updated with the ZK changes, as @daira did for the lookup and permutation pages.

src/plonk/circuit.rs
```diff
@@ -1102,7 +1083,46 @@ impl<F: Field> ConstraintSystem<F> {
...
+
+    /// Compute the number of blinding factors necessary to perfectly blind
+    /// each of the prover's witness polynomials.
+    pub fn blinding_factors(&self) -> usize {
+        // All of the prover's advice columns are evaluated at no more than
+        let factors = *self.num_advice_queries.iter().max().unwrap_or(&1);
+        // distinct points during gate checks.
+
+        // In the permutation and lookup argument the witness polynomials are
+        // evaluated at most 3 times:
```
str4d, line 1097: If I'm reading things correctly, the polynomials for each lookup argument are evaluated at most 2 times:
- `product_poly` at `cur` and `next`
- `permuted_input_poly` at `cur` and `prev`
- `permuted_table_poly` at `cur`

Meanwhile for the permutation argument, we evaluate `permutation_product_poly` at `cur`, `next`, and (for all except the final set) `last`.

So it would be somewhat clearer here to say:
```suggestion
        // - The permutation argument witness polynomials are evaluated at most 3 times.
        // - Each lookup argument has independent witness polynomials, and they are
        //   evaluated at most 2 times.
```

src/plonk/circuit.rs
```diff
@@ -1102,7 +1083,46 @@ impl<F: Field> ConstraintSystem<F> {
...
+        let factors = std::cmp::max(3, factors);
+
+        // Each polynomial is evaluated at most an additional time during
+        // multiopen:
```
str4d, line 1101:
```suggestion
        // multiopen (at x_3 to produce q_evals):
```

src/plonk/circuit.rs
```diff
@@ -1102,7 +1083,46 @@ impl<F: Field> ConstraintSystem<F> {
...
+        let factors = factors + 1;
+
+        // h(x) is derived by the other evaluations so it does not reveal
+        // anything; in fact it does not even appear in the proof.
+
+        // h(x_3) is also not revealed; the verifier only learns a single
+        // evaluation of a polynomial in x_1 which has h(x_3) and another random
+        // polynomial evaluated at x_3 as coefficients.
```
str4d, line 1109:
```suggestion
        // h(x_3) is also not revealed; the verifier only learns a single
        // evaluation of a polynomial in x_1 which has h(x_3) and another
        // random polynomial evaluated at x_3 as coefficients. Specifically,
        // this is final_poly in poly::commitment::prover::create_proof.
```

Also, to check my understanding: this evaluation is derived by the verifier from `q_evals` (which we have already accounted for in the blinding calculation), but this `h(x_3)` consideration is necessary due to the additional evaluation at `x_3` for the polynomial commitment opening proof?

src/plonk/circuit.rs
str4d: I have checked these, and left comments as relevant.

str4d (approved):
utACK 1f75dc25ee5e5ada33a37292904fe4558353fa2c

src/dev.rs
```diff
@@ -407,24 +410,43 @@ impl<F: FieldExt> MockProver<F> {
...
+        }
 
+        if instance.len() != cs.num_instance_columns {
+            return Err(Error::IncompatibleParams);
+        }
+
+        let instance = instance
+            .into_iter()
+            .map(|mut instance| {
+                if instance.len() > n - (cs.blinding_factors() + 1) {
```
daira, line 426:
```suggestion
        let blinding_factors = cs.blinding_factors();
        let usable_rows = n - (blinding_factors + 1);
        let instance = instance
            .into_iter()
            .map(|mut instance| {
                if instance.len() > usable_rows {
```

src/dev.rs
```diff
@@ -407,24 +410,43 @@ impl<F: FieldExt> MockProver<F> {
...
+                instance.resize(n, F::zero());
+                Ok(instance)
+            })
+            .collect::<Result<Vec<_>, _>>()?;
+
+        let fixed = vec![vec![None; n]; cs.num_fixed_columns];
+        let advice = vec![vec![None; n]; cs.num_advice_columns];
+        let permutation = permutation::keygen::Assembly::new(n, &cs.permutation);
+
+        let blinding_factors = cs.blinding_factors();
```
daira, line 439:
```suggestion
```

src/dev.rs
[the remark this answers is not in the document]
daira:
```suggestion
            usable_rows: ..usable_rows,
```

src/plonk/verifier.rs
```diff
@@ -140,87 +153,81 @@ pub fn verify_proof<'a, C: CurveAffine, E: EncodedChallenge<C>, T: TranscriptRea
...
+                                &|a, scalar| a * &scalar,
+                            )
+                        })
+                    }))
+                    .chain(permutation.expressions(
+                        vk,
+                        &vk.cs.permutation,
+                        &advice_evals,
+                        &fixed_evals,
+                        &instance_evals,
```
daira, line 199:
```suggestion
                        advice_evals,
                        fixed_evals,
                        instance_evals,
```

src/plonk/verifier.rs
```diff
@@ -140,87 +153,81 @@ pub fn verify_proof<'a, C: CurveAffine, E: EncodedChallenge<C>, T: TranscriptRea
...
+                                    l_0,
+                                    l_last,
+                                    l_blind,
+                                    argument,
+                                    theta,
+                                    beta,
+                                    gamma,
+                                    &advice_evals,
+                                    &fixed_evals,
+                                    &instance_evals,
```
daira, line 222:
```suggestion
                                    advice_evals,
                                    fixed_evals,
                                    instance_evals,
```

src/poly/multiopen.rs
daira, line 325: Fix this, or remove the comment.

daira (approved):
Did another review pass; utACK modulo suggestions.

src/poly/multiopen.rs
str4d: This is supposed to be wrong; it's a test checking that the bug is detected.

src/plonk/circuit.rs
ebfull: Imagine a situation where the only polynomial opened at `x` is `h'(X)` -- which is what I call the polynomial produced by computing `H_0 + [x^n] H_1 + ...`. We don't want to reveal `h'(x_3)` because it leaks information that is not trivial to blind. To avoid this, we also _always_ open a polynomial `r(X)` at `x` as well, so it always ends up in the same query set as `h'(X)` does. This `r(X)` polynomial is totally random and is `random_poly_commitment` in the vanishing argument. As a result, the proof only contains the evaluation of `x_1 h'(x_3) + r(x_3)` (and perhaps other polynomials as well) which is uniformly and independently distributed from `h'(x_3)` and therefore `h'(x_3)` is not leaked.

src/dev.rs
ebfull: I don't see why we're mentioning this here, this is just dev tooling?

src/plonk/circuit.rs
ebfull: I instead went with adding "-- this random polynomial is "random_poly" in the vanishing argument." to the end where your suggestion was.

src/dev.rs
daira: Ah, I was just responding to @str4d's comment asking for more documentation. There's a case for adding `cs.usable_rows()`, but we shouldn't block the PR on that.

PR #330, opened 2021-07-15
Merge the common pieces of the permutation argument into a separate proof component

ebfull:
The permutation argument has some fixed polynomials (which define the permutation) that are evaluated at `x`. Right now the code evaluates them for every argument, but the evaluations do not vary across circuit instances. Similarly to how we merge fixed evaluations across circuit instances, we do the same with these permutation polynomial openings.

This reduces the proof size and proving time.


daira (approved):
utACK

str4d (approved):
utACK

Issue #329, opened 2021-07-15
MockProver: Track blinding values as "poison"


str4d:
`MockProver` checks circuits manually; in particular, it takes every custom gate, and applies their constraints to every row, to confirm that they pass. For unassigned cells, `MockProver` gives them a default value of zero (to match the behaviour of the regular prover).

#316 introduced blinding rows, which have random, very likely non-zero, values assigned to all cells. `MockProver` was not updated to account for this, and as a result when I tried to update the `orchard` crate to include these changes, we ran into a problem where `MockProver` passed but full verification failed. The cause was that the dummy circuit had a custom constraint forcing all cells to equal zero, which was not gated behind a selector (and thus was applying even to blinding rows).

We _could_ assign random values in `MockProver`, but that could sometimes (though with low-probability) incorrectly pass. @daira suggested that we could instead do something similar to what LLVM does for undefined behaviour: track these as "poison", and propagate the poison where it is used.

Currently, `MockProver` stores cells as `Option<F>`, where `None` represents "unassigned", and gets mapped to zero during gate checks. We should change this to an enum with explicit cases:
- `Unassigned`
  - Maps to zero when checking constraints.
- `Assigned(F)`
- `MultipleAssigned(Vec<F>)`
  - Not necessarily in this issue, but we could subsequently do this for #257.
- `Poison`
  - Propagates when checking constraints except when it is multiplied by zero (since that is how we leave cells unconstrained).

daira:
Can you explain why the Poison and Unassigned cases need to be distinguished (i.e. why they should not both be treated as poison when checking constraints)? I think the semantics should be that cells in *selector* columns default to zero, but accessing any other cell that was not assigned should be an error. I asked this in our tripling but I didn't quite follow the answer.

str4d:
We already have a pass that checks for unassigned cells, and the current behaviour of the main prover is to default init cells to zero, so I'm not inclined to spend the time add this extra complexity to `MockProver` right now. We could maybe do that in future though.

PR #331, opened 2021-07-15
MockProver: Poison unusable advice cells

str4d:
Circuit developers need to use selectors (or equivalent) to prevent their
custom gates from being active on unusable rows. By marking the advice
cells in these rows as "poisoned" and tracking this through expressions,
we can provide a better error message to developers that indicates the
missing selector issue.

Closes zcash/halo2#329.

### DIFF
--- a/src/dev.rs
+++ b/src/dev.rs
@@ -671,56 +671,62 @@ impl<F: FieldExt> MockProver<F> {
                 .iter()
                 .enumerate()
                 .flat_map(|(lookup_index, lookup)| {
+                    let load = |expression: &Expression<F>, row| {
+                        expression.evaluate(
+                            &|scalar| Value::Real(scalar),
+                            &|index| {
+                                let query = self.cs.fixed_queries[index];
+                                let column_index = query.0.index();
+                                let rotation = query.1 .0;
+                                self.fixed[column_index]
+                                    [(row as i32 + n + rotation) as usize % n as usize]
+                                    .into()
+                            },
+                            &|index| {
+                                let query = self.cs.advice_queries[index];
+                                let column_index = query.0.index();
+                                let rotation = query.1 .0;
+                                self.advice[column_index]
+                                    [(row as i32 + n + rotation) as usize % n as usize]
+                                    .into()
+                            },
+                            &|index| {
+                                let query = self.cs.instance_queries[index];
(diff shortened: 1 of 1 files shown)


src/dev.rs
```diff
@@ -613,7 +713,7 @@ impl<F: FieldExt> MockProver<F> {
                             .iter()
                             .map(|c| load(c, input_row))
                             .collect();
-                        let lookup_passes = (0..n)
+                        let lookup_passes = (0..n as usize)
```
ebfull, line 716: Shouldn't this also be `usable_rows`?

ebfull (changes requested):

src/dev.rs
str4d: Are input expressions on usable rows actually prevented from being equal to table expressions on unusable rows?

str4d: Ok, so I've looked at the lookup argument again, and both the input and table expressions are only permuted over the usable rows. So this indeed should be `usable_rows`.

src/dev.rs
```diff
@@ -130,6 +137,76 @@ impl Region {
...
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+enum Value<F: Group + Field> {
+    Real(F),
+    Poison,
+}
+
+impl<F: Group + Field> From<CellValue<F>> for Value<F> {
+    fn from(value: CellValue<F>) -> Self {
+        match value {
+            // Cells that haven't been explicitly assigned to, default to zero.
```
daira, line 161: I still disagree with this; it should only be the case for cells in selector columns.

src/dev.rs
```diff
@@ -130,6 +137,76 @@ impl Region {
...
+    }
+}
+
+impl<F: Group + Field> Mul for Value<F> {
+    type Output = Self;
+
+    fn mul(self, rhs: Self) -> Self::Output {
+        match (self, rhs) {
+            (Value::Real(a), Value::Real(b)) => Value::Real(a * b),
+            // If poison is multiplied by zero, then we treat the poison as unconstrained
```
daira, line 186: "unconstrained" isn't the right word here, I think. The zero eliminates the poison.

daira (commented):
Post-hoc ACK with minor comments.

PR #336, opened 2021-07-16
dev: Enable visualizing equality constraints in circuit layouts

str4d:
This makes it easier to see how the global permutation argument is being used, and also enables us to hunt for missing equality constraints.

### DIFF
--- a/src/dev/graph/layout.rs
+++ b/src/dev/graph/layout.rs
@@ -60,7 +60,7 @@ impl CircuitLayout {
         self
     }
 
-    /// Draws red lines between equality-constrainted cells.
+    /// Draws red lines between equality-constrained cells.
     ///
     /// The default is to not show these, as they can get _very_ messy.
     pub fn show_equality_constraints(mut self, show: bool) -> Self {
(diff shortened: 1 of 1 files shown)


src/dev/graph/layout.rs
```diff
@@ -50,6 +52,22 @@ impl CircuitLayout {
...
 
+    /// Marks cells involved in equality constraints, in red.
+    ///
+    /// The default is to not mark these cells.
+    pub fn mark_equality_cells(mut self, show: bool) -> Self {
+        self.mark_equality_cells = show;
+        self
+    }
+
+    /// Draws red lines between equality-constrainted cells.
```
daira, line 63:
```suggestion
    /// Draws red lines between equality-constrained cells.
```

src/dev/graph/layout.rs
```diff
@@ -203,6 +221,34 @@ impl CircuitLayout {
...
+                let l_col = column_index(l_col);
+                let r_col = column_index(r_col);
+                root.draw(&Rectangle::new(
+                    [(l_col, *l_row), (l_col + 1, *l_row + 1)],
+                    ShapeStyle::from(&RED).filled(),
+                ))?;
+                root.draw(&Rectangle::new(
+                    [(r_col, *r_row), (r_col + 1, *r_row + 1)],
+                    ShapeStyle::from(&RED).filled(),
+                ))?;
```
daira, line 236: It would be nice to have a border around each cell, so that it's easier to count them (similar to assigned cells).

daira (approved):
utACK with comments.

PR #337, opened 2021-07-16
Update sha256 example to work with new API

### LINKED ISSUE #390: Update `sha256` gadget.

therealyingtong:
Closes #390. Based on #383.   

Run `cargo test --example sha256` to test.

### DIFF
--- a/examples/sha256/main.rs
+++ b/examples/sha256/main.rs
@@ -28,10 +28,7 @@ pub trait Sha256Instructions<F: FieldExt>: Chip<F> {
     type State: Clone + fmt::Debug;
     /// Variable representing a 32-bit word of the input block to the SHA-256 compression
     /// function.
-    type BlockWord: Copy + fmt::Debug;
-
-    /// The zero BlockWord
-    fn zero() -> Self::BlockWord;
+    type BlockWord: Copy + fmt::Debug + Default;
 
     /// Places the SHA-256 IV in the circuit, returning the initial state variable.
     fn initialization_vector(&self, layouter: &mut impl Layouter<F>) -> Result<Self::State, Error>;
@@ -140,7 +137,7 @@ impl<F: FieldExt, Sha256Chip: Sha256Instructions<F>> Sha256<F, Sha256Chip> {
     ) -> Result<Sha256Digest<Sha256Chip::BlockWord>, Error> {
         // Pad the remaining block
         if !self.cur_block.is_empty() {
-            let padding = vec![Sha256Chip::zero(); BLOCK_SIZE - self.cur_block.len()];
+            let padding = vec![Sha256Chip::BlockWord::default(); BLOCK_SIZE - self.cur_block.len()];
             self.cur_block.extend_from_slice(&padding);
             self.state = self.chip.initialization(&mut layouter, &self.state)?;
             self.state = self.chip.compress(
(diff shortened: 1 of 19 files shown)


examples/sha256/main.rs
```diff
@@ -169,6 +168,5 @@ impl<F: FieldExt, Sha256Chip: Sha256Instructions<F>> Sha256<F, Sha256Chip> {
         hasher.finalize(layouter.namespace(|| "finalize"))
     }
 }
-*/
 
 fn main() {}
```
daira, line 172: This example still doesn't do anything.

examples/sha256/table16/compression/compression_gates.rs
```diff
@@ -56,14 +63,12 @@ impl<F: FieldExt> CompressionGate<F> {
...
-                    + dense_check
-                    + spread_check
-                    + check_spread_and_range),
-        )
+        check_spread_and_range
+            .chain(Some(("range_check_tag_b", range_check_tag_b)))
+            .chain(Some(("range_check_tag_d", range_check_tag_d)))
+            .chain(Some(("dense_check", dense_check)))
+            .chain(Some(("spread_check", spread_check)))
+            .map(move |(name, poly)| (name, s_decompose_abcd.clone() * poly))
```
daira, line 71: Were the gates using the `selector * (check1 + check2 + ...)` pattern previously incorrect?

daira (approved):
utACK with a question.

examples/sha256/table16/compression/compression_gates.rs
therealyingtong: Yes, the checks should have been linearly independent instead of being summed.

examples/sha256/main.rs
therealyingtong: Yes, the actual test code is in `examples/sha256/benches.rs`. But I think we chose to exclude it from `sha256/main.rs` because it would slow down CI significantly.

dan-da:
how does one actually run this example?    I tried:

```
$ cargo run --example sha256
```

but that doesn't do anything because main() is empty in main.rs.  So then I tried to run benchmarks with:

```
$ cargo bench --example sha256
...
warning: 3 warnings emitted
...

running 3 tests
test table16::compression::tests::compress ... ignored
test table16::message_schedule::tests::message_schedule ... ignored
test table16::spread_table::tests::lookup_table ... ignored

test result: ok. 0 passed; 0 failed; 3 ignored; 0 measured; 0 filtered out; finished in 0.00s
```

but that doesn't seem to do anything either.   what's the trick?

examples/sha256/table16/compression/compression_util.rs
```diff
@@ -38,101 +38,155 @@ pub const SUBREGION_MAIN_WORD: usize =
...
+}
+
+impl RoundIdx {
+    pub(crate) fn as_usize(&self) -> usize {
+        match self {
+            Self::Main(idx) => *idx,
+            _ => panic!(),
+        }
+    }
+}
```
str4d, line 55: Given how many methods panic on `RoundIdx::Init`, I would define a separate `MainRound` struct for them, and leave the match on `RoundIdx` to the caller. Non-blocking, I'm interested in doing this myself in a separate PR.

str4d (approved):
utACK. I did not check this particularly thoroughly (or against the book much), but I did check that e.g. cell assignments weren't mis-refactored, and that the constraint bugfixes made sense. I'm satisfied with this being merged as an example, and subsequently migrated to `halo2_gadgets` behind an `unstable` feature flag (to indicate its lack of thorough review and likeliness to be completely rewritten).

PR #341, opened 2021-07-20
Add `RegionLayouter::constrain_instance` method

therealyingtong:
This allows us to call `region.constrain_instance()` (where previously we had to call `layouter.constrain_instance()`.

### DIFF
--- a/src/circuit/floor_planner/single_pass.rs
+++ b/src/circuit/floor_planner/single_pass.rs
@@ -226,6 +226,20 @@ impl<'r, 'a, F: Field, CS: Assignment<F> + 'a> RegionLayouter<F>
         Ok((cell, value))
     }
 
+    fn constrain_instance<'v>(
+        &mut self,
+        cell: Cell,
+        instance: Column<Instance>,
+        row: usize,
+    ) -> Result<(), Error> {
+        self.layouter.cs.copy(
+            cell.column,
+            *self.layouter.regions[*cell.region_index] + cell.row_offset,
+            instance.into(),
+            row,
+        )
+    }
+
     fn assign_fixed<'v>(
         &'v mut self,
         annotation: &'v (dyn Fn() -> String + 'v),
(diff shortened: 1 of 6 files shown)


src/circuit/floor_planner/single_pass.rs
```diff
@@ -226,6 +226,20 @@ impl<'r, 'a, F: Field, CS: Assignment<F> + 'a> RegionLayouter<F>
...
+        cell: Cell,
+        instance: Column<Instance>,
+        row: usize,
+    ) -> Result<(), Error> {
+        self.layouter.cs.copy(
+            cell.column,
+            *self.layouter.regions[*cell.region_index] + cell.row_offset,
+            instance.into(),
+            row,
+        )
```
str4d, line 240:
```suggestion
        self.layouter.constrain_instance(
            cell,
            instance,
            row,
        )
```
(plus whatever `rustfmt` wants)

src/dev.rs
```diff
@@ -66,7 +66,7 @@ pub enum VerifyFailure {
     /// A permutation did not preserve the original value of a cell.
     Permutation {
         /// The column in which this permutation is not satisfied.
-        column: Column<Any>,
+        column: metadata::Column,
```
str4d, line 69: What is the motivation for this change?

str4d (commented):
I'm not against this change, but I think it's a sign that the `Region::assign_advice_from_instance` API is misleading. We need it on `Region` in order to be able to assign to a cell within a region (which has no absolute position), but the API should only be used in contexts where we know the region logic is specific to the given circuit (or sufficiently abstracted away), since absolute references to instance column rows will make chips non-portable. We should document these APIs more clearly, and think about how better to handle this abstraction boundary.

src/dev.rs
therealyingtong: I want to be able to specify a column in an expected `VerifyFailure` error, without constructing an actual column: https://github.com/zcash/orchard/pull/101/files#diff-a61514f6b9ddb874831490698e7a84231d985315c900accb9c80c302dda52b9aR490

therealyingtong:
I no longer need this after #342.

However, I still need some patch to the MockProver `metadata`: https://github.com/zcash/halo2/pull/343

PR #342, opened 2021-07-20
Add `Region::assign_advice_from_constant` API

### LINKED ISSUE #335: Add `Region::assign_advice_from_constant` API
A common pattern inside regions is to have a custom gate that constrains some repetitive process. We often need to constrain the initial value for this process to some constant. Right now, the way we do this is to use `Region::assign_fixed` to allocate that constant, and then `Region::constrain_equal` to constrain the initial advice cell value to that constant.

This pattern requires a fixed column to be locked by the region for its duration, even if we only use a single cell in it. In the example below, red cells are equality-constrained, and you can see we require two mostly-empty fixed columns since we want to run two instances of `hash_to_point` in parallel:

image

We should instead have a `Region::assign_advice_from_constant` API, which uses one or more predefined global fixed columns for assigning these constants. The layouter would lay out constants in these columns kind of like a fixed table. The fixed columns themselves could perhaps be shared with chips, with the layouter filling in the spare gaps with these constants. If circuit synthesis would cause the available fixed columns to be filled with nowhere to overflow into, an error would be returned.

str4d:
Closes #335.

### DIFF
--- a/examples/simple-example.rs
+++ b/examples/simple-example.rs
@@ -5,7 +5,7 @@ use std::marker::PhantomData;
 use halo2::{
     arithmetic::FieldExt,
     circuit::{Cell, Chip, Layouter, Region, SimpleFloorPlanner},
-    plonk::{Advice, Circuit, Column, ConstraintSystem, Error, Instance, Selector},
+    plonk::{Advice, Circuit, Column, ConstraintSystem, Error, Fixed, Instance, Selector},
     poly::Rotation,
 };
 
@@ -16,6 +16,9 @@ trait NumericInstructions<F: FieldExt>: Chip<F> {
 
     /// Loads a number into the circuit as a private input.
     fn load_private(&self, layouter: impl Layouter<F>, a: Option<F>) -> Result<Self::Num, Error>;
+
+    /// Loads a number into the circuit as a fixed constant.
+    fn load_constant(&self, layouter: impl Layouter<F>, constant: F) -> Result<Self::Num, Error>;
 
     /// Returns `c = a * b`.
     fn mul(
@@ -62,6 +65,9 @@ struct FieldConfig {
     // This is important when building larger circuits, where columns are used by
     // multiple sets of instructions.
     s_mul: Selector,
+
+    /// The fixed column used to load constants.
+    constant: Column<Fixed>,
 }
 
 impl<F: FieldExt> FieldChip<F> {
@@ -76,8 +82,10 @@ impl<F: FieldExt> FieldChip<F> {
         meta: &mut ConstraintSystem<F>,
         advice: [Column<Advice>; 2],
         instance: Column<Instance>,
+        constant: Column<Fixed>,
     ) -> <Self as Chip<F>>::Config {
         meta.enable_equality(instance.into());
+        meta.enable_constant(constant);
         for column in &advice {
             meta.enable_equality((*column).into());
         }
@@ -117,6 +125,7 @@ impl<F: FieldExt> FieldChip<F> {
             advice,
             instance,
             s_mul,
+            constant,
         }
     }
 }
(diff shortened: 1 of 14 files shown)


therealyingtong:
[Force-pushed](https://github.com/zcash/halo2/compare/afdd0d1f659ee376854b027267d17ce1fd187280..62f088e2f249dfce2c3bcab7321cba0d99697af9) to rebase on main.

examples/simple-example.rs
```diff
@@ -172,6 +181,33 @@ impl<F: FieldExt> NumericInstructions<F> for FieldChip<F> {
...
+    ) -> Result<Self::Num, Error> {
+        let config = self.config();
+
+        let mut num = None;
+        layouter.assign_region(
+            || "load constant",
+            |mut region| {
+                let cell = region.assign_advice_from_constant(
+                    || "constant value",
+                    config.advice[0],
```
daira, line 197: Why always `config.advice[0]`?

src/circuit/floor_planner/single_pass.rs
```diff
@@ -247,6 +271,29 @@ impl<'r, 'a, F: Field, CS: Assignment<F> + 'a> RegionLayouter<F>
...
+
+        // For the simple layouter, just assign the fixed constants inside the region
+        // using the first constants column. self.next_constant_row is updated by this
+        // function call.
+        let fixed = self.assign_fixed(
+            &|| "constant".into(),
+            self.layouter.constants[0],
+            // Convert the absolute row into a relative offset within this region, but
+            // always at an offset that is not before this region (taking advantage of the
+            // fact that for this single-pass layouter, regions are always layed out in
```
daira, line 287:
```suggestion
            // fact that for this single-pass layouter, regions are always laid out in
```

src/plonk.rs
```diff
@@ -154,6 +154,9 @@ pub enum Error {
     NotEnoughRowsAvailable,
     /// Instance provided exceeds number of available rows
     InstanceTooLarge,
+    /// Circuit synthesis requires global constants, but circuit configuration did not
+    /// call [`ConstraintSystem::enable_constant`] on fixed columns with sufficient space.
+    NotEnoughColumnsForConstants,
```
daira, line 159: Please add a test that triggers this error.

daira (approved):
utACK, with request for a test that triggers `NotEnoughColumnsForConstants`.

src/circuit/layouter.rs
```diff
@@ -152,6 +171,17 @@ impl<F: Field> RegionLayouter<F> for RegionShape {
...
 
+    fn assign_advice_from_constant<'v>(
+        &'v mut self,
+        annotation: &'v (dyn Fn() -> String + 'v),
+        column: Column<Advice>,
+        offset: usize,
+        constant: Assigned<F>,
+    ) -> Result<Cell, Error> {
+        // The rest is identical to witnessing an advice cell.
+        self.assign_advice(annotation, column, offset, &mut || Ok(constant))
```
ebfull, line 182: Does this mean that the `assign_advice_from_instance` implementation just below this could be simplified as well?

ebfull (approved):

src/circuit/layouter.rs
str4d: There was a reason I decided not to do that at the time, but I don't remember what it was. I might be thinking of a different `assign_advice_from_instance` impl though. AFAICT `RegionShape::assign_advice_from_instance` could indeed be simplified like this.

str4d: Ah, this might have been it:
```
error[E0283]: type annotations needed
   --> src/circuit/layouter.rs:196:14
    |
196 |         self.assign_advice(annotation, advice, offset, &mut || unreachable!())
    |              ^^^^^^^^^^^^^ cannot infer type for type parameter `F` declared on the trait `RegionLayouter`
    |
    = note: cannot satisfy `_: Field`
```

Issue #325, opened 2021-07-14
Change canonical column ordering to Instance < Advice < Fixed


str4d:
When implementing the V1 layouter I picked somewhat arbitrarily "Advice < Instance < Fixed" as the canonical column ordering, for rendering purposes. However, #316 changes the way we do public inputs: we now copy public inputs into advice columns or vice versa, and otherwise never interact directly with instance columns. This means they can never occur inside a region, but having them be ordered in between advice and fixed columns means that regions are unnecessarily split up when rendered.

Placing instance columns on the left brings advice and fixed columns together, while keeping instance and advice columns adjacent for copy purposes (there is never a need to copy an instance cell into a fixed cell).

PR #344, opened 2021-07-20
Change canonical column ordering to Instance < Advice < Fixed

str4d:
Closes zcash/halo2#325.

### DIFF
--- a/src/plonk/circuit.rs
+++ b/src/plonk/circuit.rs
@@ -44,8 +44,8 @@ impl<C: ColumnType> Ord for Column<C> {
         // orderings.
         match (self.column_type.into(), other.column_type.into()) {
             // Indices are assigned within column types.
-            (Any::Advice, Any::Advice)
-            | (Any::Instance, Any::Instance)
+            (Any::Instance, Any::Instance)
+            | (Any::Advice, Any::Advice)
             | (Any::Fixed, Any::Fixed) => self.index.cmp(&other.index),
             // Across column types, sort Instance < Advice < Fixed.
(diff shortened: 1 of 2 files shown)


src/plonk/circuit.rs
```diff
@@ -47,13 +47,13 @@ impl<C: ColumnType> Ord for Column<C> {
             (Any::Advice, Any::Advice)
             | (Any::Instance, Any::Instance)
             | (Any::Fixed, Any::Fixed) => self.index.cmp(&other.index),
-            // Across column types, sort Advice < Instance < Fixed.
-            (Any::Advice, Any::Instance)
+            // Across column types, sort Instance < Advice < Fixed.
```
str4d, line 50: This does not have any practical effect on the V1 layouter, since instance columns never actually occur within regions, but it keeps the inherent ordering in line with the rendered ordering.

src/plonk/circuit.rs
```diff
@@ -47,13 +47,13 @@ impl<C: ColumnType> Ord for Column<C> {
             (Any::Advice, Any::Advice)
             | (Any::Instance, Any::Instance)
```
daira, line 48:
```suggestion
            (Any::Instance, Any::Instance)
            | (Any::Advice, Any::Advice)
```

daira (approved):
utACK with nonblocking suggestion.

daira (approved):
utACK c99347be4a7dfe667f3ef52cf6b7ecc5be61f6e3

ebfull (approved):

PR #347, opened 2021-07-21
Selector optimization

ebfull:


src/circuit/layouter.rs
```diff
@@ -110,10 +110,49 @@ pub trait RegionLayouter<F: Field>: fmt::Debug {
 #[derive(Clone, Debug)]
 pub struct RegionShape {
     pub(super) region_index: RegionIndex,
-    pub(super) columns: HashSet<Column<Any>>,
+    pub(super) columns: HashSet<RegionColumn>,
     pub(super) row_count: usize,
 }
 
+/// The virtual column involved in a region. This includes normal "logical"
+/// columns as well as selectors that are not definite columns at this stage.
```
daira, line 118:
```suggestion
/// The virtual column involved in a region. This includes concrete columns,
/// as well as selectors that are not concrete columns at this stage.
```

src/circuit/layouter.rs
```diff
@@ -110,10 +110,49 @@ pub trait RegionLayouter<F: Field>: fmt::Debug {
...
+#[derive(Eq, PartialEq, Copy, Clone, Debug, Hash)]
+pub enum RegionColumn {
+    /// Logical column
```
daira, line 121:
```suggestion
    /// Concrete column
```

src/dev.rs
```diff
@@ -352,12 +354,7 @@ impl<F: Field + Group> Assignment<F> for MockProver<F> {
         self.regions.push(self.current_region.take().unwrap());
     }
 
-    fn enable_selector<A, AR>(
-        &mut self,
-        annotation: A,
-        selector: &Selector,
-        row: usize,
-    ) -> Result<(), Error>
+    fn enable_selector<A, AR>(&mut self, _: A, selector: &Selector, row: usize) -> Result<(), Error>
```
daira, line 357: If we ignore the annotation here even in dev, we should probably just remove it.

ebfull: I suspect it'll come in handy later, but removing the parameter would be disruptive right now. I'd rather wait to remove until we're sure it's useless, because if we remove now then downstream callers have to remove the annotations only to potentially put them back if it turns out they're useful. Out of scope for this PR I think.

src/dev.rs
```diff
@@ -518,6 +516,7 @@ impl<F: FieldExt> MockProver<F> {
 
         // Fixed columns contain no blinding factors.
         let fixed = vec![vec![CellValue::Unassigned; n]; cs.num_fixed_columns];
+        let selectors = vec![vec![false; n]; cs.num_selectors];
```
daira, line 519: Consider using [bitvec](https://crates.io/crates/bitvec) for the inner vectors.

src/plonk/circuit.rs
```diff
@@ -228,13 +228,19 @@ impl TryFrom<Column<Any>> for Column<Instance> {
 /// }
 /// ```
 #[derive(Clone, Copy, Debug, PartialEq, Eq, Hash)]
-pub struct Selector(pub(crate) Column<Fixed>);
+pub struct Selector(pub(crate) usize, bool);
```
daira, line 231: Consider naming the fields.

src/plonk/circuit.rs
```diff
@@ -679,6 +695,46 @@ impl<F: Field> Expression<F> {
     pub fn square(self) -> Self {
         self.clone() * self
     }
+
+    /// Returns whether or not this expression contains a `Selector`.
```
daira, line 699:
```suggestion
    /// Returns whether or not this expression contains a simple `Selector`.
```

src/plonk/circuit.rs
```diff
@@ -688,28 +744,37 @@ impl<F: Field> Neg for Expression<F> {
...
+        }
         Expression::Sum(Box::new(self), Box::new(rhs))
     }
 }
 
 impl<F: Field> Sub for Expression<F> {
     type Output = Expression<F>;
     fn sub(self, rhs: Expression<F>) -> Expression<F> {
+        if self.contains_simple_selector() || rhs.contains_simple_selector() {
+            panic!("attempted to add to a simple selector");
```
daira, line 761:
```suggestion
            panic!("attempted to use a simple selector in a subtraction");
```

src/plonk/circuit.rs
```diff
@@ -688,28 +744,37 @@ impl<F: Field> Neg for Expression<F> {
     }
 }
 
-impl<F> Add for Expression<F> {
+impl<F: Field> Add for Expression<F> {
     type Output = Expression<F>;
     fn add(self, rhs: Expression<F>) -> Expression<F> {
+        if self.contains_simple_selector() || rhs.contains_simple_selector() {
+            panic!("attempted to add to a simple selector");
```
daira, line 751:
```suggestion
            panic!("attempted to use a simple selector in an addition");
```

src/plonk/circuit.rs
```diff
@@ -688,28 +744,37 @@ impl<F: Field> Neg for Expression<F> {
...
         Expression::Sum(Box::new(self), Box::new(-rhs))
     }
 }
 
-impl<F> Mul for Expression<F> {
+impl<F: Field> Mul for Expression<F> {
     type Output = Expression<F>;
     fn mul(self, rhs: Expression<F>) -> Expression<F> {
+        if self.contains_simple_selector() && rhs.contains_simple_selector() {
+            panic!("attempted to multiply two simple selectors");
```
daira, line 771:
```suggestion
            panic!("attempted to multiply two expressions containing simple selectors");
```

src/plonk/circuit.rs
```diff
@@ -679,6 +695,46 @@ impl<F: Field> Expression<F> {
     pub fn square(self) -> Self {
         self.clone() * self
     }
+
+    /// Returns whether or not this expression contains a `Selector`.
+    fn contains_simple_selector(&self) -> bool {
```
daira, line 700: Potential performance issue: since this is called on each subexpression when adding, subtracting or multiplying expressions, the cost of constructing an expression is now quadratic in the number of nodes in its AST.

src/plonk/circuit.rs
```diff
@@ -1072,11 +1156,280 @@ impl<F: Field> ConstraintSystem<F> {
...
+    ///
+    /// Do not call this twice. Yes, this should be a builder pattern instead.
+    pub(crate) fn compress_selectors(mut self, selectors: Vec<Vec<bool>>) -> (Self, Vec<Vec<F>>) {
+        if selectors.is_empty() {
+            return (self, vec![]);
+        }
+        // The length of all provided selectors must be the same.
+        let n = selectors[0].len();
+        assert!(selectors.iter().all(|a| a.len() == n));
+        // and the number of provided selectors must be the same.
```
daira, line 1173:
```suggestion
        // and the number of provided selectors must be the number we counted
        // for this ConstraintSystem.
```

src/plonk/circuit.rs
```diff
@@ -1072,11 +1156,280 @@ impl<F: Field> ConstraintSystem<F> {
...
+        let mut degrees = vec![0; selectors.len()];
+        for expr in self.gates.iter().flat_map(|gate| gate.polys.iter()) {
+            let degree = expr.degree();
+            if let Some(selector) = expr.extract_simple_selector() {
+                degrees[selector.0] = max(degrees[selector.0], degree);
+            }
+        }
+
+        // Any elements of `degrees` which still has zero must either be a
+        // simple selector or appear in no columns, so we don't need to compute
```
daira, line 1188:
```suggestion
        // Any element of `degrees` which still has zero must either be a
        // complex selector or appear in no columns, so we don't need to compute
```

src/plonk/circuit.rs
```diff
@@ -1072,11 +1156,280 @@ impl<F: Field> ConstraintSystem<F> {
...
+        // Compute the maximal degree of every simple selector. We only consider
+        // the expressions in gates, as lookup arguments cannot support simple
+        // selectors.
+        let mut degrees = vec![0; selectors.len()];
+        for expr in self.gates.iter().flat_map(|gate| gate.polys.iter()) {
+            let degree = expr.degree();
+            if let Some(selector) = expr.extract_simple_selector() {
+                degrees[selector.0] = max(degrees[selector.0], degree);
+            }
+        }
```
daira, line 1185: At this point we should stable-sort the selectors by decreasing degree.
> Rationale: The number of selectors in a combination is limited by the maximum degree of any polynomial constraint for those selectors. The rest of the algorithm will attempt to combine selectors by iterating over them in order. If this order is such that we prefer to combine selectors for which the maximum degree is about the same, intuitively we should be able to combine more of them. In practice this leads to slightly better results, i.e. fewer combinations are typically needed.

src/plonk/circuit.rs
```diff
@@ -1072,11 +1156,280 @@ impl<F: Field> ConstraintSystem<F> {
...
+
+        // Any elements of `degrees` which still has zero must either be a
+        // simple selector or appear in no columns, so we don't need to compute
+        // anything for those. They will appear on their own.
+        let complex_selectors = degrees.iter().map(|&i| i == 0).collect::<Vec<_>>();
+
+        // Compute the exclusion matrix that has (j, k) = true if selector j and
+        // selector k conflict -- that is, they are both enabled on the same
+        // row. This matrix is symmetric so we only need to store the lower
+        // diagonal.
```
daira, line 1195:
```suggestion
        // row. This matrix is symmetric and the diagonal entries are false, so
        // we only need to store the lower triangular entries.
```

src/plonk/circuit.rs
ebfull: Good point. I think if this gets out of hand we can address it, but note that none of these operations with `Expression` happen outside keygen (or tests) in practice.

src/plonk/circuit.rs
```diff
@@ -1072,11 +1156,280 @@ impl<F: Field> ConstraintSystem<F> {
...
+                    }
+
+                    // Look at what selectors are active at the same row
+                    if other_selector[set_index] {
+                        // Mark them as incompatible
+                        exclusion_matrix[i][j] = true;
+                    }
+                }
+            }
+        }
```
daira, line 1223: Let n be the number of rows.
Let s be the number of selector columns (assume for now there are no complex selectors).
Let q[r] be the number of enabled selectors on row r.

This algorithm costs O(n.s + sum_r {q[r].s}) time, whereas the one in the Python prototype takes O(n.s + sum_r {q[r]^2}) time, where q[r] will typically be smaller than s. Here's the Python code adapted for column-major order: https://gist.github.com/daira/7c9a2a372a89606ce1d2207e7e46ba79

src/plonk/circuit.rs
```diff
@@ -1072,11 +1156,280 @@ impl<F: Field> ConstraintSystem<F> {
...
+                seen[i] = true; // We're definitely adding this to a combination.
+                let mut combination = vec![i];
+                assert!(degrees[i] <= max_degree); // Otherwise, self.degree() is wrong.
+
+                // Keep track of how large the substitution is.
+                let mut usage = 1;
+
+                // Keep track of how large of a gate any of the members of this
+                // combination can support.
+                let mut room = max_degree - degrees[i];
```
daira, line 1263: The Python version has less code here because it isn't treating the first selector added to a combination as a special case (i.e. it's adding it in the loop over j). I don't see why it needs to be a special case?

src/plonk/circuit.rs
```diff
@@ -1072,11 +1156,280 @@ impl<F: Field> ConstraintSystem<F> {
...
+                        continue;
+                    }
+
+                    // Skip selectors that have been added to previous combinations
+                    if seen[j] {
+                        continue;
+                    }
+
+                    // Are the two selectors excluded from co-existing?
+                    // j > i and so (j, i) is always on the lower diagonal
```
daira, line 1284:
```suggestion
                    // j > i and so (j, i) is always in the lower triangle.
```

src/plonk/circuit.rs
```diff
@@ -1072,11 +1156,280 @@ impl<F: Field> ConstraintSystem<F> {
...
+                    // Can the new selector join the combination?
+                    if degrees[j] + usage <= max_degree {
+                        // Yes, it can. Let's do it.
+                        combination.push(j);
+                        // The remaining room available is the lesser of the
+                        // current remaining room available for previous
+                        // selector(s) after adjusting for adding this new
+                        // selector, and the room available in the new selector
+                        // after adjusting for adding the previous selectors.
+                        room = std::cmp::min(room - 1, max_degree - (degrees[j] + usage));
```
daira, line 1298: This is correct but I think it's a complicated way of doing it. It's simpler to keep track of the maximum degree of this combination so far, and then instead of shortcutting on `room == 0`, you can shortcut on `degree_of_combination == max_degree`.

src/plonk/circuit.rs
```diff
@@ -1072,11 +1156,280 @@ impl<F: Field> ConstraintSystem<F> {
...
+                    // This simple selector was added to a previous combination.
+                    continue;
+                }
+                selector_map[i] = combinations.len();
+                seen[i] = true; // We're definitely adding this to a combination.
+                let mut combination = vec![i];
+                assert!(degrees[i] <= max_degree); // Otherwise, self.degree() is wrong.
+
+                // Keep track of how large the substitution is.
+                let mut usage = 1;
```
daira, line 1259: This is always equal to `combination.len()`. Since taking the length of a `Vec` is efficient, there's no need for the extra variable.

src/plonk/circuit.rs
```diff
@@ -1072,11 +1156,280 @@ impl<F: Field> ConstraintSystem<F> {
...
+            let column = self.fixed_column();
+            new_columns.push(Some(column));
+            let query = Expression::Fixed {
+                query_index: self.query_fixed_index(column, Rotation::cur()),
+                column_index: column.index,
+                rotation: Rotation::cur(),
+            };
+
+            // Every virtual selector gets a different integer (starting at 1)
+            // that it assigns when it wants to be active.
```
daira, line 1323:
```suggestion
            // Every virtual selector in this combination gets a different integer
            // (starting at 1) that is assigned when it is enabled.
```
(You *could* assign disjoint integers across all combinations. I guess that might have minor advantages for debugging, but it would be very slightly less efficient, due to larger scalars in the multiexp.)

src/plonk/circuit.rs
```diff
@@ -1149,13 +1161,23 @@ impl<F: Field> ConstraintSystem<F> {
...
         ) {
             *expr = expr.evaluate(
                 &|constant| Expression::Constant(constant),
-                &|selector| Expression::Fixed {
-                    query_index: selector_queries[selector.0],
-                    column_index: selector_map[selector.0].index(),
-                    rotation: Rotation::cur(),
+                &|selector| {
+                    if must_be_nonsimple {
+                        // Complex selectors are prohibited from appearing in
```
str4d, line 1369:
````suggestion
                        // Simple selectors are prohibited from appearing in
```

src/plonk/circuit.rs
```diff
@@ -1072,11 +1156,280 @@ impl<F: Field> ConstraintSystem<F> {
...
+                }
+                // Construct a polynomial that has a root at 0 and at integers
+                // not including what this selector has been assigned to. This
+                // is a generalization of a boolean selector when the selector
+                // is multiplying a value that is constrained to equal zero.
+                let mut expr = query.clone();
+                let mut root = F::one();
+                for _ in 0..combination.len() {
+                    if root != assignment {
+                        expr = expr * (query.clone() - Expression::Constant(root));
```
daira, line 1339:
```suggestion
                        expr = expr * (Expression::Constant(root) - query.clone());
```
per https://github.com/zcash/orchard/issues/143.

str4d (changes requested):
Reviewed up to e23d148b72ebc8d1a37f10b72a80a8adf2a0711e (all except the last commit that implements the optimization algorithm).

src/plonk/circuit.rs
```diff
@@ -1072,11 +1156,280 @@ impl<F: Field> ConstraintSystem<F> {
...
+            replace_selectors(expr, &selector_replacements, false);
+        }
+
+        // Substitute non-simple selectors for the real fixed columns in all
+        // lookup expressions
+        for expr in self.lookups.iter_mut().flat_map(|lookup| {
+            lookup
+                .input_expressions
+                .iter_mut()
+                .chain(lookup.table_expressions.iter_mut())
```
daira, line 1409: At https://github.com/zcash/halo2/pull/347/files#diff-c7d3f048cff35b9298e74eb82f24b7755a72927c699f387552c0ac97f7c6c4a8R993 we required tables to be tuples of `Column<Fixed>`, so it shouldn't be necessary to substitute table expressions. (But it looks as though it should work anyway, so what was the rationale for the `Column<Fixed>` restriction?)

src/plonk/lookup/prover.rs
```diff
@@ -134,6 +135,7 @@ impl<F: FieldExt> Argument<F> {
                 .map(|expression| {
                     expression.evaluate(
                         &|scalar| pk.vk.domain.constant_extended(scalar),
+                        &|_| panic!("virtual selectors are optimized away during keygen"),
```
daira, line 138:
```suggestion
                        &|_| panic!("virtual selectors are removed during optimization"),
```

src/plonk/lookup/verifier.rs
```diff
@@ -134,6 +134,7 @@ impl<C: CurveAffine> Evaluated<C> {
                     .map(|expression| {
                         expression.evaluate(
                             &|scalar| scalar,
+                            &|_| panic!("virtual selectors are optimized away during keygen"),
```
daira, line 137:
```suggestion
                            &|_| panic!("virtual selectors are removed during optimization"),
```

daira (approved):
utACK with comments. I think it would be preferable to do the sorting by degree now.

src/plonk/circuit.rs
daira: Oh, maybe it is slightly more efficient, because when you get to each combination you're not looping over selectors that have definitely already been seen. Yeah that's fine.

src/plonk/circuit.rs
ebfull: I wanted to prevent selectors from appearing in the lookup tables. The quick and dirty solution is to just restrict it to `Column<Fixed>` on the forward facing API (the `lookup` method on `ConstraintSystem`) and not have to go through the entire lookup argument and modify it to special case the tables. This way we can reintroduce expressions for tables if we need them someday.

src/plonk/circuit.rs
```diff
@@ -1072,11 +1156,282 @@ impl<F: Field> ConstraintSystem<F> {
...
+                    }
+
+                    // Are the two selectors excluded from co-existing?
+                    // j > i and so (j, i) is always on the lower diagonal
+                    if exclusion_matrix[j][i] {
+                        continue;
+                    }
+
+                    // Can the new selector join the combination?
+                    if degrees[j] + usage <= max_degree {
```
daira, line 1291: This has an off-by-one error: usage hasn't yet been incremented, and so the actual maximum degree of any gate involving selector j is `degrees[j] + usage + 1`.

(The `room` check above accounted for the case where the degree bound would be violated by adding another term to a gate *not* involving selector j.)

src/plonk/circuit.rs
```diff
@@ -1072,11 +1156,282 @@ impl<F: Field> ConstraintSystem<F> {
...
+            expr: &mut Expression<F>,
+            selector_map: &[Expression<F>],
+            must_be_nonsimple: bool,
+        ) {
+            *expr = expr.evaluate(
+                &|constant| Expression::Constant(constant),
+                &|selector| {
+                    if must_be_nonsimple {
+                        // Simple selectors are prohibited from appearing in
+```
```
daira, line 1371:
```suggestion
```

daira:
I don't know what's going on with the compilation failure in CI (src/circuit/floor_planner/single_pass.rs:122:73). It works for me with
```
cargo +1.51.0 check --all-targets --all-features --examples
```

ebfull:
> I don't know what's going on with the compilation failure in CI (src/circuit/floor_planner/single_pass.rs:122:73). It works for me with
> 
> ```
> cargo +1.51.0 check --all-targets --all-features --examples
> ```

Yeah, I don't understand either. Something is borked with the CI. It's not testing what's actually on this branch.

src/plonk/circuit.rs
ebfull: Added some notes about this in various places.

src/plonk/circuit.rs
ebfull: This is gone now.

src/plonk/circuit.rs
ebfull: Yeah, I switched to the way you do it in your python code.

src/plonk/circuit.rs
ebfull: I was unable to find an example where sorting helped.

src/plonk/circuit/compress_selectors.rs
```diff
@@ -0,0 +1,341 @@
...
+                    // Every selector should be assigned to a combination
+                    assert!(selector.combination_index < combination_assignments.len());
+                    assert!(!selectors_seen.contains(&selector.selector));
+                    selectors_seen.push(selector.selector);
+                }
+            }
+
+            // Test that, for each selector, the provided expression
+            //  1. evaluates to zero on rows where the selector's activation is off
+            //  2. evaluates to nonzero on rows where the selector's activation is off
```
ebfull: Oops, I meant to say "on" here.

src/plonk/circuit/compress_selectors.rs
```diff
@@ -0,0 +1,341 @@
...
+            //  1. evaluates to zero on rows where the selector's activation is off
+            //  2. evaluates to nonzero on rows where the selector's activation is on
+            //  3. is of degree d such that d + (selector.max_degree - 1) <= max_degree
+            //     OR selector.max_degree is zero
+            for selector in selector_assignments {
+                assert_eq!(
+                    selectors[selector.selector].activations.len(),
+                    combination_assignments[selector.combination_index].len()
+                );
+                for (&activation, &assignment) in selectors[selector.selector].activations.iter().zip(combination_assignments[selector.combination_index].iter()) {
```
str4d, line 308: Formatting:
```suggestion
                for (&activation, &assignment) in selectors[selector.selector]
                    .activations
                    .iter()
                    .zip(combination_assignments[selector.combination_index].iter())
                {
```

src/dev/graph/layout.rs
```diff
@@ -83,28 +84,36 @@ impl CircuitLayout {
...
+        let mut layout = Layout::new(n, cs.num_selectors);
         ConcreteCircuit::FloorPlanner::synthesize(
             &mut layout,
             circuit,
             config,
             cs.constants.clone(),
         )
         .unwrap();
+        let (cs, selector_polys) = cs.compress_selectors(layout.selectors);
+        let non_selector_fixed_columns = cs.num_fixed_columns - selector_polys.len();
```
str4d, line 107: The selector columns of the layout diagram are actually rather broken after this change (due to how regions end up overlapping), but I'll fix that in a separate PR.

str4d (changes requested):
One bug to fix, utACK otherwise with suggestions.

str4d (approved):
utACK

PR #350, opened 2021-07-23
Fix SimpleFloorPlanner's handling of global constants

str4d:
The previous behaviour assumed that all regions were always assigned
in increasing start-row order. That assumption is broken in e.g. the
Merkle path tests, which position two Sinsemilla calls in parallel.
The layouter was working fine for current tests, but in an adjacent
PR that changed the Merkle path chip's behaviour, it started to
produce invalid circuits.

The new behaviour emulates the `floor_planner::V1` behaviour in an
(intentionally) inefficient way: constants are positioned immediately
after the regions they are assigned to, in the first constants column.
If that column is also used in regions, those regions will now be
positioned without overlapping any assigned global constants.

Co-authored-by: therealyingtong <yingtong@z.cash>


ebfull (approved):

daira (approved):
utACK

PR #351, opened 2021-07-24
Sort selectors by decreasing degree before applying the combining optimization

daira:
Rationale: The number of selectors in a combination is limited by the maximum degree of any polynomial constraint for those selectors. The rest of the algorithm will attempt to combine selectors by iterating over them in order. If this order is such that we prefer to combine selectors for which the maximum degree is about the same, intuitively we should be able to combine more of them. In practice this leads to slightly better results on average (although it happens not to help for the Action circuit).

### DIFF
--- a/src/plonk/circuit/compress_selectors.rs
+++ b/src/plonk/circuit/compress_selectors.rs
@@ -49,7 +49,7 @@ pub struct SelectorAssignment<F> {
 ///
 /// This function is completely deterministic.
 pub fn process<F: Field, E>(
-    mut selectors: Vec<SelectorDescription>,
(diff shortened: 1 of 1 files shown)


daira:
I suggest reviewing with "Hide whitespace changes".

daira:
This has now been updated for the current selector-optimization branch (which makes the changes much simpler) and tested with the Action circuit. By coincidence of the distribution of degrees and the layout of the Action circuit, the sorting does not reduce the number of combinations (16). However, I have verified that it is doing the right thing, and it might help for other circuits.

daira:
I miscounted; it actually increases the number of combinations to 17. Closing.

PR #352, opened 2021-07-25
`Expression::Negated`

str4d:
Previously, we were storing negations as "scaled by -1", which meant we needed to evaluate them with multiplications instead of cheaper negations or additions.


ebfull (approved):

daira (approved):
utACK

PR #353, opened 2021-07-26
Developer tooling improvements

str4d:
- `CircuitGates` tooling for printing circuit gate details
- `CircuitCost::marginal_proof_size` API
- Improvements to `CircuitLayout`

### DIFF
--- a/src/dev/gates.rs
+++ b/src/dev/gates.rs
@@ -1,0 +1,227 @@
+use std::fmt;
+
+use ff::PrimeField;
+
+use crate::plonk::{Circuit, ConstraintSystem};
+
+#[derive(Debug)]
+struct Constraint {
+    name: &'static str,
+    expression: String,
+}
+
+#[derive(Debug)]
+struct Gate {
+    name: &'static str,
+    constraints: Vec<Constraint>,
+}
+
+/// A struct for collecting and displaying the gates within a circuit.
+///
+/// # Examples
+///
+/// ```
+/// use ff::Field;
+/// use halo2::{
+///     circuit::{Layouter, SimpleFloorPlanner},
+///     dev::CircuitGates,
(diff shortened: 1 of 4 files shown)


ebfull (approved):

src/dev/gates.rs
```diff
@@ -0,0 +1,227 @@
...
+        let mut cs = ConstraintSystem::default();
+        let _ = C::configure(&mut cs);
+
+        let format_scalar = |s: F| {
+            // Format scalar as hex.
+            let s = format!("{:?}", s);
+            // Remove leading zeroes.
+            let s = s.strip_prefix("0x").unwrap();
+            let s = s.trim_start_matches('0');
+            format!("0x{}", s)
```
daira, line 107: This formats 0 as just "0x".

str4d: We never reach here with `0`; that constant is handled explicitly below.

str4d: Hmm, actually we'd reach here if we multiplied an expression by the scalar 0, so we should fix this. But that seems more indicative of a circuit bug.

src/dev/gates.rs
```diff
@@ -0,0 +1,227 @@
...
+                                } else {
+                                    format_scalar(s)
+                                }
+                            },
+                            &|selector| format!("S{}", selector.0),
+                            &|_, column, rotation| format!("F{}@{}", column, rotation.0),
+                            &|_, column, rotation| format!("A{}@{}", column, rotation.0),
+                            &|_, column, rotation| format!("I{}@{}", column, rotation.0),
+                            &|a| {
+                                if a.contains(' ') {
```
daira, line 138: That's clever :-)

daira (commented):
utACK

PR #338, opened 2021-07-19
Improved error handling

str4d:

PR #354, opened 2021-07-27
Add a `Layouter::assign_table` API

daira:
The new API correctly handles lookup table assignments; in particular, layouters now ensure that every column used in a table has a valid table entry on every usable row.

### DIFF
--- a/src/circuit.rs
+++ b/src/circuit.rs
@@ -6,7 +6,7 @@ use ff::Field;
 
 use crate::{
     arithmetic::FieldExt,
-    plonk::{Advice, Any, Assigned, Column, Error, Fixed, Instance, Selector},
+    plonk::{Advice, Any, Assigned, Column, Error, Fixed, Instance, Selector, TableColumn},
 };
 
 pub mod floor_planner;
@@ -245,9 +245,6 @@ impl<'r, F: Field> Region<'r, F> {
 }
 
 /// A lookup table in the circuit.
-///
-/// TODO: Add wrapper type to `ConstraintSystem` to allow us to only deal with lookup
-/// columns here.
 #[derive(Debug)]
 pub struct Table<'r, F: Field> {
     table: &'r mut dyn layouter::TableLayouter<F>,
@@ -266,7 +263,7 @@ impl<'r, F: Field> Table<'r, F> {
     pub fn assign_fixed<'v, V, VR, A, AR>(
         &'v mut self,
         annotation: A,
-        column: Column<Fixed>,
+        column: TableColumn,
         offset: usize,
         mut to: V,
     ) -> Result<(), Error>
--- a/src/circuit.rs
+++ b/src/circuit.rs
@@ -257,10 +257,12 @@ impl<'r, F: Field> From<&'r mut dyn layouter::TableLayouter<F>> for Table<'r, F>
 }
 
 impl<'r, F: Field> Table<'r, F> {
-    /// Assign a fixed value.
+    /// Assigns a fixed value to a table cell.
+    ///
+    /// Returns an error if the table cell has already been assigned to.
     ///
     /// Even though `to` has `FnMut` bounds, it is guaranteed to be called at most once.
-    pub fn assign_fixed<'v, V, VR, A, AR>(
+    pub fn assign_cell<'v, V, VR, A, AR>(
         &'v mut self,
         annotation: A,
         column: TableColumn,
@@ -274,7 +276,7 @@ impl<'r, F: Field> Table<'r, F> {
         AR: Into<String>,
     {
         self.table
-            .assign_fixed(&|| annotation().into(), column, offset, &mut || {
+            .assign_cell(&|| annotation().into(), column, offset, &mut || {
                 to().map(|v| v.into())
             })
     }
--- a/src/circuit.rs
+++ b/src/circuit.rs
@@ -244,6 +244,45 @@ impl<'r, F: Field> Region<'r, F> {
     }
 }
 
+/// A lookup table in the circuit.
+///
+/// TODO: Add wrapper type to `ConstraintSystem` to allow us to only deal with lookup
+/// columns here.
+#[derive(Debug)]
+pub struct Table<'r, F: Field> {
+    table: &'r mut dyn layouter::TableLayouter<F>,
+}
+
+impl<'r, F: Field> From<&'r mut dyn layouter::TableLayouter<F>> for Table<'r, F> {
+    fn from(table: &'r mut dyn layouter::TableLayouter<F>) -> Self {
+        Table { table }
+    }
+}
+
+impl<'r, F: Field> Table<'r, F> {
+    /// Assign a fixed value.
+    ///
+    /// Even though `to` has `FnMut` bounds, it is guaranteed to be called at most once.
+    pub fn assign_fixed<'v, V, VR, A, AR>(
+        &'v mut self,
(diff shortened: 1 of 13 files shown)


src/circuit.rs
```diff
@@ -244,6 +244,45 @@ impl<'r, F: Field> Region<'r, F> {
     }
 }
 
+/// A lookup table in the circuit.
+///
+/// TODO: Add wrapper type to `ConstraintSystem` to allow us to only deal with lookup
+/// columns here.
```
str4d, line 250: This is a safety concern: without this type separation, chip developers can ignore `Layouter::assign_table` and continue to use `Layouter::assign_region` which does not perform default-value-filling. We don't need to implement this before we merge the Action circuit PR for Orchard, but we _should_ do this before we start encouraging others to use the `halo2` crate.

src/circuit/floor_planner/single_pass.rs
```diff
@@ -314,6 +370,74 @@ impl<'r, 'a, F: Field, CS: Assignment<F> + 'a> RegionLayouter<F>
...
+            || {
+                let res = to();
+                value = res.as_ref().ok().cloned();
+                res
+            },
+        )?;
+
+        match (entry.0.is_none(), offset) {
+            (true, 0) => entry.0 = Some(value),
+            (false, 0) => return Err(Error::SynthesisError), // TODO better error
```
therealyingtong, line 429:
```suggestion
            // Use the value at offset 0 as the default value for this table column.
            (true, 0) => entry.0 = Some(value),
            // Since there is already an existing default value for this table column,
            // the caller should not be attempting to assign another value at offset 0.
            (false, 0) => return Err(Error::SynthesisError), // TODO better error
```

src/circuit.rs
str4d: I added a `TableColumn` struct to address this.

src/circuit/floor_planner/single_pass.rs
```diff
@@ -314,6 +370,73 @@ impl<'r, 'a, F: Field, CS: Assignment<F> + 'a> RegionLayouter<F>
     }
 }
 
+pub(crate) struct SimpleTableLayouter<'r, 'a, F: Field, CS: Assignment<F> + 'a> {
+    cs: &'a mut CS,
+    used_columns: &'r [TableColumn],
+    // maps from a fixed column to a pair (default value, vector saying which rows are assigned)
+    pub(crate) default_and_assigned: HashMap<TableColumn, (Option<Option<Assigned<F>>>, Vec<bool>)>,
```
therealyingtong, line 377: Note to self: understand why there is a double `Option<Option<Assigned<F>>>`. e.g. could we have a default value be `Some(None)`?

therealyingtong (approved):
utACK

src/circuit/floor_planner/single_pass.rs
```diff
@@ -314,6 +370,82 @@ impl<'r, 'a, F: Field, CS: Assignment<F> + 'a> RegionLayouter<F>
...
+    cs: &'a mut CS,
+    used_columns: &'r [TableColumn],
+    // maps from a fixed column to a pair (default value, vector saying which rows are assigned)
+    pub(crate) default_and_assigned: HashMap<TableColumn, (DefaultTableValue<F>, Vec<bool>)>,
+}
+
+impl<'r, 'a, F: Field, CS: Assignment<F> + 'a> fmt::Debug for SimpleTableLayouter<'r, 'a, F, CS> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("SimpleTableLayouter")
+            .field("default_and_assigned", &self.default_and_assigned)
```
daira, line 392:
```suggestion
            .field("used_columns", &self.used_columns)
            .field("default_and_assigned", &self.default_and_assigned)
```

src/circuit/floor_planner/single_pass.rs
```diff
@@ -143,6 +146,59 @@ impl<'a, F: Field, CS: Assignment<F> + 'a> Layouter<F> for SingleChipLayouter<'a
...
+                    } else {
+                        None
+                    }
+                })
+                .reduce(|acc, item| match (acc, item) {
+                    (Some(a), Some(b)) if a == b => Some(a),
+                    _ => None,
+                }) {
+                Some(Some(len)) => len,
+                _ => return Err(Error::SynthesisError), // TODO better error
```
daira, line 182: We need something like a `CircuitConstructionError`, for errors in API usage when constructing a circuit.

src/circuit/floor_planner/v1.rs
```diff
@@ -270,6 +286,60 @@ impl<'p, 'a, F: Field, CS: Assignment<F> + 'a> AssignmentPass<'p, 'a, F, CS> {
         Ok(result)
     }
 
+    fn assign_table<A, AR, N, NR>(&mut self, name: N, mut assignment: A) -> Result<AR, Error>
+    where
+        A: FnMut(Table<'_, F>) -> Result<AR, Error>,
+        N: Fn() -> NR,
+        NR: Into<String>,
+    {
```
daira, line 294:
```suggestion
    {
        // Maintenance hazard: there is near-duplicate code in `SingleChipLayouter::assign_table`.

```

src/circuit/floor_planner/single_pass.rs
```diff
@@ -143,6 +146,59 @@ impl<'a, F: Field, CS: Assignment<F> + 'a> Layouter<F> for SingleChipLayouter<'a
         Ok(result)
     }
 
+    fn assign_table<A, N, NR>(&mut self, name: N, mut assignment: A) -> Result<(), Error>
+    where
+        A: FnMut(Table<'_, F>) -> Result<(), Error>,
+        N: Fn() -> NR,
+        NR: Into<String>,
+    {
```
daira, line 154:
```suggestion
    {
        // Maintenance hazard: there is near-duplicate code in `v1::AssignmentPass::assign_table`.
```

daira (commented):
utACK @str4d's changes (e855ac6adbd292212b582e41d2fb543b89b67350 and 57877f5f180175e3473617e5154fa20678d23de3). Note that my review shouldn't count toward the review threshold for the first two commits.

src/circuit/floor_planner/single_pass.rs
str4d: I'll handle this in #338.

src/circuit/layouter.rs
```diff
@@ -105,6 +104,22 @@ pub trait RegionLayouter<F: Field>: fmt::Debug {
...
 }
 
+/// Helper trait for implementing a custom [`Layouter`].
+///
+/// This trait is used for implementing table assignments.
+///
+/// [`Layouter`]: super::Layouter
+pub trait TableLayouter<F: Field>: fmt::Debug {
+    /// Assign a fixed value
+    fn assign_fixed<'v>(
```
ebfull, line 114: I think we should rename this to `assign_table` or something, as it's restricted only to `TableColumn`.

ebfull (changes requested):

src/circuit/layouter.rs
str4d: Oh yeah, I was going to rename this to `assign_cell` in a follow-up commit (to keep the `TableColumn` commit simpler) and forgot.

ebfull (approved):

PR #356, opened 2021-07-31
book: Fix rendering of cost-model output.

parazyd:
Currently on the bottom of the dev-tools HTML page[0] we can see HTML
tags being rendered inside the code tag.

This commit should stop html tags appearing inside the `<code>` tags, and
allow for proper printing of the book.

[0] https://zcash.github.io/halo2/user/dev-tools.html


daira (approved):
ACK

PR #363, opened 2021-09-05
Notation refactoring

### LINKED ISSUE #170: ELI15: Generating sets
## What was unclear?

When discussing the generating sets, I think that the expression $\prod\limits_{i=1}^{k} g_i^{a_i}$ could be clearer. The exponent $a_i$ should be from $\mathbb{Z}$, but $a_i$ is previously used to denote a group element.

## What would help to make it clearer to you?

It might be useful to change $g_{1..k}$ to $g_{1..n}$ for example, and then $\prod\limits_{i=1}^{k} g_i^{a_i}$ to $\prod\limits_{i=1}^{n} g_i^{k_i}$ since $k$ was already used to denote exponents. I would also explicitly state that $k_i \in \mathbb{Z}$.

I hope I'm not wrong though.

upbqdn:
This PR closes #170.

I couldn't figure out why the CRT implies that a group with composite order has proper subgroups. I'd be glad if anyone can give me a hint why this holds. I looked into the cited material, but I didn't find such a claim there.

### DIFF
--- a/book/src/background/fields.md
+++ b/book/src/background/fields.md
@@ -86,9 +86,9 @@ notation). The order _of the group_ is the number of elements.
 
 Groups always have a [generating set], which is a set of elements such that we can produce
 any element of the group as (in multiplicative terminology) a product of powers of those
-elements. So if the generating set is $g_{1..k}$, we can produce any element of the group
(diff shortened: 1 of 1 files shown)


upbqdn:
My question is solved now. Thank you for the references @daira.

PR #374, opened 2021-09-30
Clean up our trait usage

str4d:
This removes our dependence on several `FieldExt` and `CurveAffine` methods, preparing us for their removal from `pasta_curves`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ and this project adheres to Rust's notion of
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Removed
+- `halo2::arithmetic::BatchInvert` (use `ff::BatchInvert` instead).
 
 ## [0.1.0-beta.1] - 2021-09-24
 Initial beta release!
--- a/src/plonk.rs
+++ b/src/plonk.rs
@@ -8,6 +8,7 @@
 use blake2b_simd::Params as Blake2bParams;
 
 use crate::arithmetic::{CurveAffine, FieldExt};
+use crate::helpers::CurveRead;
 use crate::poly::{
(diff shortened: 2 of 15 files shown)


CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ and this project adheres to Rust's notion of
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Removed
+- `halo2::arithmetic::BatchInvert` (use `ff::BatchInvert` instead).
```
str4d, line 10: Sigh, I thought I'd done this before the first beta (which depends on `ff 0.11` which includes the upstreamed trait).

src/transcript.rs
```diff
@@ -277,12 +278,18 @@ impl<C: CurveAffine> EncodedChallenge<C> for Challenge255<C> {
 
     fn new(challenge_input: &[u8; 64]) -> Self {
         Challenge255(
-            C::Scalar::from_bytes_wide(challenge_input).to_bytes(),
+            C::Scalar::from_bytes_wide(challenge_input)
+                .to_repr()
+                .as_ref()
+                .try_into()
+                .expect("Scalar fits into 256 bits"),
```
str4d, line 285: We need a better way to handle these challenges with curves that aren't necessarily 255 bits.

daira (approved):
utACK modulo fixing or suppressing clippy warnings.

nuttycom (approved):
utACK

src/plonk.rs
```diff
@@ -8,6 +8,7 @@
 use blake2b_simd::Params as Blake2bParams;
 
 use crate::arithmetic::{CurveAffine, FieldExt};
+use crate::helpers::CurveRead;
```
str4d, line 11: This lints as unused because `CurveAffine` also imports a `read` function. In a subsequent PR we'll update to a version of `pasta_curves` where `CurveAffine::read` is removed.

Issue #364, opened 2021-09-09
Debug: can MockProver.verify method output more cell value when poly constraint failed?


DreamWuGit:
hi,  Buddies
    Currently we use halo2 lib to write custom gadget, while found it is difficult to debug when constraint fails., comparing to bellman or dusk. because the concrete value can not be retrieve directly for failed expressions.    

Could you have good ways to trouble shooting or make  MockProver.verify output more useful info(values regarding expressions i.e. a - b =0 fails , then print a, b  )?

thanks !

daira:
Do you mean on proving, rather than verification? That information isn't available on verification.

DreamWuGit:
no, I mean use src/dev.rs module 's MockProver.verify() method to dev verification purpose, not generate real proof & verify.

str4d:
This wasn't trivial to do for `halo2` like it was for `bellman`, because you have different kinds of gates (and thus don't immediately know what cells are being used). But it can definitely be done, and we can contain the extra computational work to the failing cases.

str4d:
We can probably repurpose or extend the logic in https://github.com/zcash/halo2/blob/main/src/dev/gates.rs for this.

DreamWuGit:
@str4d  thanks for your reply , currently I add some code to output cells' value in certain row  locally. as the error message will print affected row .  this can help me to trouble shooting to some degree .  also can make pr if you think that make sense.

str4d:
@DreamWuGit could you try #376 and see if it provides what you need?

PR #376, opened 2021-10-01
dev: Collect values of queried cells for unsatisfied constraints

str4d:
This collects and provides values for non-selector queries.

Closes zcash/halo2#364.

### DIFF
--- a/src/dev.rs
+++ b/src/dev.rs
@@ -12,12 +12,13 @@ use crate::{
     arithmetic::{FieldExt, Group},
     plonk::{
         permutation, Advice, Any, Assignment, Circuit, Column, ColumnType, ConstraintSystem, Error,
-        Expression, Fixed, FloorPlanner, Instance, Selector,
+        Expression, Fixed, FloorPlanner, Instance, Selector, VirtualCell,
     },
     poly::Rotation,
 };
 
 pub mod metadata;
+mod util;
 
 pub mod cost;
 pub use cost::CircuitCost;
@@ -54,6 +55,8 @@ pub enum VerifyFailure {
         constraint: metadata::Constraint,
         /// The row on which this constraint is not satisfied.
         row: usize,
+        /// The values of the virtual cells used by this constraint.
+        cell_values: Vec<(metadata::VirtualCell, String)>,
     },
     /// A constraint was active on an unusable row, and is likely missing a selector.
     ConstraintPoisoned {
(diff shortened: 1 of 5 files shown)


src/dev.rs
```diff
@@ -640,44 +656,91 @@ impl<F: FieldExt> MockProver<F> {
...
+                                "-1".into()
+                            } else {
+                                // Format value as hex.
+                                let s = format!("{:?}", v);
+                                // Remove leading zeroes.
+                                let s = s.strip_prefix("0x").unwrap();
+                                let s = s.trim_start_matches('0');
+                                format!("0x{}", s)
                             }
+                        };
```
daira, line 714: Can this be moved to a utility function, rather than being inlined in this already-very-convoluted method?

src/dev.rs
```diff
@@ -689,15 +752,59 @@ impl<F: FieldExt> MockProver<F> {
...
+                                        &|mut a, mut b| {
+                                            a.append(&mut b);
+                                            a
+                                        },
+                                        &|mut a, mut b| {
+                                            a.append(&mut b);
+                                            a
+                                        },
+                                        &|a, _| a,
+                                    );
```
daira, line 795: Again this is a lot of code to be adding here.

src/poly.rs
```diff
@@ -285,7 +285,7 @@ impl<'a, F: Field, B: Basis> Mul<F> for Polynomial<F, B> {
 /// Describes the relative rotation of a vector. Negative numbers represent
 /// reverse (leftmost) rotations and positive numbers represent forward (rightmost)
 /// rotations. Zero represents no rotation.
-#[derive(Copy, Clone, Debug, PartialEq)]
+#[derive(Copy, Clone, Debug, PartialEq, Eq, PartialOrd, Ord)]
```
ebfull, line 288: I'm a _little_ worried about comparisons between `Rotation`s because for some circuits `Rotation(x)` is the same "logical" rotation as `Rotation(y)` even if `x != y`.

str4d: This is here because the dev tooling uses it in a `BTreeMap`, but we've established a pattern of having separate copies in `crate::dev::metadata`, so we could do the same here.

src/dev.rs
str4d: Extracted this as a utility function.

str4d:
Rebased on `main` and cleaned up the commit history.

therealyingtong (approved):
Tested ACK.

ebfull (approved):

Issue #288, opened 2021-06-04
Provide abstractions for typed cells


daira:
https://github.com/zcash/orchard/pull/89 implements a `CellValue` abstraction that represents a particular cell containing a field element. But it is error-prone to work with "untyped" field elements. In many cases, we want the circuit implementation to keep track of a high-level type that can be represented as a subset of field elements (for example, a bit, or a bit sequence of fixed length up to the capacity of the field). One possibility is to have a type like `CellValue` (or a trait like `Var`) that has the high-level type as a parameter, where that type is `TryFrom<F>` and `Into<F>` for field type `F`. Then,
* the cell-value returned by a copy constraint, for example, could automatically preserve the high-level type;
* gadget APIs could take typed cell-values as inputs and provide them as outputs.

str4d:
Closed by #383.

PR #383, opened 2021-10-11
circuit.rs: Introduce `AssignedCell` struct.

therealyingtong:
Addresses https://github.com/zcash/halo2/issues/288.

### DIFF
--- a/benches/plonk.rs
+++ b/benches/plonk.rs
@@ -110,7 +110,7 @@ fn bench_with_k(name: &str, k: u32, c: &mut Criterion) {
                     region.assign_fixed(|| "b", self.config.sb, 0, || Ok(FF::zero()))?;
                     region.assign_fixed(|| "c", self.config.sc, 0, || Ok(FF::one()))?;
                     region.assign_fixed(|| "a * b", self.config.sm, 0, || Ok(FF::one()))?;
-                    Ok((lhs, rhs, out))
+                    Ok((lhs.cell(), rhs.cell(), out.cell()))
                 },
             )
         }
@@ -152,7 +152,7 @@ fn bench_with_k(name: &str, k: u32, c: &mut Criterion) {
                     region.assign_fixed(|| "b", self.config.sb, 0, || Ok(FF::one()))?;
                     region.assign_fixed(|| "c", self.config.sc, 0, || Ok(FF::one()))?;
                     region.assign_fixed(|| "a * b", self.config.sm, 0, || Ok(FF::zero()))?;
-                    Ok((lhs, rhs, out))
+                    Ok((lhs.cell(), rhs.cell(), out.cell()))
                 },
             )
         }
--- a/examples/simple-example.rs
+++ b/examples/simple-example.rs
@@ -217,7 +217,7 @@ impl<F: FieldExt> NumericInstructions<F> for FieldChip<F> {
                 b.0.copy_advice(|| "rhs", &mut region, config.advice[1], 0)?;
 
                 // Now we can assign the multiplication result into the output position.
-                let value = a.0.value().and_then(|a| b.0.value().map(|b| a * b));
+                let value = a.0.value().and_then(|a| b.0.value().map(|b| *a * *b));
 
                 // Finally, we return a variable representing the output,
                 // to be used in another part of the circuit.
--- a/examples/simple-example.rs
+++ b/examples/simple-example.rs
@@ -4,7 +4,7 @@ use std::marker::PhantomData;
 
 use halo2::{
     arithmetic::FieldExt,
-    circuit::{Cell, Chip, Layouter, Region, SimpleFloorPlanner},
+    circuit::{AssignedCell, Chip, Layouter, Region, SimpleFloorPlanner},
     plonk::{Advice, Circuit, Column, ConstraintSystem, Error, Fixed, Instance, Selector},
     poly::Rotation,
 };
@@ -149,10 +149,7 @@ impl<F: FieldExt> Chip<F> for FieldChip<F> {
 // ANCHOR: instructions-impl
 /// A variable representing a number.
 #[derive(Clone)]
-struct Number<F: FieldExt> {
-    cell: Cell,
-    value: Option<F>,
-}
+struct Number<F: FieldExt>(AssignedCell<F, F>);
 
 impl<F: FieldExt> NumericInstructions<F> for FieldChip<F> {
     type Num = Number<F>;
@@ -164,21 +161,19 @@ impl<F: FieldExt> NumericInstructions<F> for FieldChip<F> {
     ) -> Result<Self::Num, Error> {
         let config = self.config();
 
-        let mut num = None;
         layouter.assign_region(
             || "load private",
             |mut region| {
-                let cell = region.assign_advice(
-                    || "private input",
-                    config.advice[0],
-                    0,
-                    || value.ok_or(Error::Synthesis),
-                )?;
-                num = Some(Number { cell, value });
-                Ok(())
+                region
+                    .assign_advice(
+                        || "private input",
+                        config.advice[0],
+                        0,
+                        || value.ok_or(Error::Synthesis),
+                    )
+                    .map(Number)
             },
-        )?;
-        Ok(num.unwrap())
+        )
     }
 
     fn load_constant(
@@ -188,24 +183,14 @@ impl<F: FieldExt> NumericInstructions<F> for FieldChip<F> {
     ) -> Result<Self::Num, Error> {
         let config = self.config();
 
-        let mut num = None;
         layouter.assign_region(
             || "load constant",
             |mut region| {
-                let cell = region.assign_advice_from_constant(
-                    || "constant value",
-                    config.advice[0],
-                    0,
-                    constant,
-                )?;
-                num = Some(Number {
-                    cell,
-                    value: Some(constant),
-                });
-                Ok(())
+                region
+                    .assign_advice_from_constant(|| "constant value", config.advice[0], 0, constant)
+                    .map(Number)
             },
-        )?;
-        Ok(num.unwrap())
+        )
     }
 
     fn mul(
@@ -216,7 +201,6 @@ impl<F: FieldExt> NumericInstructions<F> for FieldChip<F> {
     ) -> Result<Self::Num, Error> {
         let config = self.config();
 
-        let mut out = None;
         layouter.assign_region(
             || "mul",
             |mut region: Region<'_, F>| {
@@ -229,38 +213,24 @@ impl<F: FieldExt> NumericInstructions<F> for FieldChip<F> {
                 // but we can only rely on relative offsets inside this region. So we
                 // assign new cells inside the region and constrain them to have the
                 // same values as the inputs.
-                let lhs = region.assign_advice(
-                    || "lhs",
-                    config.advice[0],
-                    0,
-                    || a.value.ok_or(Error::Synthesis),
-                )?;
-                let rhs = region.assign_advice(
-                    || "rhs",
-                    config.advice[1],
-                    0,
-                    || b.value.ok_or(Error::Synthesis),
-                )?;
-                region.constrain_equal(a.cell, lhs)?;
-                region.constrain_equal(b.cell, rhs)?;
+                a.0.copy_advice(|| "lhs", &mut region, config.advice[0], 0)?;
+                b.0.copy_advice(|| "rhs", &mut region, config.advice[1], 0)?;
 
                 // Now we can assign the multiplication result into the output position.
-                let value = a.value.and_then(|a| b.value.map(|b| a * b));
-                let cell = region.assign_advice(
-                    || "lhs * rhs",
-                    config.advice[0],
-                    1,
-                    || value.ok_or(Error::Synthesis),
-                )?;
+                let value = a.0.value().and_then(|a| b.0.value().map(|b| a * b));
 
                 // Finally, we return a variable representing the output,
                 // to be used in another part of the circuit.
-                out = Some(Number { cell, value });
-                Ok(())
+                region
+                    .assign_advice(
+                        || "lhs * rhs",
+                        config.advice[0],
+                        1,
+                        || value.ok_or(Error::Synthesis),
+                    )
+                    .map(Number)
             },
-        )?;
-
-        Ok(out.unwrap())
+        )
     }
 
     fn expose_public(
@@ -271,7 +241,7 @@ impl<F: FieldExt> NumericInstructions<F> for FieldChip<F> {
     ) -> Result<(), Error> {
         let config = self.config();
 
-        layouter.constrain_instance(num.cell, config.instance, row)
+        layouter.constrain_instance(num.0.cell(), config.instance, row)
     }
 }
 // ANCHOR_END: instructions-impl
--- a/examples/simple-example.rs
+++ b/examples/simple-example.rs
@@ -216,11 +216,12 @@ impl<F: FieldExt> NumericInstructions<F> for FieldChip<F> {
                 a.0.copy_advice(|| "lhs", &mut region, config.advice[0], 0)?;
                 b.0.copy_advice(|| "rhs", &mut region, config.advice[1], 0)?;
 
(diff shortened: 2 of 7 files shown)


src/circuit.rs
```diff
@@ -92,6 +92,63 @@ pub struct Cell {
...
+    pub fn value_field(&self) -> Option<F> {
+        self.value.map(|v| v.into())
+    }
+
+    /// Returns the cell.
+    pub fn cell(&self) -> Cell {
+        self.cell
+    }
+
+    /// Copies the value to a given advice cell and constrains them to be equal.
```
ebfull, line 128: This isn't constraining them to be equal though, as far as I can tell? It's just assigning.

src/circuit.rs
```diff
@@ -92,6 +92,64 @@ pub struct Cell {
...
+#[derive(Clone, Copy, Debug)]
+pub struct AllocatedCell<F: FieldExt, T: Copy + Into<F>> {
+    value: Option<T>,
+    cell: Cell,
+    _marker: PhantomData<F>,
+}
+
+impl<F: FieldExt, T: Copy + Into<F>> AllocatedCell<F, T> {
+    /// Instantiate a new AllocatedCell.
+    pub fn new(value: Option<T>, cell: Cell) -> Self {
```
therealyingtong, line 105: Should this be `pub` or `pub(crate)`? The `halo2-gadgets` crate will likely need to use it.

src/circuit.rs
```diff
@@ -92,6 +92,91 @@ pub struct Cell {
...
+    pub fn value_field(&self) -> Option<F> {
+        self.value.map(|v| v.into())
+    }
+
+    /// Returns the cell.
+    pub fn cell(&self) -> Cell {
+        self.cell
+    }
+
+    /// Copies the value to a given advice cell and constrains them to be equal.
```
daira, line 155:
```suggestion
    /// Copies the value to a given advice cell and constrains them to be equal.
    ///
    /// Returns an error if either this cell or the given cell are in columns
    /// where equality has not been enabled.
```

src/circuit.rs
```diff
@@ -92,6 +92,91 @@ pub struct Cell {
...
+                    region.assign_advice(annotation, column.try_into().unwrap(), offset, || {
+                        value.ok_or(Error::SynthesisError)
+                    })?
+                }
+                Any::Fixed => {
+                    region.assign_fixed(annotation, column.try_into().unwrap(), offset, || {
+                        value.ok_or(Error::SynthesisError)
+                    })?
+                }
+                _ => unreachable!(),
```
ebfull, line 129: This should be a `panic!("cells in an instance column cannot be assigned to")` or an error, and the documentation for the method should explain.

src/circuit.rs
```diff
@@ -92,6 +92,91 @@ pub struct Cell {
...
+#[derive(Clone, Copy, Debug)]
+pub struct AssignedCell<F: Field, T: Copy + Into<F>> {
+    value: Option<T>,
+    cell: Cell,
+    _marker: PhantomData<F>,
+}
+
+impl<F: Field, T: Copy + Into<F>> AssignedCell<F, T> {
+    /// Assign a new AssignedCell.
+    pub fn assign<A, AR>(
```
ebfull, line 105: Perhaps this should be called `assign_unchecked` (or whatever) so that if you have an `AssignedCell<F, T>` you aren't misled into thinking that the circuit has constrained the cell to be a value of `T` (which this code does not do). The result is that reviewers of circuit code will be more careful when looking at `assign_unchecked` code, so it'll be better to isolate its usage in abstractions like `AssignedBool` and so on.

src/circuit.rs
```diff
@@ -92,6 +92,91 @@ pub struct Cell {
     column: Column<Any>,
 }
 
+/// An assigned cell.
+#[derive(Clone, Copy, Debug)]
+pub struct AssignedCell<F: Field, T: Copy + Into<F>> {
```
ebfull, line 97: `Copy` is overly strict, you could use `Clone` instead. (Which will effectively act as a copy for `Copy` types.)

src/circuit.rs
```diff
@@ -92,6 +92,91 @@ pub struct Cell {
...
+    cell: Cell,
+    _marker: PhantomData<F>,
+}
+
+impl<F: Field, T: Copy + Into<F>> AssignedCell<F, T> {
+    /// Assign a new AssignedCell.
+    pub fn assign<A, AR>(
+        region: &mut Region<'_, F>,
+        annotation: A,
+        column: Column<Any>,
```
ebfull, line 108: Idea: `column: impl Into<Column<Any>>` (maybe we can do this in other places too...)

src/circuit/floor_planner/single_pass.rs
```diff
@@ -302,11 +302,15 @@ impl<'r, 'a, F: Field, CS: Assignment<F> + 'a> RegionLayouter<F>
...
         constant: Assigned<F>,
-    ) -> Result<Cell, Error> {
+    ) -> Result<AssignedCell<F, F>, Error> {
         let advice = self.assign_advice(annotation, column, offset, &mut || Ok(constant))?;
         self.constrain_constant(advice, constant)?;
 
-        Ok(advice)
+        Ok(AssignedCell {
+            cell: advice,
+            value: Some(constant.evaluate()),
```
str4d, line 311: This undoes the optimisation that we were trying to make with the `Assigned` type, to enable batching inversions (because this method evaluates the inversion to obtain an `F`). I think the type should be `Option<Assigned<F>>`.

src/circuit/layouter.rs
```diff
@@ -69,7 +69,7 @@ pub trait RegionLayouter<F: Field>: fmt::Debug {
         column: Column<Advice>,
         offset: usize,
         constant: Assigned<F>,
-    ) -> Result<Cell, Error>;
+    ) -> Result<AssignedCell<F, F>, Error>;
```
str4d, line 72: This feels like an abstraction failure, because `RegionLayouter::assign_advice` and `RegionLayouter::assign_advice_from_instance` still return `Cell`. Either they should all return `AssignedCell`, or none of them should (and instead the caller should use `AssignedCell` constructors, as this PR currently provides for advice assigmnent).

str4d (commented):
Flusing comments from my pairing with @therealyingtong.

src/circuit/layouter.rs
```diff
@@ -232,9 +232,14 @@ impl<F: Field> RegionLayouter<F> for RegionShape {
...
         offset: usize,
         constant: Assigned<F>,
-    ) -> Result<Cell, Error> {
+    ) -> Result<AssignedCell<F, F>, Error> {
         // The rest is identical to witnessing an advice cell.
-        self.assign_advice(annotation, column, offset, &mut || Ok(constant))
+        let cell = self.assign_advice(annotation, column, offset, &mut || Ok(constant))?;
+        Ok(AssignedCell {
+            cell,
+            value: None,
```
str4d, line 240: Technically we always have the constant value here, so we should just return it.

src/circuit.rs
```diff
@@ -92,6 +92,96 @@ pub struct Cell {
...
+        Ok(AssignedCell {
+            value,
+            cell,
+            _marker: PhantomData,
+        })
+    }
+
+    /// Returns the value of the AssignedCell.
+    pub fn value(&self) -> Option<T> {
+        self.value.clone()
```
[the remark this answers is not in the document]
therealyingtong: Trying to remove the `Clone` bound on `T` led me to a situation where I had to take an immutable borrow and a mutable borrow of `layouter` at the same time.

examples/two-chip.rs
```diff
@@ -218,38 +214,24 @@ impl<F: FieldExt> AddInstructions<F> for AddChip<F> {
...
 
-                // Now we can assign the multiplication result into the output position.
-                let value = a.value.and_then(|a| b.value.map(|b| a + b));
-                let cell = region.assign_advice(
-                    || "lhs * rhs",
-                    config.advice[0],
-                    1,
-                    || value.ok_or(Error::Synthesis),
-                )?;
+                // Now we can assign the addition result into the output position.
```
daira, line 220:
```suggestion
                // Now we can compute the addition result, which is to be
                // assigned into the output position.
```
(The code up to the next comment no longer does the assignment.)

examples/two-chip.rs
```diff
@@ -218,38 +214,24 @@ impl<F: FieldExt> AddInstructions<F> for AddChip<F> {
...
+                let value = a.0.value().and_then(|a| b.0.value().map(|b| *a + *b));
 
                 // Finally, we return a variable representing the output,
                 // to be used in another part of the circuit.
```
daira, line 224:
```suggestion
                // Finally, we do the assignment to the output, returning a variable
                // to be used in another part of the circuit.
```

examples/simple-example.rs
```diff
@@ -229,38 +213,24 @@ impl<F: FieldExt> NumericInstructions<F> for FieldChip<F> {
...
-                    config.advice[1],
-                    0,
-                    || b.value.ok_or(Error::Synthesis),
-                )?;
-                region.constrain_equal(a.cell, lhs)?;
-                region.constrain_equal(b.cell, rhs)?;
+                a.0.copy_advice(|| "lhs", &mut region, config.advice[0], 0)?;
+                b.0.copy_advice(|| "rhs", &mut region, config.advice[1], 0)?;
 
                 // Now we can assign the multiplication result into the output position.
```
daira, line 219:
```suggestion
                // Now we can assign the multiplication result, which is to be assigned
                // into the output position.
```

examples/simple-example.rs
```diff
@@ -229,38 +213,24 @@ impl<F: FieldExt> NumericInstructions<F> for FieldChip<F> {
...
-                let cell = region.assign_advice(
-                    || "lhs * rhs",
-                    config.advice[0],
-                    1,
-                    || value.ok_or(Error::Synthesis),
-                )?;
+                let value = a.0.value().and_then(|a| b.0.value().map(|b| *a * *b));
 
                 // Finally, we return a variable representing the output,
                 // to be used in another part of the circuit.
```
daira, line 223:
```suggestion
                // Finally, we do the assignment to the output, returning a
                // variable to be used in another part of the circuit.
```

examples/two-chip.rs
```diff
@@ -358,38 +339,24 @@ impl<F: FieldExt> MulInstructions<F> for MulChip<F> {
...
-                    config.advice[1],
-                    0,
-                    || b.value.ok_or(Error::Synthesis),
-                )?;
-                region.constrain_equal(a.cell, lhs)?;
-                region.constrain_equal(b.cell, rhs)?;
+                a.0.copy_advice(|| "lhs", &mut region, config.advice[0], 0)?;
+                b.0.copy_advice(|| "rhs", &mut region, config.advice[1], 0)?;
 
                 // Now we can assign the multiplication result into the output position.
```
daira, line 345:
```suggestion
                // Now we can compute the multiplication result, which is to be assigned
                // into the output position.
```

examples/two-chip.rs
```diff
@@ -358,38 +339,24 @@ impl<F: FieldExt> MulInstructions<F> for MulChip<F> {
...
-                let cell = region.assign_advice(
-                    || "lhs * rhs",
-                    config.advice[0],
-                    1,
-                    || value.ok_or(Error::Synthesis),
-                )?;
+                let value = a.0.value().and_then(|a| b.0.value().map(|b| *a * *b));
 
                 // Finally, we return a variable representing the output,
                 // to be used in another part of the circuit.
```
daira, line 349:
```suggestion
                // Finally, we do the assignment to the output, returning a
                // variable to be used in another part of the circuit.
```

benches/plonk.rs
```diff
@@ -110,7 +110,7 @@ fn bench_with_k(name: &str, k: u32, c: &mut Criterion) {
                     region.assign_fixed(|| "b", self.config.sb, 0, || Ok(FF::zero()))?;
                     region.assign_fixed(|| "c", self.config.sc, 0, || Ok(FF::one()))?;
                     region.assign_fixed(|| "a * b", self.config.sm, 0, || Ok(FF::one()))?;
-                    Ok((lhs, rhs, out))
+                    Ok((lhs.cell(), rhs.cell(), out.cell()))
```
str4d, line 113: We should clean up the examples to use the newer APIs where possible. I've opened #412 to track this.

src/circuit.rs
```diff
@@ -92,6 +92,64 @@ pub struct Cell {
     column: Column<Any>,
 }
 
+/// An assigned cell.
+#[derive(Clone, Debug)]
```
str4d, line 96: Something neat I probably already knew but [just became aware of](https://stegosaurusdormant.com/understanding-derive-clone/) is that if you add `#[derive(Clone)]` to a struct with generics, Rust will implement it in a way that the struct is `Clone` when its type parameters are all `Clone`.

I think it's okay to have `impl Clone for AssignedCell`, because the underlying `Cell` is a pointer to the actual cell, so what we'd be cloning is the pointer in addition to the value, which is fine to do without layouter awareness. I don't believe this will be confused with `AssignedCell::copy_advice`, which clones only the value and assigns a new `Cell`, because `Clone` can't be used to satisfy a constraint (as the cell will be in the wrong location).

str4d (approved):
utACK 839f9ecd66f5ab301db319c04425bffcdf4b6141

src/circuit.rs
```diff
@@ -139,17 +197,28 @@ impl<'r, F: Field> Region<'r, F> {
...
         AR: Into<String>,
     {
-        self.region
-            .assign_advice(&|| annotation().into(), column, offset, &mut || {
-                to().map(|v| v.into())
-            })
+        let mut value = None;
+        let cell =
+            self.region
+                .assign_advice(&|| annotation().into(), column, offset, &mut || {
```
ebfull, line 210: Why isn't the `annotation` just passed in directly in these sorts of places, instead of wrapping it in another closure?

ebfull (approved):

src/circuit.rs
str4d: Lifetimes IIRC.

str4d:
GitHub wouldn't let me batch all of @daira's code comment suggestions, so I created several commits, then [no-op force-pushed](https://github.com/zcash/halo2/compare/c556c92ece103ac2df82cef53e49735ef2238c06..4b8f0f0c6bc29a2079526fc03000085c2bbcf8e7) to squash them into a single commit.

examples/simple-example.rs
```diff
@@ -229,38 +213,25 @@ impl<F: FieldExt> NumericInstructions<F> for FieldChip<F> {
...
-
-                // Finally, we return a variable representing the output,
-                // to be used in another part of the circuit.
-                out = Some(Number { cell, value });
-                Ok(())
+                a.0.copy_advice(|| "lhs", &mut region, config.advice[0], 0)?;
+                b.0.copy_advice(|| "rhs", &mut region, config.advice[1], 0)?;
+
+                // Now we can assign the multiplication result, which is to be assigned
+                // into the output position.
```
daira, line 220: This comment was missed; change it to be like the others.

src/circuit.rs
```diff
@@ -212,17 +293,28 @@ impl<'r, F: Field> Region<'r, F> {
...
-            .assign_fixed(&|| annotation().into(), column, offset, &mut || {
-                to().map(|v| v.into())
-            })
+        let mut value = None;
+        let cell =
+            self.region
+                .assign_fixed(&|| annotation().into(), column, offset, &mut || {
+                    let v = to()?;
+                    let value_f = (&v).into();
+                    value = Some(v);
```
daira, line 309: Is the closure guaranteed to be called immediately by `assign_fixed` (if it is called at all)?

daira (approved):
utACK with comments.

src/circuit.rs
daira: As @str4d pinned out, the lifetime of the closure is such that it could only be called before `assign_fixed` returns if it is called at all.

examples/simple-example.rs
str4d: To avoid another round of re-ACKs, I'll defer this to #412.

PR #394, opened 2021-11-03
impl std::error::Error for halo2::plonk::Error

dconnolly:
We're calling `orchard::circuit::Proof::verify()` from zebra and are trying to handle the returned error type `halo2::plonk::Error` nicely. If we impl `std::error::Error` on it here, we don't need to create wrapper types for it.


str4d:
A `std::error::Error` implementation is now included as part of #338.

PR #403, opened 2021-11-15
Improve error message when `k` is too small

### LINKED ISSUE #402: Provide more useful error messages when `k` is too small
A common error condition is when a circuit developer picks a `k` that is too small to provide sufficient rows to contain their circuit. Currently however, this error manifests as an out-of-bounds panic from the depths of the proving or dev code. For example, when using a small `k` with `halo2::dev::CircuitLayout`:
```
thread 'main' panicked at 'index out of bounds: the len is 32 but the index is 32', halo2/src/dev/graph/layout.rs:420:9
```

We should specifically detect the cases where an out-of-bounds error is caused by insufficient rows, and return a more useful error indicating the problem (and suggesting the value of `k` that would resolve the problem).

str4d:
Closes #402.


therealyingtong (approved):
utACK

PR #406, opened 2021-11-19
Multiopen static ordering

### LINKED ISSUE #210: Make multiopen point_index_map dependent only on the rotation
Right now, the multiopen API allows you to specify arbitrary points. However, in practice we _always_ open at points of the form `omega^i x` for some fixed `x` and rotation `i`. The current implementation of multiopen will create a point index map that essentially creates an ordering on the points. This ordering is determined by `x`.

However, in a recursive circuit we need this ordering to be fixed by the rotations; if it varied by the concrete `x` (which is a random challenge) then the circuit would have to determine this ordering, which is not trivial. Let's make the behavior of multiopen depend only on the rotations.

ebfull:
Closes #210

This PR comes in multiple stages (corresponding with each commit):

* We add a constructor for `ProverQuery` that doesn't change behavior in.
* We force the prover to provide the `x` challenge and rotation used to derive the point and then assert that they computed it correctly in order to avoid a subtle typo somewhere.
* We do the same thing but with the verifier.
* We remove a test (and keep the roundtrip test) which was basically a silly wannabe proptest, we can add a proptest for this another time (or in this PR if anyone insists).
* We finally change the way the point sets are constructed so that they are determined statically by the rotations (which don't vary between proofs).
* And then we remove the unnecessary (now) assertions together with any of the old logic for computing rotated points that remained


src/plonk/lookup/prover.rs
```diff
@@ -543,57 +543,35 @@ impl<C: CurveAffine> Constructed<C> {
 }
```
daira, line 543: Clippy says some struct fields are never read after this change. If that's correct, remove those fields.

ebfull:
Hm. I think that I misread the original code. As far as I can tell, the existing code _is_ static. It establishes an ordering of the points internally, based on the order that the queries are defined, and the `BTreeMap`s are just used for mapping purposes. (That is, the order of things placed into those maps doesn't matter. We don't iterate over any `BTreeMap`s.)

PR #408, opened 2021-11-23
multiopen: Replace test_coherence with proptest.

### LINKED ISSUE #407: multiopen: Write proptest over arbitrary rotations.
This will replace the test deleted in https://github.com/zcash/halo2/pull/406/files.

therealyingtong:
Closes #407.

### DIFF
--- a/src/poly/multiopen.rs
+++ b/src/poly/multiopen.rs
@@ -102,6 +102,7 @@ impl<'r, 'params: 'r, C: CurveAffine> PartialEq for CommitmentReference<'r, 'par
     }
 }
 
+#[derive(Debug)]
 struct CommitmentData<F, T: PartialEq> {
     commitment: T,
     set_index: usize,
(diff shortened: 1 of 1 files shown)


src/poly/multiopen.rs
```diff
@@ -382,138 +383,116 @@ mod tests {
...
         }
+    }
+
+    prop_compose! {
+        // List of point indices corresponding to each query index.
+        fn arb_queries_inner(num_points: usize, num_queries: usize)(
+            point_indices in vec(select((0..num_points).collect::<Vec<_>>()), num_queries)
+        ) -> Vec<usize> {
+            point_indices
+        }
```
str4d, line 454: From PR gardening with @nuttycom and @therealyingtong: currently we end up with point indices being potentially duplicated (good, corresponds to matching rotations), but commitment indices being entirely unique (meaning we're only testing one rotation per column). This should return a list of tuples representing each query, so that we can also get duplicate commitment indices.

daira (approved):
utACK

ebfull (approved):

PR #409, opened 2021-11-23
Fix various clippy lints from beta toolchain

str4d:


str4d:
Remaining lints are all unused fields:

- `Lookup.columns` in the circuit cost logic; this may be used soonish once I start adding the prover and verifier cost logic.
- Most of the `CommitmentData` fields in the multiopen logic; I think those start being used by #408, so if that PR merges before this one, I can rebase this to recheck.

ebfull (approved):

PR #413, opened 2021-12-03
sha256: Fix `AssignedCell` `From` impls for `AssignedBits` and `Bits`

str4d:
zcash/halo2#337 had been updated to account for `AssignedCell`, before
the change to simplify its bounds was made. `AssignedCell` now requires
the `From` impls to take `&VR`.


str4d:
Self-ACK, trivially-correct PR that only touches the SHA-256 example code.

PR #417, opened 2021-12-08
Expand `Assigned<F>` APIs

str4d:
This PR adds various usability methods to `Assigned<F: Field>`, to make it easier to use in contexts where `F: Field` was previously used.

Closes zcash/halo2#425.

### DIFF
--- a/halo2_proofs/src/plonk/assigned.rs
+++ b/halo2_proofs/src/plonk/assigned.rs
@@ -152,6 +152,18 @@ impl<F: Field> Assigned<F> {
         }
     }
 
+    /// Returns true iff this element is zero.
+    pub fn is_zero_vartime(&self) -> bool {
+        match self {
+            Self::Zero => true,
+            Self::Trivial(x) => x.is_zero_vartime(),
+            // Assigned maps x/0 -> 0.
+            Self::Rational(numerator, denominator) => {
+                numerator.is_zero_vartime() || denominator.is_zero_vartime()
+            }
+        }
+    }
+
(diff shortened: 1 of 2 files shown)


str4d:
Converted this to a draft, because we actually need a bunch more usability methods on this type. I'll add them to the PR tomorrow.

ebfull (approved):

ebfull (dismissed):
Actually, this is missing corresponding changelog updates I think.

halo2_proofs/src/plonk/assigned.rs
```diff
@@ -440,15 +528,31 @@ mod proptests {
...
         )(
             values in vec(
                 prop_oneof![Just(Assigned::Zero), arb_trivial(), arb_rational()],
-                num_operations + 1),
-            operations in vec(select(OPERATIONS), num_operations),
-        ) -> (Vec<Assigned<Fp>>, Vec<Operation>) {
+                // Ensure that:
+                // - we have at least one value to apply unary operators to.
+                // - we can apply every binary operator pairwise sequentially.
+                cmp::max(if num_unary > 0 { 1 } else { 0 }, num_binary + 1)),
```
therealyingtong, line 560: In the case where `num_unary == num_binary == 0`, do we need any values?

therealyingtong (approved):
utACK with a question about the proptest.

halo2_proofs/src/plonk/assigned.rs
str4d: It makes the rest of the rest easier to write if we always have at least one value (so we always have something to check at the end).

Issue #419, opened 2021-12-08
CI: Add benchmark tracking


str4d:
We should use something like https://github.com/benchmark-action/github-action-benchmark to track the repo benchmarks over time, so we can detect regressions.

str4d:
This is now live! https://zcash.github.io/halo2/dev/bench/

PR #421, opened 2021-12-08
CI: Add benchmarks workflow

str4d:
The workflow will comment on commits that cause a performance regression
of at least 200% (e.g. proving taking twice as long); we'll tune this as
we figure out how well benchmarking works on standard GitHub builders.

Part of zcash/halo2#419.


str4d:
Self-ACK; this doesn't alter any existing code. It only adds a new CI workflow, which needs to reach `main` before it can be tested.

PR #422, opened 2021-12-08
CI: Benchmark tweaks

str4d:
Part of zcash/halo2#419.

### DIFF
--- a/benches/plonk.rs
+++ b/benches/plonk.rs
@@ -11,15 +11,12 @@ use halo2::transcript::{Blake2bRead, Blake2bWrite, Challenge255};
 
 use std::marker::PhantomData;
 
-use criterion::Criterion;
+use criterion::{BenchmarkId, Criterion};
 
-fn bench_with_k(name: &str, k: u32, c: &mut Criterion) {
+fn criterion_benchmark(c: &mut Criterion) {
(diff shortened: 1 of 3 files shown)


benches/plonk.rs
```diff
@@ -247,61 +244,74 @@ fn bench_with_k(name: &str, k: u32, c: &mut Criterion) {
...
+    }
+    keygen_group.finish();
+
+    let mut prover_group = c.benchmark_group("plonk-prover");
+    prover_group.sample_size(10);
+    for k in k_range.clone() {
+        let (params, pk) = keygen(k);
+
+        prover_group.bench_with_input(
+            BenchmarkId::from_parameter(k),
```
str4d, line 292: @ebfull this change will create benchmark names like `plonk-prover/11`, which assumes this benchmark will only vary `k`. Does that seem likely to change at all, or should I alter the naming to make it `plonk-prover/k/11`?

daira: This seems fine to me. It doesn't preclude adding other parameters later.

daira (approved):
utACK

benches/plonk.rs
str4d: Yeah, we can always go into the `gh-pages` branch and rename the existing benchmarks to preserve past data.

PR #431, opened 2021-12-20
Improve `dev::VerifyFailure::Lookup` information

str4d:
Previously, we only reported the absolute row on which a failure occurred; this is unhelpful when an optimising floor planner is in use. Now, if the lookup failure overlaps a region, we report that instead.

### DIFF
--- a/src/dev.rs
+++ b/src/dev.rs
@@ -33,6 +33,29 @@ mod graph;
 #[cfg_attr(docsrs, doc(cfg(feature = "dev-graph")))]
 pub use graph::{circuit_dot_graph, layout::CircuitLayout};
 
+/// The location at which a particular lookup is not satisfied.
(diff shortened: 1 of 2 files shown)


src/dev.rs
```diff
@@ -969,4 +1067,110 @@ mod tests {
...
+
+        let prover = MockProver::run(K, &FaultyCircuit {}, vec![]).unwrap();
+        assert_eq!(
+            prover.verify(),
+            Err(vec![VerifyFailure::Lookup {
+                lookup_index: 0,
+                location: LookupFailure::InRegion {
+                    region: (2, "Faulty synthesis".to_owned()).into(),
+                    offset: 1,
+                }
```
daira, line 1172: There is no test for the OutOfRegion case, but I think that's fine?

daira (approved):
utACK

PR #433, opened 2021-12-21
dev: Enable `VerifyFailure::ConstraintNotSatisfied` to point to region offsets

str4d:
Follows on from #431.


therealyingtong (approved):
utACK

PR #442, opened 2021-12-23
[book] Fix SHA256 interpolation polynomial

### LINKED ISSUE #434: [book] Error in an interpolation polynomial for SHA-256
The polynomial for ℓ<sub>3</sub>(a) at https://zcash.github.io/halo2/design/gadgets/sha256/table16.html#3-bit-spread is incorrect. (All the interpolation polynomials need to be checked; that's just the first I found that is definitely wrong.)

This does not affect the Orchard circuit, which does not use SHA-256. It probably also doesn't affect the SHA-256 circuit implementation, because there [the polynomials are computed](https://github.com/zcash/halo2/blob/2a7985014b39f41a37fc8fed317dc483f55a0de0/examples/sha256/table16/gates.rs#L11-L61) rather than written out by hand.

therealyingtong:
Closes #434


daira (commented):
Post-hoc ACK

PR #445, opened 2021-12-28
Fix mock prover performance regression for lookup arguments

### LINKED ISSUE #398: [MockProver] Testing is slower after #389
### Problem

The test speed is slower when #389 is applied.

We use the test of this branch for experiment https://github.com/appliedzkp/zkevm-circuits/pull/156

`cargo test --release test_rho_gate -p keccak256  -- --nocapture` is about 11 sec on old commit `4283713e`, but is about 34 sec on `1723fed1`, which is `4283713e` with `git cherry-pick b94aab76a338870fec778d4da3a1c6789bc92bc4`.

The hardware is the apple M1 silicon.

### Solution

Need investigation

daira:
closes #398

### DIFF
--- a/halo2_proofs/src/dev.rs
+++ b/halo2_proofs/src/dev.rs
@@ -915,8 +915,18 @@ impl<F: FieldExt> MockProver<F> {
                     };
 
                     assert!(lookup.table_expressions.len() == lookup.input_expressions.len());
-                    let fill_row = vec![Value::Real(F::zero()); lookup.table_expressions.len()];
-                    let mut table_has_fill_row = false;
+                    assert!(self.usable_rows.end > 0);
+
+                    // We optimize on the basis that the last usable row might have been repeated to fill
+                    // the table (it doesn't matter if it wasn't). Note that this "fill row" necessarily
+                    // exists in the table, and we use that fact to slightly simplify the optimization:
+                    // we're only trying to check that all input rows are contained in the table, and so
+                    // we can safely just drop input rows that match the fill row.
+                    let fill_row: Vec<_> = lookup
+                        .table_expressions
+                        .iter()
+                        .map(move |c| load(c, self.usable_rows.end - 1))
+                        .collect();
 
                     // In the real prover, the lookup expressions are never enforced on
                     // unusable rows, due to the (1 - (l_last(X) + l_blind(X))) term.
@@ -933,7 +943,6 @@ impl<F: FieldExt> MockProver<F> {
                             if t != fill_row {
                                 Some(t)
                             } else {
-                                table_has_fill_row = true;
                                 None
                             }
                         })
@@ -950,8 +959,7 @@ impl<F: FieldExt> MockProver<F> {
                                 .map(move |c| load(c, input_row))
                                 .collect();
 
-                            // Include fill rows in `inputs` only if `table` has no fill rows.
-                            if !table_has_fill_row || t != fill_row {
+                            if t != fill_row {
                                 // Also keep track of the original input row, since we're going to sort.
                                 Some((t, input_row))
                             } else {
--- a/halo2_proofs/src/dev.rs
+++ b/halo2_proofs/src/dev.rs
@@ -917,11 +917,12 @@ impl<F: FieldExt> MockProver<F> {
                     assert!(lookup.table_expressions.len() == lookup.input_expressions.len());
                     assert!(self.usable_rows.end > 0);
 
-                    // We optimize on the basis that the last usable row might have been repeated to fill
-                    // the table (it doesn't matter if it wasn't). Note that this "fill row" necessarily
-                    // exists in the table, and we use that fact to slightly simplify the optimization:
-                    // we're only trying to check that all input rows are contained in the table, and so
-                    // we can safely just drop input rows that match the fill row.
+                    // We optimize on the basis that the table might have been filled so that the last
+                    // usable row now has the fill contents (it doesn't matter if there was no filling).
+                    // Note that this "fill row" necessarily exists in the table, and we use that fact to
+                    // slightly simplify the optimization: we're only trying to check that all input rows
+                    // are contained in the table, and so we can safely just drop input rows that
+                    // match the fill row.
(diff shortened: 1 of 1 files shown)


daira:
I recommend to review ignoring whitespace changes.

src/dev.rs
```diff
@@ -835,90 +835,128 @@ impl<F: FieldExt> MockProver<F> {
...
+                        .iter()
+                        .filter_map(move |(input, input_row)| {
+                            let input_row = *input_row;
+                            while i < table.len() && &table[i] < input {
+                                i += 1;
+                            }
+                            if i == table.len() || &table[i] > input {
+                                assert!(table.binary_search(input).is_err());
+
+                                let input_columns = calc_input_columns();
```
daira, line 924: This doesn't depend on the input row so we could cache it, but that's a separate optimization, and would only affect the case where we have lots of lookup failures.

src/dev.rs
```diff
@@ -835,90 +835,128 @@ impl<F: FieldExt> MockProver<F> {
...
+                                println!("Lookup input columns: {:?}", &input_columns);
```
ebfull, line 925: Remove this print statement.

ebfull (changes requested):

src/dev.rs
daira:
```suggestion
```

str4d:
This will conflict with #433; I'd prefer to merge that PR before this one, in part because it will simplify the diff here.

src/dev.rs
```diff
@@ -835,90 +835,127 @@ impl<F: FieldExt> MockProver<F> {
                         )
                     };
 
+                    assert!(lookup.table_expressions.len() == lookup.input_expressions.len());
+                    let zeros = vec![Value::Real(F::zero()); lookup.table_expressions.len()];
+                    let mut table_has_zeros = false;
```
therealyingtong, line 840: I suspect that the `table_has_zeros` case will be less common after #448.

therealyingtong (approved):
utACK with comments.

daira:
Changing this back to draft because I need to change it to optimize fill rows rather than zero rows.

halo2_proofs/src/dev.rs
```diff
@@ -914,40 +914,86 @@ impl<F: FieldExt> MockProver<F> {
...
+                            }
+                        })
+                        .collect();
+                    inputs.sort_unstable();
+
+                    let mut i = 0;
+                    inputs
+                        .iter()
+                        .filter_map(move |(input, input_row)| {
+                            let input_row = *input_row;
```
daira, line 976:
```suggestion
```

halo2_proofs/src/dev.rs
```diff
@@ -914,40 +914,86 @@ impl<F: FieldExt> MockProver<F> {
...
+                            }
+                            if i == table.len() || &table[i] > input {
+                                assert!(table.binary_search(input).is_err());
+
+                                Some(VerifyFailure::Lookup {
+                                    lookup_index,
+                                    location: FailureLocation::find_expressions(
+                                        &self.cs,
+                                        &self.regions,
+                                        input_row,
```
daira, line 988:
```suggestion
                                        *input_row,
```

daira (commented):
Minor simplification.

daira:
This is still up-to-date and needs review.

halo2_proofs/src/dev.rs
```diff
@@ -915,8 +915,18 @@ impl<F: FieldExt> MockProver<F> {
                     };
 
                     assert!(lookup.table_expressions.len() == lookup.input_expressions.len());
-                    let fill_row = vec![Value::Real(F::zero()); lookup.table_expressions.len()];
-                    let mut table_has_fill_row = false;
+                    assert!(self.usable_rows.end > 0);
+
+                    // We optimize on the basis that the last usable row might have been repeated to fill
+                    // the table (it doesn't matter if it wasn't). Note that this "fill row" necessarily
```
str4d, line 921: This doesn't match what we actually do in either `SimpleFloorPlanner` or `floor_planner::V1`, which is to use the _first_ row of the table assignment as the fill row:

https://github.com/zcash/halo2/blob/221766986b52d2a35f60d94266647d9dcf908231/halo2_proofs/src/circuit/floor_planner/single_pass.rs#L437-L444

As you say, it doesn't really matter; the performance improvement we observe is not due to this optimisation.

str4d (approved):
ACK; I observe the desired speed-up.

halo2_proofs/src/dev.rs
daira: This is a misunderstanding of the comment. Whichever policy was used to pick the fill row, provided that there were any fill rows, the table contents after filling will have the contents of the fill row in the last usable row. (If there were no fill rows, treating the last usable row as the potential fill contents is still harmless.)

halo2_proofs/src/dev.rs
```diff
@@ -914,40 +914,85 @@ impl<F: FieldExt> MockProver<F> {
...
                     };
 
+                    assert!(lookup.table_expressions.len() == lookup.input_expressions.len());
+                    assert!(self.usable_rows.end > 0);
+
+                    // We optimize on the basis that the last usable row might have been repeated to fill
+                    // the table (it doesn't matter if it wasn't). Note that this "fill row" necessarily
+                    // exists in the table, and we use that fact to slightly simplify the optimization:
+                    // we're only trying to check that all input rows are contained in the table, and so
+                    // we can safely just drop input rows that match the fill row.
```
daira, line 924:
```suggestion
                    // We optimize on the basis that the table might have been filled so that the last
                    // usable row now has the fill contents (it doesn't matter if there was no filling).
                    // Note that this "fill row" necessarily exists in the table, and we use that fact to
                    // slightly simplify the optimization: we're only trying to check that all input rows
                    // are contained in the table, and so we can safely just drop input rows that
                    // match the fill row.
```

str4d (approved):
re-utACK b48b032041d74a6ac6148d72f9b44b8f52972192

PR #446, opened 2021-12-29
fix typos

HAOYUatHZ:


daira (approved):
ACK, trivial error-message-only change in a benchmark.

Issue #360, opened 2021-08-12
Smarter parallelization strategy


ebfull:
Right now our parallelization strategy is applied inside of each individual polynomial operation ([see here for example](https://github.com/zcash/halo2/blob/27c4187673a9c6ade13fbdbd4f20955530c22d7f/src/poly.rs#L246)), which means that for every operation between polynomials we end up starting a new threadpool, splitting the polynomial into chunks and then doing the pairwise operation.

This is clearly inefficient in many cases because _everything_ is pairwise -- we should be able to split the polynomials up at a higher layer and then each thread applies all the various operations pairwise on its local section. This avoids overheads from spinning up threads that are quickly thrown away which constitutes a pretty significant portion of the current proving time.

str4d:
Fixing this will likely also fix a performance problem for large circuits where the same rotation is computed multiple times: https://github.com/zcash/halo2/issues/427#issuecomment-991170170

Issue #315, opened 2021-06-30
Address performance regression for rayon::ThreadPool on WSL2


daira:
Using the rayon threadpool resulted in significant speed-ups on native Linux and Windows, but caused a regression to verification performance (and a slight regression to prover performance) on WSL2, relative to using `crossbeam_utils::thread::scope` without a threadpool. See #311 for context and discussion.

Since few users will be on WSL2, we decided to merge #311 without any way to disable the threadpool usage. This ticket is to address the regression, either by detecting WSL2 or by configuring/fixing rayon to solve the underlying cause.

PR #447, opened 2021-12-30
Add `poly::Evaluator` for building polynomial operation ASTs

str4d:
Once complete, this will close zcash/halo2#360.


src/plonk/circuit.rs
```diff
@@ -761,6 +761,125 @@ impl<F: Field> Expression<F> {
...
+                    context,
+                    constant,
+                    selector_column,
+                    fixed_column,
+                    advice_column,
+                    instance_column,
+                    negated,
+                    sum,
+                    product,
+                    scaled,
```
daira, line 808: Perhaps it would be more concise to bundle these into a struct?

src/poly/evaluator.rs
```diff
@@ -0,0 +1,386 @@
...
+    ///
+    /// This API registers the polynomial with the given identifier, as a cheaper way of
+    /// detecting duplicate polynomials than an equality check. The caller must ensure the
+    /// one-to-one mapping between identifiers and polynomials.
+    pub(crate) fn register_poly_as(
+        &mut self,
+        id: Id,
+        poly: &Polynomial<F, B>,
+    ) -> AstLeaf<E, Id, B> {
+        self.polys.entry(id).or_insert_with(|| poly.clone());
```
daira, line 96: Consider treating reuse of the same id as an error.

src/plonk/lookup/prover.rs
```diff
@@ -94,82 +96,110 @@ impl<F: FieldExt> Argument<F> {
...
                 .collect();
 
             // Compressed version of expressions
-            let compressed_expression = unpermuted_expressions
-                .iter()
-                .fold(domain.empty_lagrange(), |acc, expression| {
-                    acc * *theta + expression
-                });
+            let compressed_expression = unpermuted_expressions.iter().fold(
+                poly::Ast::ConstantTerm(C::Scalar::zero()),
```
daira, line 189: Consider using the `.reduce(...).unwrap_or(zero)` pattern here, if it would help.

src/poly/evaluator.rs
```diff
@@ -0,0 +1,386 @@
...
+pub(crate) struct AstLeaf<E, Id, B: Basis> {
+    id: Id,
+    rotation: Rotation,
+    _evaluator: PhantomData<(E, B)>,
+}
+
+impl<E, Id: Copy, B: Basis> AstLeaf<E, Id, B> {
+    /// Produces a new `AstLeaf` node corresponding to the underlying polynomial at a
+    /// _new_ rotation. Existing rotations applied to this leaf node are ignored and the
+    /// returned polynomial is not rotated _relative_ to the previous structure.
```
daira, line 27: THis is kind-of odd, because it depends on how the original `AstLeaf` was constructed rather than what it currently represents, which is a leaky abstraction. Applying a rotation relative to the previous one would solve that issue (I think that in practice we're normally calling `with_rotation` on unrotated nodes, so it would still work).

src/poly/evaluator.rs
str4d: The point of having the ID specified was to avoid a more costly polynomial equality check, while being able to ensure that registering the "same" polynomial resulted in the same `AstLeaf`. But I guess we still get the benefit of the `HashMap` removing the O(n) registration time, so I can add this check and then we can always remove it later if it becomes a bottleneck.

src/plonk/lookup/prover.rs
str4d: Here it does not: we are using `.iter()` so we get an iterator of references, which means that `reduce()` attempts to use `&Ast` as the accumulator which doesn't work. We'd need to `.iter().cloned()` which would deep-clone all of the ASTs, whereas `fold()` adds a zero constant to the operation.

src/poly/evaluator.rs
str4d: I originally proposed making this a relative rotation, but @ebfull disagreed. I don't recall his exact rationale.

str4d: We could alternatively replace this API with a registration API that takes a rotation, and under the hood ensures the `AstLeaf` references the same polynomial of this API is called alongside the no-rotation API. There will definitely be contexts where we can't access both the evaluator and the original polynomial, but I don't know how often we will need to get a rotation of an `AstLeaf` there rather than at the beginning.

ebfull: I am fine with an API where you start with something that hasn't been rotated and then get to rotate it, but an API that allows you to rotate on top of a already-rotated leaf is annoying because it requires us to reduce modulo `2^k`, which requires passing the `k` constant around all over the place.

src/poly/evaluator.rs
str4d: I've removed the ID type parameter, as we figured out how to avoid it in a pairing.

src/plonk/circuit.rs
str4d: I've removed this change, as we we figured out how to avoid it in a pairing.

src/poly/evaluator.rs
daira: Why would we need to reduce modulo 2<sup>k</sup>? We're representing an element of Z<sub>2<sup>k</sup></sub> as an integer, which is a valid representation even if it's not canonical. The only weirdness is that equivalent rotations will not compare equal, but if I understand correctly that's not a problem: it could only affect efficiency.

In practice I think it will also not affect efficiency, because practical circuits will only use small rotations (except possibly for [when we span columns](https://zcash.github.io/halo2/design/proving-system/permutation.html#spanning-a-large-number-of-columns) in the permutation argument, which requires a rotation of u = 2k - t - 1, although we could write that as - t - 1).

src/plonk/lookup/prover.rs
```diff
@@ -402,101 +394,82 @@ impl<C: CurveAffine> Permuted<C> {
     }
 }
 
-impl<'a, C: CurveAffine> Committed<C> {
+impl<'a, C: CurveAffine, Ev: Copy + Send + Sync + 'a> Committed<C, Ev> {
     /// Given a Lookup with input expressions, table expressions, permuted input
     /// expression, permuted table expression, and grand product polynomial, this
     /// method constructs constraints that must hold between these values.
     /// This method returns the constraints as a vector of polynomials in
```
daira, line 401:
```suggestion
    /// This method returns the constraints as a vector of polynomial ASTs in
```

src/plonk/prover.rs
```diff
@@ -347,13 +431,15 @@ pub fn create_proof<
...
+                        &value_evaluator,
+                        &mut coset_evaluator,
                         theta,
-                        &advice.advice_values,
-                        &pk.fixed_values,
-                        &instance.instance_values,
-                        &advice.advice_cosets,
-                        &pk.fixed_cosets,
-                        &instance.instance_cosets,
+                        &advice_values,
```
daira, line 437:
```suggestion
                        dvice_values,
```

src/plonk/prover.rs
```diff
@@ -347,13 +431,15 @@ pub fn create_proof<
...
+                        &fixed_values,
+                        &instance_values,
```
daira, line 439:
```suggestion
                        instance_values,
```

src/plonk/prover.rs
```diff
@@ -347,13 +431,15 @@ pub fn create_proof<
...
+                        &advice_cosets,
```
daira, line 440:
```suggestion
                        advice_cosets,
```

src/plonk/prover.rs
```diff
@@ -347,13 +431,15 @@ pub fn create_proof<
...
+                        &fixed_cosets,
+                        &instance_cosets,
```
daira, line 442:
```suggestion
                        instance_cosets,
```

src/poly/evaluator.rs
```diff
@@ -0,0 +1,542 @@
...
+            .collect();
+
+        // We're working in a single basis, so all polynomials are the same length.
+        let poly_len = self.polys.first().unwrap().len();
+        // Check the level of parallelization we have available.
+        let num_threads = multicore::current_num_threads();
+        // We scale the number of chunks by a constant factor, to ensure that if not all
+        // threads are available, we can achieve more uniform throughput and don't end up
+        // waiting on a couple of threads to process the last chunks.
+        let num_chunks = num_threads * 4;
```
daira, line 145: This probably needs to be larger. The aim is to balance the overhead of having fewer-than-available threads executing the final few chunks, against the overhead of work stealing. The latter will be relatively small per chunk, and my intuition is that it's likely to still be smaller than a few percent until about `num_threads * 32`. We can estimate that the former is less than 1/32 at that point, i.e. also a few percent.

src/poly/evaluator.rs
```diff
@@ -0,0 +1,542 @@
...
+        // Calculate the ideal chunk size for the desired throughput. We use ceiling
+        // division to ensure the minimum chunk size is 1.
+        //     chunk_size = ceil(poly_len / num_chunks)
+        let chunk_size = (poly_len + num_chunks - 1) / num_chunks;
```
daira, line 149: We probably want to impose a minimum greater than 1; otherwise the cost of traversing the AST will be significant compared to doing real work on the polynomial chunks. (This mitigates the possibility of having selected too many chunks above on machines that have many threads, with a smallish circuit.)

src/poly/evaluator.rs
```diff
@@ -0,0 +1,542 @@
...
+
+        // Split each rotated polynomial into chunks.
+        let chunks: Vec<HashMap<_, _>> = (0..num_chunks)
+            .map(|i| {
+                rotated
+                    .iter()
+                    .map(|(leaf, poly)| {
+                        (
+                            *leaf,
+                            poly.chunks(chunk_size)
```
daira, line 162: It's inefficient to waste a thread on a small final chunk. We should instead split the poly more evenly, i.e. with some chunks having ceil(poly_len / num_chunks) elements and some having floor(poly_len / num_chunks).

src/poly/evaluator.rs
```diff
@@ -0,0 +1,542 @@
...
+    ) -> Vec<F> {
+        // Take every power of omega within the chunk, and multiply by scalar.
+        let omega = domain.get_omega();
+        let start = chunk_size * chunk_index;
+        (0..chunk_size)
+            .scan(omega.pow_vartime(&[start as u64]), |cur_omega, _| {
+                let ret = *cur_omega * scalar;
+                *cur_omega *= omega;
+                Some(ret)
+            })
```
daira, line 492:
```suggestion
            .scan(omega.pow_vartime(&[start as u64]) * scalar, |acc, _| {
                let ret = *acc;
                *acc *= omega;
                Some(ret)
            })
```

src/poly/evaluator.rs
```diff
@@ -0,0 +1,542 @@
...
+    fn linear_term<F: FieldExt>(
+        domain: &EvaluationDomain<F>,
+        chunk_size: usize,
+        chunk_index: usize,
+        scalar: F,
+    ) -> Vec<F> {
+        // Take every power of omega within the chunk, and multiply by scalar.
+        let omega = domain.get_omega();
+        let start = chunk_size * chunk_index;
+        (0..chunk_size)
```
daira, line 487: Does this work correctly for the final (potentially smaller) chunk?

src/poly/evaluator.rs
```diff
@@ -0,0 +1,542 @@
...
+        let omega = domain.get_extended_omega();
+        let start = chunk_size * chunk_index;
+        (0..chunk_size)
+            .scan(
+                omega.pow_vartime(&[start as u64]) * F::ZETA,
+                |cur_omega, _| {
+                    let ret = *cur_omega * scalar;
+                    *cur_omega *= omega;
+                    Some(ret)
+                },
```
daira, line 530:
```suggestion
                omega.pow_vartime(&[start as u64]) * F::ZETA * scalar,
                |acc, _| {
                    let ret = *acc;
                    *acc *= omega;
                    Some(ret)
                },
```

src/poly/evaluator.rs
```diff
@@ -0,0 +1,542 @@
...
+        vec![scalar; chunk_size]
+    }
+
+    fn linear_term<F: FieldExt>(
+        domain: &EvaluationDomain<F>,
+        chunk_size: usize,
+        chunk_index: usize,
+        scalar: F,
+    ) -> Vec<F> {
+        // Take every power of the extended omega within the chunk, and multiply by scalar.
```
daira, line 520: Is this supposed to be omega^i * zeta * scalar, or (omega * zeta)^i * scalar?

src/poly/evaluator.rs
str4d: Yes, as long as the AST isn't simply `Ast::LinearTerm`. In any non-trivial AST, this term will be combined via either `Ast::Add` or `Ast::Mul`; if the final chunk is short, then the `.zip()` in either case's impl will ensure we only process up to the end of the chunk, dropping the remainder of the `LinearTerm`.

src/poly/evaluator.rs
str4d: I picked this as a starting point to build the logic around. Once the rest of the PR is complete, we can tune this in a production environment (i.e. Orchard) and see what kinds of overheads we are getting.

src/poly/evaluator.rs
str4d: Yep; again, let's do that in tuning.

src/plonk/prover.rs
```diff
@@ -330,13 +331,96 @@ pub fn create_proof<
...
+        .collect();
+
+    // Register advice values with the polynomial evaluator.
+    let advice_values: Vec<_> = advice
+        .iter()
+        .map(|advice| {
+            advice
+                .advice_values
+                .iter()
+                .map(|poly| value_evaluator.register_poly(poly.clone()))
```
ebfull, line 351: I think it should be possible for us to avoid these clones if we're able to consume these values, since I believe we don't need these after this registration has taken place. But this can happen separately.

src/poly/evaluator.rs
ebfull: Here is what's going on here: we need the polynomial `p(X) = scalar X` evaluated over the coset of the extended domain, which means we need `p(\zeta X)` evaluated over the extended domain (`\zeta` generates the coset), which means we need each `scalar * \zeta * extended_omega^i`.

ebfull (approved):
ACK but did we want to remove the parallel impls of `Add/Mul` for `Polynomial` in this PR?

ebfull:
In a pairing we figured out that replacing `Box` with `Rc` would help a _lot_ in avoiding deep clones and destructor runs for the Ast, let's definitely do that before merging this.

str4d:
I have made the `s/Box/Arc` change (`Arc` is needed because `Rc` can't sent between threads).

Comparing this branch vs current `main` using the https://github.com/zcash/orchard benchmarks (on a Ryzen 9 5950X), we see over 20% improvement:

```
Benchmarking proving/bundle/1: Warming up for 3.0000 s
Warning: Unable to complete 10 samples in 5.0s. You may wish to increase target time to 26.5s.
proving/bundle/1        time:   [2.6532 s 2.6673 s 2.6801 s]                              
                        change: [-22.320% -21.849% -21.319%] (p = 0.00 < 0.05)
                        Performance has improved.
Benchmarking proving/bundle/2: Warming up for 3.0000 s
Warning: Unable to complete 10 samples in 5.0s. You may wish to increase target time to 26.4s.
proving/bundle/2        time:   [2.6432 s 2.6576 s 2.6730 s]                              
                        change: [-23.320% -22.819% -22.300%] (p = 0.00 < 0.05)
                        Performance has improved.
Benchmarking proving/bundle/3: Warming up for 3.0000 s
Warning: Unable to complete 10 samples in 5.0s. You may wish to increase target time to 38.2s.
proving/bundle/3        time:   [3.7982 s 3.8219 s 3.8457 s]                              
                        change: [-23.214% -22.623% -22.029%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 10 measurements (20.00%)
  1 (10.00%) low mild
  1 (10.00%) high mild
Benchmarking proving/bundle/4: Warming up for 3.0000 s
Warning: Unable to complete 10 samples in 5.0s. You may wish to increase target time to 50.4s.
proving/bundle/4        time:   [4.9632 s 4.9994 s 5.0348 s]                              
                        change: [-24.128% -23.487% -22.862%] (p = 0.00 < 0.05)
                        Performance has improved.
```

str4d:
> ACK but did we want to remove the parallel impls of `Add/Mul` for `Polynomial` in this PR?

I've removed the APIs that _are_ now unused. Addition and scalar multiplication are still (currently) used by e.g. the multiopen prover.

str4d:
On that note, I have not yet propagated usage of `poly::Evaluator` into `poly::multiopen` or `poly::commitment`. Doing so might provide some improvements, but probably not as much, since we need to pretty regularly evaluate polynomials in order to commit to them or perform some other operation (e.g. Kate division) that the evaluator can't handle.

src/poly/evaluator.rs
str4d: I tried bumping the factor from `4` to `32`, and proving times for Orchard get slightly worse:
```
proving/bundle/1        time:   [2.6756 s 2.6925 s 2.7089 s]                              
                        change: [+0.3332% +1.0938% +1.8778%] (p = 0.02 < 0.05)
proving/bundle/2        time:   [2.6923 s 2.7061 s 2.7188 s]                              
                        change: [+1.4111% +2.1757% +2.9105%] (p = 0.00 < 0.05)
                        Performance has regressed.
proving/bundle/3        time:   [3.8655 s 3.8811 s 3.8968 s]                              
                        change: [+0.9132% +1.6442% +2.3634%] (p = 0.00 < 0.05)
proving/bundle/4        time:   [5.1000 s 5.1249 s 5.1464 s]                              
                        change: [+0.8925% +1.7243% +2.4899%] (p = 0.00 < 0.05)
```

daira: Right; the trade-off here is adding some average-case overhead in order to reduce the worst case when threads are contended. It's also worth trying 8 or 16.

src/poly/evaluator.rs
daira: Ok; to avoid unexpected surprises (e.g. if the code is reused) we should probably make it work for the trivial case as well.

CHANGELOG.md
```diff
@@ -37,6 +37,11 @@ and this project adheres to Rust's notion of
 ### Removed
 - `halo2::arithmetic::BatchInvert` (use `ff::BatchInvert` instead).
 - `impl Default for halo2::poly::Rotation` (use `Rotation::cur()` instead).
+- `halo2::poly`:
+  - `EvaluationDomain::{add_extended, sub_extended, mul_extended}`
+  - `Polynomial::one_minus`
+  - `impl Neg, Sub for Polynomial`
+  - `impl Mul for Polynomial<_, ExtendedLagrageCoeff>`
```
daira, line 44:
```suggestion
  - `impl Mul for Polynomial<_, ExtendedLagrangeCoeff>`
```

daira (approved):
utACK with comments.

Hurray for the performance improvement! 😻

daira:
My intuition is that this should fix #315, since the problem there was synchronization overhead.

Speaking of which, was the ~20% performance improvement measured on WSL2 or on native Linux?

str4d:
> @str4d, was the ~20% performance improvement measured on WSL2 or on native Linux?

Native Linux (Ubuntu Server 20.04.3 LTS).

PR #448, opened 2022-01-04
[dev] Implement `Assignment::fill_from_row()` for `MockProver`.

therealyingtong:
This fills the unused rows of a table column with its default value in the `MockProver`.

Review this commit-by-commit, checking that the `bad_lookup` test:
- passes in the first commit (wrong behaviour),
- fails in the second commit (expected behaviour), and
- passes in the last commit after being fixed.

### DIFF
--- a/src/dev.rs
+++ b/src/dev.rs
@@ -996,7 +996,10 @@ mod tests {
     use super::{LookupFailure, MockProver, VerifyFailure};
     use crate::{
         circuit::{Layouter, SimpleFloorPlanner},
-        plonk::{Advice, Any, Circuit, Column, ConstraintSystem, Error, Selector, TableColumn},
+        plonk::{
+            Advice, Any, Circuit, Column, ConstraintSystem, Error, Expression, Selector,
+            TableColumn,
+        },
         poly::Rotation,
     };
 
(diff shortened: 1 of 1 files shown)


daira (approved):
utACK. Checked the behaviour on each commit is as described.

src/dev.rs
```diff
@@ -553,14 +553,18 @@ impl<F: Field + Group> Assignment<F> for MockProver<F> {
...
-        _: Option<Assigned<F>>,
+        to: Option<Assigned<F>>,
     ) -> Result<(), Error> {
         if !self.usable_rows.contains(&from_row) {
             return Err(Error::not_enough_rows_available(self.k));
         }
 
+        for row in self.usable_rows.clone().skip(from_row) {
+            self.assign_fixed(|| "", col, row, || to.ok_or(Error::Synthesis))?;
+        }
```
str4d, line 566: This commit results in the following test error:
```
thread 'dev::tests::bad_lookup' panicked at 'assertion failed: `(left == right)`
  left: `Err([Lookup { lookup_index: 0, location: InRegion { region: Region { index: 2, name: "Faulty synthesis" }, offset: 1 } }, Lookup { lookup_index: 0, location: OutsideRegion { row: 4 } }, Lookup { lookup_index: 0, location: OutsideRegion { row: 5 } }, Lookup { lookup_index: 0, location: OutsideRegion { row: 6 } }, Lookup { lookup_index: 0, location: OutsideRegion { row: 7 } }, Lookup { lookup_index: 0, location: OutsideRegion { row: 8 } }, Lookup { lookup_index: 0, location: OutsideRegion { row: 9 } }])`,
 right: `Err([Lookup { lookup_index: 0, location: InRegion { region: Region { index: 2, name: "Faulty synthesis" }, offset: 1 } }])`', src/dev.rs:1170:9
```

It's nice to see the improved error details from #431: we can clearly identify that the unintended errors are caused by lookup failures outside of any assigned region.

str4d (approved):
Tested ACK.

PR #451, opened 2022-01-08
fix(ci): wasm CI now uses the correct targets

kobigurk:


daira (approved):
utACK

kobigurk:
@daira I fixed getrandom to 0.2.3 to be able to set the `js` feature downstream. it's a bit of a ugly hack though. @str4d got a better idea maybe?

daira:
If we're changing the MSRV to 1.54.0 then [README.md](https://github.com/zcash/halo2#minimum-supported-rust-version) needs to change, and as mentioned there we need a minor version bump. But I want to get @ebfull's feedback on that first.

kobigurk:
> If we're changing the MSRV to 1.54.0 then [README.md](https://github.com/zcash/halo2#minimum-supported-rust-version) needs to change, and as mentioned there we need a minor version bump. But I want to get @ebfull's feedback on that first.

Good point - done. Let's see what @ebfull thinks

daira:
Version numbers are major.minor.micro[-beta.n], so bumping the micro version would mean 0.2.0-beta.1.

src/lib.rs
```diff
@@ -1,11 +1,11 @@
 //! # halo2
 
 #![cfg_attr(docsrs, feature(doc_cfg))]
-// Build without warnings on stable 1.51 and later.
+// Build without warnings on stable 1.54 and later.
 #![allow(unknown_lints)]
-// Disable old lint warnings until our MSRV is at least 1.51.
+// Disable old lint warnings until our MSRV is at least 1.54.
```
daira, line 6: I think this is not a correct comment change. We should revisit whether this is necessary on Rust 1.54+.

daira:
I could be mistaken, but I think that `halo2` depending on `getrandom 0.2.3` does not guarantee that `rand` will use `getrandom 0.2.3` (or later). It happens to do so at the moment because `rand` doesn't specify a version for its `getrandom` dependency (when `rand`'s `std` option is enabled).

kobigurk:
> I could be mistaken, but I think that `halo2` depending on `getrandom 0.2.3` does not guarantee that `rand` will use `getrandom 0.2.3` (or later). It happens to do so at the moment because `rand` doesn't specify a version for its `getrandom` dependency (when `rand`'s `std` option is enabled). But a future version of `rand` could change that.

It indeed doesn't. We could end up with different versions of `getrandom` if `rand` starts specifying an explicit different version.

str4d:
> @daira I fixed getrandom to 0.2.3 to be able to set the js feature downstream. it's a bit of a ugly hack though. @str4d got a better idea maybe?

This is currently unnecessary: downstream users just need to add a direct dependency on `getrandom` that includes that feature, and `cargo` will unify features across the build. I do this myself in several repos. However, I don't know how that behaves with the v2 resolver; I need to test it.

str4d:
The `bumpalo` problem is because `bumpalo 3.8` had an MSRV of 1.44, while `bumpalo 3.9` (released two weeks ago) has increased the MSRV to 1.54. `wasm-bindgen-backend` just depends on `bumpalo 3.0.0`, and thus `cargo` treats these as equivalent. I'm _really_ looking forward to `cargo` paying attention to MSRVs, but for now we have to deal with this kind of unexpected breakage...

In any case, the next MSRV bump will be to 1.56, not 1.54. I wanted to delay any MSRV bump until after the versioned crate releases needed for `zcashd 4.7.0`, which means I'm not inclined to merge this PR yet. However, if we instead add a manual pin for `bumpalo = "=3.8.0"` (with whatever WASM-specific target conditional is required) instead of the 1.5.4 bump, then I'd be okay to include this CI fix in the next versioned (beta) crate release.

Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "halo2"
-version = "0.1.0-beta.1"
+version = "0.2.0-beta.1"
```
str4d, line 3: This should not have been altered. We have not cut an `0.1.0` release yet, and altering the MSRV during development is perfectly acceptable.

Cargo.toml
```diff
@@ -40,6 +40,7 @@ group = "0.11"
 pasta_curves = "0.3"
 rand = "0.8"
 blake2b_simd = "1"
+getrandom = { version = "0.2.3", features = [ "js" ] }
```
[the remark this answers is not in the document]
str4d: I've opened a separate PR to implement a 1.51-compatible approach, and I see why this was included. It needs to be conditional though so we don't enable it for everyone; I'll do this in my PR.

str4d:
Replaced by #463 (where I've preserved the MSRV, and in doing so figured out several other ways this fix needed to be altered).

PR #463, opened 2022-01-20
Fix WASM builder in CI

str4d:


str4d:
It will take the union, but after this PR we only depend on `getrandom` from `halo2` in dev dependencies. While developing this crate `getrandom` will be compiled as a matter of course (due to cargo limitations), but downstream users of `halo2` will not depend on `getrandom` unless they explicitly bring it into their own tree (in which case they'd need to add that feature flag themselves anyway).

nuttycom (approved):
utACK

Cargo.toml
```diff
@@ -38,18 +38,29 @@ rayon = "1.5.1"
...
 # Developer tooling dependencies
 plotters = { version = "0.3.0", optional = true }
 tabbycat = { version = "0.1", features = ["attributes"], optional = true }
 
+[target.'cfg(target_arch = "wasm32")'.dependencies]
+# plotters depends on web-sys, which eventually depends on bumpalo 3. This dependency is
+# required because our MSRV is 1.51, but bumpalo 3.9 increased its MSRV to 1.54. We can
+# remove this once our MSRV is 1.54+ (and should do so, because currently this makes it a
+# required dependency even if the dev-graph feature flag is not enabled).
+bumpalo = "=3.8.0"
```
ebfull, line 53: This seems too strict, we should allow for e.g. `3.8.1`. I forgot how to do that with semver.

ebfull (changes requested):

Cargo.toml
ebfull: Looks like that's done with `~3.8.0`.

str4d: [This is now](https://github.com/zcash/halo2/compare/f68dfe4a88102cfb774a49c636725187e75dd3f6..9e63eeff1b2c4407bd9dfd714d5a0a44b275cbc9) `bumpalo = ">=3,<3.9.0"`.

ebfull (approved):

Issue #462, opened 2022-01-20
Refactor into workspace with `halo2` and `halo2_proofs`


str4d:
The current `halo2` crate will become the `halo2_proofs` crate, while `halo2` will be replaced by an empty library.

The example SHA-256 code will not be moved as part of this issue (so it will temporarily be uncompiled); it will be moved into `halo2_gadgets` as that is imported.

PR #465, opened 2022-01-20
Refactor into workspace with `halo2` and `halo2_proofs`

str4d:
Closes zcash/halo2#462.


halo2_proofs/Cargo.toml
```diff
@@ -0,0 +1,75 @@
+[package]
+name = "halo2_proofs"
+version = "0.1.0-beta.1"
```
str4d, line 3: I left this at the same version as `halo2` is currently published as, because I think it will be clearer if `halo2_proofs 0.1.0-beta.2` is the follow-on release from `halo2 0.1.0-beta.1`.

nuttycom (approved):
utACK

str4d:
Also note that `examples/sha256/` was left where it is, meaning that it is not currently being tested. This is temporary, just to avoid introducing an extra unnecessary move into the git history, as it will be moved into place when `halo2_gadgets` is merged in.

ebfull (approved):

PR #471, opened 2022-01-26
Fix handling of short chunks in `poly::Evaluator`

str4d:
Also includes a fix for `impl Debug for poly::Ast` which was never usable with typical evaluators.

### DIFF
--- a/halo2_proofs/src/poly/evaluator.rs
+++ b/halo2_proofs/src/poly/evaluator.rs
@@ -1,4 +1,5 @@
 use std::{
+    cmp,
     collections::{HashMap, HashSet},
     hash::{Hash, Hasher},
     marker::PhantomData,
@@ -179,6 +180,7 @@ impl<E, F: Field, B: Basis> Evaluator<E, F, B> {
 
         struct AstContext<'a, E, F: FieldExt, B: Basis> {
             domain: &'a EvaluationDomain<F>,
+            poly_len: usize,
             chunk_size: usize,
             chunk_index: usize,
             leaves: &'a HashMap<AstLeaf<E, B>, &'a [F]>,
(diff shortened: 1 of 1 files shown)


halo2_proofs/src/poly/evaluator.rs
```diff
@@ -442,10 +500,12 @@ impl BasisOps for Coeff {
 
     fn linear_term<F: FieldExt>(
         _: &EvaluationDomain<F>,
+        poly_len: usize,
         chunk_size: usize,
         chunk_index: usize,
         scalar: F,
     ) -> Vec<F> {
+        let chunk_size = cmp::min(chunk_size, poly_len - chunk_size * chunk_index);
         let mut chunk = vec![F::zero(); chunk_size];
```
daira, line 509: Sorry, the change I suggested we make here is actually wrong. `chunk_size` on line 513 should be the *original* chunk size, because if `chunk_index == 1` then what we also want to check is whether chunk 0 (not chunk 1) was of size 1.

```suggestion
        let mut chunk = vec![F::zero(); cmp::min(chunk_size, poly_len - chunk_size * chunk_index)];
```

Also maybe add to the comment to explain this.

halo2_proofs/src/poly/evaluator.rs
```diff
@@ -541,3 +613,61 @@ impl BasisOps for ExtendedLagrangeCoeff {
...
+    use crate::{
+        multicore,
+        poly::{Coeff, EvaluationDomain, ExtendedLagrangeCoeff, LagrangeCoeff},
+    };
+
+    #[test]
+    fn short_chunk_regression_test() {
+        // Pick the smallest polynomial length that is guaranteed to produce a short chunk
+        // on this machine.
+        let k = match (1..31)
```
daira, line 633: Maybe this should be smaller so that we prefer to not run the test over testing with an unreasonably large circuit.
```suggestion
        let k = match (1..16)
```

halo2_proofs/src/poly/evaluator.rs
```diff
@@ -541,3 +613,61 @@ impl BasisOps for ExtendedLagrangeCoeff {
...
+
+            // Construct ASTs that trigger the bug: they need either a constant or linear
+            // term in a left-hand slot.
+            for ast in iter::empty()
+                .chain(Some(Ast::ConstantTerm(pallas::Base::zero())))
+                .chain(Some(Ast::LinearTerm(pallas::Base::zero())))
+            {
+                // With the bug present, this panics.
+                let _ = evaluator.evaluate(&ast, &domain);
+            }
```
daira, line 666:
```suggestion
            // With the bug present, these will panic.
            let _ = evaluator.evaluate(&Ast::ConstantTerm(pallas::Base::zero()), &domain);
            let _ = evaluator.evaluate(&Ast::LinearTerm(pallas::Base::zero()), &domain);
```

daira (changes requested):
Fix needed in `linear_term` for `Coeff`.

daira (approved):
utACK

PR #472, opened 2022-01-26
halo2_proofs: Improve `plonk::verify_proof` API

str4d:
Previously `plonk::verify_proof` took an `MSM` as an argument, to enable
batch verification. However, this also required that it take a source of
randomness in order to enforce separation of proofs within a batch. This
made single-proof verification unnecessarily non-deterministic.

We now have a `VerificationStrategy` trait encapsulating the necessary
details, and separate `SingleVerifier` and `BatchVerifier` structs for
the specific variants. Proof verifiers no longer need to create and
manage the `MSM` themselves, and single-proof verifiers no longer need
to supply a source of randomness.

Co-authored-by: Sean Bowe <sean@electriccoin.co>

### DIFF
--- a/halo2_proofs/src/plonk/verifier.rs
+++ b/halo2_proofs/src/plonk/verifier.rs
@@ -14,21 +14,111 @@ use crate::poly::{
 };
 use crate::transcript::{read_n_points, read_n_scalars, EncodedChallenge, TranscriptRead};
 
+/// Trait representing a strategy for verifying Halo 2 proofs.
+pub trait VerificationStrategy<'params, C: CurveAffine> {
+    /// The output type of this verification strategy after processing a proof.
+    type Output;
+
+    /// Obtains an MSM from the verifier strategy and yields back the strategy's
+    /// output.
+    fn process<E: EncodedChallenge<C>>(
+        self,
+        f: impl FnOnce(MSM<'params, C>) -> Result<Guard<'params, C, E>, Error>,
+    ) -> Result<Self::Output, Error>;
+}
+
+/// A verifier that checks a single proof at a time.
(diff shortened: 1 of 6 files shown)


halo2_proofs/src/plonk/verifier.rs
```diff
@@ -14,21 +14,112 @@ use crate::poly::{
...
 use crate::transcript::{read_n_points, read_n_scalars, EncodedChallenge, TranscriptRead};
 
+/// Trait representing a strategy for verifying Halo 2 proofs.
+pub trait VerificationStrategy<'params, C: CurveAffine> {
+    /// The output type of this verification strategy after processing a proof.
+    type Output;
+
+    /// Obtains an MSM from the verifier strategy and yields back the strategy's
+    /// output.
+    #[must_use]
```
daira, line 24:
```suggestion
```
`#[must_use]` is implicit because this returns a `Result`.

halo2_proofs/src/plonk/verifier.rs
```diff
@@ -14,21 +14,112 @@ use crate::poly::{
...
+        BatchVerifier {
+            msm: MSM::new(params),
+            rng,
+        }
+    }
+
+    /// Finalizes the batch and checks its validity.
+    ///
+    /// Returns `false` if *some* proof was invalid. The caller will need to re-process
+    /// the proofs separately to identify specific failing proofs.
```
daira, line 82:
```suggestion
    /// Returns `false` if *some* proof was invalid. If the caller needs to identify
    /// specific failing proofs, it must re-process the proofs separately.
```

daira (approved):
utACK

Issue #477, opened 2022-02-01
Add a helper for checking `MockProver::verify` results


str4d:
Currently, the way someone checks the result of `MockProver::verify` is with something like:
```rust
assert_eq!(
    prover.verify(),
    Err(vec![VerifyFailure::Lookup {
        lookup_index: 0,
        location: FailureLocation::InRegion {
            region: (2, "Faulty synthesis").into(),
            offset: 1,
        }
    }])
);
```

However, if this fails, all the developer sees is the structural difference between the two, which uses the `Debug` impl of `VerifyFailure`. This is useful if the dev knows what they are doing, but it would be much better if we had a better error renderer, that either leveraged the `Display` impl of `VerifyFailure`, or implemented an even more verbose / instructive error (like `rustc` does 🦀❤️).

What I'm imagining is a `MockProver::verify_valid`, that calls `MockProver::verify` and then if it returns errors, it uses `eprintln!()` to pretty-print an explanation of the errors, and then `panic!()`s to indicate the failure (effectively inlining an `assert_eq!(self.verify(), Ok(()))`).

str4d:
The motivation for this was the error from #478. What the user saw was:
```
  left: `Err([ConstraintNotSatisfied { constraint: Constraint { gate: Gate { index: 0, name: "constraint decompose" }, index: 0, name: "" }, row: 1, cell_values: [(VirtualCell { name: "", column: Column { column_type: Advice, index: 0 }, rotation: 0 }, "0x5"), (VirtualCell { name: "", column: Column { column_type: Advice, index: 0 }, rotation: 1 }, "0x5")] }])`,
 right: `Ok(())
```
while what would actually be helpful is something like:
```
error: constraint not satisfied
   Row    Advice
  Offset   0
     0   | a |    <-- { Gate "constraint decompose", constraint "foo"
     1   | b |        [ b + b * 0x100 + b * 0x10000 + b * 0x100_0000 - a = 0

  Assigned cell values:
      a = 0x5
      b = 0x5
```

str4d:
https://rustc-dev-guide.rust-lang.org/diagnostics.html#error-messages documents how `rustc` builds its useful error reporting.

str4d:
We probably wouldn't want anything nearly as fancy (to begin with), but here's the emitter for `rustc`: https://github.com/rust-lang/rust/blob/master/compiler/rustc_errors/src/emitter.rs

daira:
[...]

PR #480, opened 2022-02-01
Add `MockProver::assert_satisfied` with pretty-printed failures

str4d:
This is equivalent to `assert_eq!(mock_prover.verify(), Ok(()))`, but pretty-prints the verification failures instead of debug-printing them.

This PR has special handling for the following failure cases:
- `VerifyFailure::CellNotAssigned`
- `VerifyFailure::ConstraintNotSatisfied`
- `VerifyFailure::Lookup`

All other failure cases print the `Display` impl.

Closes zcash/halo2#477.


str4d:
Rebased to fix merge conflicts in the changelog.

therealyingtong (approved):
utACK 🦜

therealyingtong:
Idea: we could add pretty-printing as a feature of MockProver, not just in case of failures.

halo2_proofs/src/dev.rs
```diff
@@ -1012,6 +803,25 @@ impl<F: FieldExt> MockProver<F> {
             Err(errors)
         }
     }
+
+    /// Panics if the circuit being checked by this `MockProver` is not satisfied.
+    ///
+    /// Any verification failures will be pretty-printed to stderr before the function
+    /// panics.
+    ///
+    /// This method is equivalent to:
```
daira, line 812:
```suggestion
    /// Apart from the stderr output, this method is equivalent to:
```

halo2_proofs/src/dev/failure.rs
```diff
@@ -0,0 +1,530 @@
...
+        row: usize,
+    },
+}
+
+impl fmt::Display for FailureLocation {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Self::InRegion { region, offset } => write!(f, "in {} at offset {}", region, offset),
+            Self::OutsideRegion { row } => {
+                write!(f, "on row {}", row)
```
daira, line 40:
```suggestion
                write!(f, "outside any region on row {}", row)
```

halo2_proofs/src/dev/failure/emitter.rs
```diff
@@ -0,0 +1,164 @@
...
+                .unwrap()
+                .get(&(Any::Instance, column).into())
+                .unwrap()
+                .clone()
+        },
+        &|a| {
+            if a.contains(' ') {
+                format!("-({})", a)
+            } else {
+                format!("-{}", a)
```
daira, line 140: This doesn't contain space, so technically you could end up with `--x` instead of `-(-x)`. It probably doesn't matter; I can't see a case where the result would be ambiguous.

daira (approved):
utACK

daira:
re-utACK 69c138c25c0969b2ece2cf2f0bba28f7b4f0bec1

Issue #457, opened 2021-10-31
Upgrade from deprecated `bigint` crate to `uint`


teor2345:
The `bigint` crate is deprecated, and hasn't been updated for over a year:
https://crates.io/crates/bigint

Parity recommends using `uint` instead:
https://crates.io/crates/uint

Zebra has done this upgrade, and it is pretty simple.
Most APIs are a drop-in replacement.

PR #481, opened 2022-02-01
halo2_gadgets: Migrate from bigint to uint

str4d:
Closes zcash/halo2#457.

### DIFF
--- a/halo2_gadgets/src/ecc/chip/mul.rs
+++ b/halo2_gadgets/src/ecc/chip/mul.rs
@@ -8,7 +8,6 @@ use std::{
     ops::{Deref, Range},
 };
 
-use bigint::U256;
 use ff::PrimeField;
 use halo2_proofs::{
     arithmetic::FieldExt,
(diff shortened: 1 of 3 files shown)


halo2_gadgets/src/ecc/chip/mul.rs
```diff
@@ -425,6 +425,10 @@ impl<F: FieldExt> Deref for Z<F> {
 }
 
 fn decompose_for_scalar_mul(scalar: Option<&pallas::Base>) -> Vec<Option<bool>> {
+    construct_uint! {
+        struct U256(4);
+    }
+
```
nuttycom, line 431: I'm surprised that this is only needed locally! Would it make sense at all for this to be a top-level declaration?

nuttycom (approved):
utACK

halo2_gadgets/src/ecc/chip/mul.rs
str4d: We only require this because we need to perform a non-reducing addition internal to the decomposition. It's a specific artifact of how variable-base scalar multiplication is implemented for this chip; it does not need to be part of the public API, or even be interoperable with other uses of it elsewhere in `halo2_gadgets`. That being said, if we do ever find ourselves in need of such interoperability, we can refactor at that time.

PR #488, opened 2022-02-04
CI: Fix rustdoc rendering to handle workspace

str4d:
`cargo rustdoc` only works for a single package. To render docs for
a workspace while passing config options to `rustdoc`, we need to use
the `RUSTDOCFLAGS` environment variable.

We also add several other flags to handle the switch to `cargo doc`:
- `--no-deps` ensures we only build packages in the workspace.
- `--enable-index-page` (unstable) adds a landing page showing the list
  of rendered crate docs.

### DIFF
--- a/.github/workflows/book.yml
+++ b/.github/workflows/book.yml
@@ -32,8 +32,10 @@ jobs:
       - name: Build latest rustdocs
         uses: actions-rs/cargo@v1
         with:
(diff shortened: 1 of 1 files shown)


str4d:
Self-ACK; this only alters the book rendering workflow, which we can't test live until this reaches `main`. I have tested the commands locally.

PR #489, opened 2022-02-04
CI: Pin nightly version in book workflow

str4d:
We require nightly Rust to build the documentation, but latest nightly
has a bug where `--no-deps` causes an ICE.


str4d:
Self-ACK, only affects the book workflow, and we can't test this until it lands in `main`.

PR #492, opened 2022-02-05
Nix support

ursi:
This adds a flake with a nix shell for using cargo and running the example that generates the circuit image, and packages for builds of each workspace.


ursi:
This was supposed to be a PR into our fork, I didn't realize it defaulted to here.

PR #497, opened 2022-02-10
[WIP] Test vectors for poseidon MDS matrix

defuse:


halo2_gadgets/src/primitives/poseidon/mds.rs
```diff
@@ -130,6 +130,11 @@ mod tests {
 
         const T: usize = 3;
         let mut grain = Grain::new(super::super::grain::SboxType::Pow, T as u16, 8, 56);
+
+        // Skip over the bits the reference implementation uses to generate the round constants
+        for _ in 0..103020 {
+            grain.next();
+        }
```
daira, line 137: This also applies to the `poseidon_mds` test. Factor out the MDS generation into a test helper (or maybe fixing the TODO will deal with that).

Issue #500, opened 2022-02-14
IPA check is needlessly complicated


ebfull:
We currently check

```
P' + \sum([u_j^{-1}] L_j) + \sum([u_j] R_j) + [-c] G'_0 + [-cbz] U + [c - f] W = 0
```

which assumes that G'_0 (part of the accumulator) has a blinding factor of 1, which is then compensated for in the equation. The original purpose of this was to ensure that no matter what challenges are chosen, every element is not the point at infinity. 

However, there's no choice of _valid_ challenges anymore that would cause G'_0 to be the point at infinity, let alone any of its coefficients being zero. We should change this, which has the added benefit of corresponding exactly with what is written in the book.

PR #502, opened 2022-02-14
halo2_proofs: change IPA check equation to match the book

ebfull:
Closes https://github.com/zcash/halo2/issues/500

### DIFF
--- a/halo2_proofs/src/poly/commitment/verifier.rs
+++ b/halo2_proofs/src/poly/commitment/verifier.rs
@@ -35,13 +35,12 @@ impl<'a, C: CurveAffine, E: EncodedChallenge<C>> Guard<'a, C, E> {
     pub fn use_challenges(mut self) -> MSM<'a, C> {
         let s = compute_s(&self.u, self.neg_c);
         self.msm.add_to_g_scalars(&s);
-        self.msm.add_to_w_scalar(self.neg_c);
 
         self.msm
     }
 
     /// Lets caller supply the purported G point and simply appends
-    /// [-a] G to return an updated MSM.
+    /// [-c] G to return an updated MSM.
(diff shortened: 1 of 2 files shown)


daira (approved):
utACK modulo @str4d's comment.

halo2_proofs/src/poly/commitment/verifier.rs
```diff
@@ -57,9 +56,7 @@ impl<'a, C: CurveAffine, E: EncodedChallenge<C>> Guard<'a, C, E> {
     pub fn compute_g(&self) -> C {
```
str4d, line 56: The comment for this function is now incorrect.

halo2_proofs/src/poly/commitment/verifier.rs
```diff
@@ -118,15 +115,10 @@ pub fn verify_proof<'a, C: CurveAffine, E: EncodedChallenge<C>, T: TranscriptRea
     // equals (given b = \mathbf{b}_0, and the prover's values c, f),
     // the right-hand side
     //   = [c] (G'_0 + [b * z] U) + [f] W
-    // except that we wish for the prover to supply G'_0 as Commit(g(X); 1) so
-    // we must substitute G'_0 with G'_0 - W to get
-    //   = [c] ((G'_0 - W) + [b * z] U) + [f] W
-    //   = [c] G'_0 + [-c] W + [cbz] U + [f] W
-    //   = [c] G'_0 + [cbz] U + [f - c] W
     // and then subtracting the right-hand side from both sides
     // to get
```
str4d, line 119:
```suggestion
    // Subtracting the right-hand side from both sides, we get
```

str4d (approved):
utACK modulo comment changes. I confirmed that this now matches the protocol document.

Issue #449, opened 2022-01-04
`ViewingKey` serialization is not round-trip compatible


str4d:
in https://github.com/appliedzkp/zkevm-circuits/blob/main/circuit-benchmarks/src/evm_circuit.rs#L84 using the vk in memory works fine, but saving&reloading the VK (not PK mentioned in this issue) doesn't work. (and https://github.com/zcash/halo2/blob/main/examples/sha256/benches.rs works good)
Would you happen to have a chance to know any clue?

_Originally posted by @HAOYUatHZ in https://github.com/zcash/halo2/issues/443#issuecomment-1004612419_

HAOYUatHZ:
a bit more of the context:

1. we switch to pairing bn254 instead of pasta
2. the reported error is `virtual selectors are removed during optimization`

ursi:
the following code can be added to the end of the `main` function in [simple-circuit.rs](https://github.com/zcash/halo2/blob/9617336d46a18f3949260731d02bd711fba9d863/halo2_proofs/examples/simple-example.rs) to demonstrate this error.

```rust
use halo2_proofs::{
    plonk::{create_proof, keygen_pk, keygen_vk, VerifyingKey},
    poly::commitment::Params,
    transcript::Blake2bWrite,
};
use pasta_curves::{vesta, EqAffine};
use rand_core::OsRng;

let params: Params<EqAffine> = halo2_proofs::poly::commitment::Params::new(k);

let vk = keygen_vk(&params, &circuit).unwrap();

let mut vk_buffer = vec![];
vk.write(&mut vk_buffer).unwrap();
let vk =
    VerifyingKey::<EqAffine>::read::<_, MyCircuit<Fp>>(&mut &vk_buffer[..], &params).unwrap();

let pk = keygen_pk(&params, vk, &circuit).unwrap();
let mut transcript = Blake2bWrite::<_, vesta::Affine, _>::init(vec![]);

create_proof(
    &params,
    &pk,
    &[circuit],
    &[&[&[c]]],
    &mut OsRng,
    &mut transcript,
);
```

ursi:
Possible helpful info:
If you change [this line](https://github.com/zcash/halo2/blob/9617336d46a18f3949260731d02bd711fba9d863/halo2_proofs/examples/simple-example.rs#L116) to `vec![(lhs * rhs - out)]` (and comment out the assertions from the MockProver section), you will no longer get any errors.

ursi:
[...]

PR #521, opened 2022-03-16
Remove selector_map from pinned verification key and remove VerificationKey serialization

ebfull:
Two problems this PR is addressing.

1. The verification key deserialization logic fails to correctly deserialize when virtual selectors exist because it cannot reconstruct the gate expressions and the number of combined selector columns -- that information only exists right now during keygen. (#449) So, this PR removes it for the time being until we can write a working implementation later.
2. The pinned verification key stores a `selector_map` *which isn't used anywhere* and doesn't describe anything that isn't already fully determined by the rest of the key. We want to remove this from the pinned verification key so that we don't have to reconstruct this in the future. (Note that this makes proofs incompatible with previous versions because it changes Fiat-Shamir transcript initialization.)


daira (approved):
utACK

therealyingtong (approved):
utACK

PR #522, opened 2022-03-17
[ECC gadget] Distinguish base field element case in variable-base scalar mul.

### LINKED ISSUE #483: Refactor variable-base scalar mul to distinguish the "base field element" case
For fixed-base scalar multiplication, we need to prepare the base point differently for three cases:
- Full-width scalar
- Short signed scalar
- Base field element used as scalar (where the base field fits into the scalar field).

The gadget that handles this has specific types covering these three cases

We also implement variable-base scalar multiplication. However, in Orchard we only needed to handle one case, where `ivk` was the scalar - but it is in fact a base field element that we treat as a scalar, because it's an output of Sinsemilla. Outside the circuit, we convert it to a Pallas scalar and use the normal scalar multiplication APIs. However, inside the circuit we keep it as a Pallas base field element right up to the multiplication point.

The upshot is that `halo2_gadgets` only currently provides variable-base scalar multiplication for scalars that are represented canonically as base field elements. This is going to trip people up, especially because the chip is documented as supporting the calculation of [α]T for α∈[0,q), i.e. a full-width scalar.

We should refactor the variable-base scalar multiplication API, to have two separate constructors: one for base field elements used as scalars, and one for full-width scalars witnessed in chunks.

therealyingtong:
Closes #483.

### DIFF
--- a/halo2_gadgets/CHANGELOG.md
+++ b/halo2_gadgets/CHANGELOG.md
@@ -11,6 +11,23 @@ and this project adheres to Rust's notion of
 - `halo2_gadgets::utilities`:
   - `FieldValue` trait.
   - `RangeConstrained` newtype wrapper.
+- `halo2_gadgets::ecc`:
+  - `EccInstructions::witness_scalar_var` API to witness a full-width scalar
+    used in variable-base scalar multiplication.
+  - `BaseFitsInScalarInstructions` trait that can be implemented for a curve
+    whose base field fits into its scalar field. This provides a method
+    `scalar_var_from_base` that converts a base field element that exists as
+    a variable in the circuit, into a scalar to be used in variable-base
+    scalar multiplication.
+  - `ScalarVar::new` and `ScalarVar::from_base` gadget APIs.
+- `halo2_gadgets::ecc::chip`:
+  - `ScalarVar` enum with `BaseFieldElem` and `FullWidth` variants. `FullWidth`
+    is unimplemented for `halo2_gadgets v0.1.0`.
+
+### Changed
+- `halo2_gadgets::ecc`:
+  - `EccInstructions::mul` now takes a `Self::ScalarVar` as argument, instead
+    of assuming that the scalar fits in a base field element `Self::Var`.
 
 ## [0.1.0-beta.3] - 2022-04-06
 ### Changed
--- a/halo2_gadgets/CHANGELOG.md
+++ b/halo2_gadgets/CHANGELOG.md
@@ -12,6 +12,8 @@ and this project adheres to Rust's notion of
   - `FieldValue` trait.
   - `RangeConstrained` newtype wrapper.
 - `halo2_gadgets::ecc`:
+  - `EccInstructions::ScalarVar` is now treated as a full-width scalar, instead
+    of being restricted to a base field element.
   - `EccInstructions::witness_scalar_var` API to witness a full-width scalar
     used in variable-base scalar multiplication.
   - `BaseFitsInScalarInstructions` trait that can be implemented for a curve
@@ -28,6 +30,8 @@ and this project adheres to Rust's notion of
 - `halo2_gadgets::ecc`:
   - `EccInstructions::mul` now takes a `Self::ScalarVar` as argument, instead
     of assuming that the scalar fits in a base field element `Self::Var`.
+- `halo2_gadgets::ecc::chip`:
+  - `ScalarKind` has been renamed to `FixedScalarKind`.
 
 ## [0.1.0-beta.3] - 2022-04-06
 ### Changed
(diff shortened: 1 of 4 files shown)


halo2_gadgets/src/ecc.rs
```diff
@@ -16,17 +16,7 @@ pub mod chip;
...
-    ///
-    /// It is not true in general that a scalar field element fits in a curve's
-    /// base field, and in particular it is untrue for the Pallas curve, whose
-    /// scalar field `Fq` is larger than its base field `Fp`.
-    ///
-    /// However, the only use of variable-base scalar mul in the Orchard protocol
-    /// is in deriving diversified addresses `[ivk] g_d`,  and `ivk` is guaranteed
-    /// to be in the base field of the curve. (See non-normative notes in
-    /// https://zips.z.cash/protocol/nu5.pdf#orchardkeycomponents.)
+    /// Variable representing a scalar used in variable-base scalar mul.
```
str4d, line 19:
```suggestion
    /// Variable representing a scalar used in variable-base scalar mul.
    ///
    /// This type is treated as a full-width scalar. However, if `Self` implements
    /// [`BaseFitsInScalarInstructions`] then this may also be constructed from an element
    /// of the base field.
```

halo2_gadgets/src/ecc.rs
```diff
@@ -259,7 +268,7 @@ impl<C: CurveAffine, EccChip: EccInstructions<C>> NonIdentityPoint<C, EccChip> {
     pub fn mul(
         &self,
         mut layouter: impl Layouter<C::Base>,
-        by: &EccChip::Var,
+        by: &EccChip::ScalarVar,
     ) -> Result<(Point<C, EccChip>, ScalarVar<C, EccChip>), Error> {
```
str4d, line 272:
```suggestion
        by: &mut ScalarVar<C, EccChip>,
    ) -> Result<Point<C, EccChip>, Error> {
```

or, if we take @ebfull's preferred [move-based API approach](https://github.com/zcash/halo2/issues/483#issuecomment-1072587613):
```suggestion
        by: ScalarVar<C, EccChip>,
    ) -> Result<(Point<C, EccChip>, ScalarVar<C, EccChip>), Error> {
```

Both of these require adding new constructors to `ScalarVar`:
```rust
impl<C: CurveAffine, EccChip: EccInstructions<C>> ScalarVar<C, EccChip> {
    /// Witnesses the given full-width scalar.
    ///
    /// Depending on the `EccChip` implementation, this may either witness the scalar
    /// immediately, or delay witnessing until its first use in [`NonIdentityPoint::mul`].
    pub fn new(
        chip: EccChip,
        mut layouter: impl Layouter<C::Base>,
        value: Option<C::Scalar>,
    ) -> Self {
        let scalar = chip.witness_scalar_var(&mut layouter, value);
        scalar.map(|inner| ScalarVar { chip, inner })
    }
}

impl<C: CurveAffine, EccChip: BaseFitsInScalarInstructions<C>> ScalarVar<C, EccChip> {
    /// Constructs a scalar from an existing base-field element.
    pub fn from_base(
        chip: EccChip,
        mut layouter: impl Layouter<C::Base>,
        base: &Self::Var,
    ) -> Self {
        let scalar = chip.scalar_var_from_base(&mut layouter, value);
        scalar.map(|inner| ScalarVar { chip, inner })
    }
}
```

We also need to check that the scalar and point chips match.

str4d (changes requested):
Generally LGTM

halo2_gadgets/src/ecc.rs
```diff
@@ -74,6 +64,13 @@ pub trait EccInstructions<C: CurveAffine>:
         value: Option<C>,
     ) -> Result<Self::NonIdentityPoint, Error>;
 
+    /// Witnesses a full-width scalar to be used in variable-base multiplication.
+    fn witness_scalar_var(
```
str4d, line 68: Add this (and other additions / changes) to the changelog.

str4d (changes requested):

halo2_gadgets/CHANGELOG.md
```diff
@@ -11,6 +11,27 @@ and this project adheres to Rust's notion of
 - `halo2_gadgets::utilities`:
   - `FieldValue` trait.
   - `RangeConstrained` newtype wrapper.
+- `halo2_gadgets::ecc`:
+  - `EccInstructions::ScalarVar` is now treated as a full-width scalar, instead
+    of being restricted to a base field element.
```
str4d, line 16:
```suggestion
```

str4d (approved):
re-utACK 5a543f93189b61a763bd87dce353ecb5fb529b67

ebfull:
post-hoc ACK

daira (commented):
Post-hoc utACK

Issue #506, opened 2022-02-17
Docs not rendering on docs.rs


ebfull:
I guess we need to put `katex-header.html` into each crate subdir so it ends up in the package that docs.rs needs.

PR #524, opened 2022-03-17
Fix docs.rs build

str4d:
The published source code for each package needs to include the required
header file, and the path to that header file needs to be relative to
the package source (not the repository source). We therefore need to
have the header file present in each workspace package.

Closes zcash/halo2#506.


nuttycom (approved):
utACK, looks right.

str4d:
This PR will be tested by the `halo2_proofs 0.1.0-beta.3` release; if the doc build for that release fails, we can try again.

PR #473, opened 2022-01-27
Import `halo2_gadgets` as extracted from `orchard`

str4d:
The steps I took to produce this:
- Created a branch starting from https://github.com/zcash/orchard/commit/159ab53da59ba762794862dc4569c0273c67ed1c.
- Prepared the `halo2_gadgets` crate by:
  - Deleting Orchard-specific content.
  - Moving files into a `halo2_gadgets/` subdirectory (or otherwise ensuring they wouldn't collide with files in this repo).
- Created this branch starting from https://github.com/zcash/halo2/commit/bb56139414332da8964f92c4d5332af5848fe706.
- Added my local `zcash/orchard` repo as a git remote.
- `git merge --allow-unrelated-histories orchard/gadgets-extraction`
  - This brings the history of `zcash/orchard` into this repository, in a way that allows git history to function intuitively.
- Performed post-merge integration of `halo2_gadgets` into the workspace.

Best reviewed by-commit, starting from c2866beb4781790f1cf6c644b8b632bdd8ab843e for obvious reasons. The relevant diff is https://github.com/zcash/halo2/pull/473/files/159ab53da59ba762794862dc4569c0273c67ed1c..5312343e6d68b1bdd141337b52a5b73cce752bff

Part of zcash/orchard#183.

Issue #334, opened 2021-07-16
Optimisation: Enable layouter to assign "equality-only" cells or regions


str4d:
We have a bunch of places in the Orchard Action circuit where we assign a value (or several values) to one or more cells in a region, and then only use those cells for equality constraints (into other parts of the circuit).

We should add a `Layouter::assign_equality_region` API to enable explicitly optimizing this case. The layouter would track all cells in this region as "equality-only", with no real cell location, and the region itself would never appear in the circuit. Then, when the cells from that region are used:

- If this is the first time the cell is used in an equality constraint, the layouter assigns the value instead of equality-constraining, and marks that cell as "assigned".
- If this is the second or later time, the layouter adds an equality constraint to the first assigned cell.
- If this is anything other than an equality constraint, circuit synthesis returns an error.

PR #529, opened 2022-03-22
Endoscaling instructions and chip.

### LINKED ISSUE #249: Implement circuit logic for handling public inputs
This will have two components:
- Public input packing: Creating the commitment to the public inputs efficiently on one curve.
- Unpacking the public inputs on the other curve.

therealyingtong:
Partially closes #249. Based on #473.

### DIFF
--- a/halo2/src/recursion.rs
+++ b/halo2/src/recursion.rs
@@ -1,0 +1,3 @@
+//! Recursive infrastructure for halo2 circuits.
+
+pub mod endoscale;
--- a/halo2/src/recursion/endoscale.rs
+++ b/halo2/src/recursion/endoscale.rs
@@ -7,6 +7,7 @@ use halo2_proofs::{
 };
 use pasta_curves::arithmetic::CurveAffine;
 
+mod chip;
 pub mod primitive;
 
 /// Instructions to map bitstrings to and from endoscalars.
--- a/halo2/src/recursion/endoscale.rs
+++ b/halo2/src/recursion/endoscale.rs
@@ -1,3 +1,101 @@
 //! Gadget for endoscaling.
 
+use ff::PrimeFieldBits;
+use halo2_proofs::{
+    circuit::{AssignedCell, Layouter},
+    plonk::Error,
+};
+use pasta_curves::arithmetic::CurveAffine;
+
 pub mod primitive;
+
+/// Instructions to map bitstrings to and from endoscalars.
+pub trait EndoscaleInstructions<C: CurveAffine>
+where
+    C::Base: PrimeFieldBits,
+{
+    /// A bitstring to be used in endoscaling.
+    type Bitstring: std::fmt::Debug + Clone;
+    /// The maximum number of bits that can be represented by [`Self::Bitstring`].
+    const MAX_BITSTRING_LENGTH: usize;
+
+    /// Witnesses a slice of bools as a vector of [`Self::Bitstring`]s.
+    fn witness_bitstring(bits: &[bool]) -> Vec<Self::Bitstring>;
+
+    /// Computes commitment (Alg 1) to a variable-length bitstring using the endoscaling
+    /// algorithm. Uses a fixed base.
+    ///
+    /// The bitstring is decomposed into pairs of bits using a running sum outside of
+    /// this gadget.
+    ///
+    /// # Panics
+    /// Panics if `NUM_BITS` is larger than `MAX_BITSTRING_LENGTH`.
+    /// Panics if `NUM_BITS` is an odd number.
+    #[allow(clippy::type_complexity)]
+    fn endoscale_fixed_base<L: Layouter<C::Base>, const NUM_BITS: usize, const NUM_WINDOWS: usize>(
+        &self,
+        layouter: L,
+        base: C,
+        bitstring: &Self::Bitstring,
+    ) -> Result<
+        (
+            AssignedCell<C::Base, C::Base>,
+            AssignedCell<C::Base, C::Base>,
+        ),
+        Error,
+    >;
+
+    /// Computes commitment (Alg 1) to a variable-length bitstring using the endoscaling
+    /// algorithm. Uses a variable base witnessed elsewhere in the circuit.
+    ///
+    /// The bitstring is decomposed into pairs of bits using a running sum outside of
+    /// this gadget.
+    ///
+    /// # Panics
+    /// Panics if `NUM_BITS` is larger than `MAX_BITSTRING_LENGTH`.
+    /// Panics if `NUM_BITS` is an odd number.
+    #[allow(clippy::type_complexity)]
+    fn endoscale_var_base<L: Layouter<C::Base>, const NUM_BITS: usize, const NUM_WINDOWS: usize>(
+        &self,
+        layouter: L,
+        base: (
+            AssignedCell<C::Base, C::Base>,
+            AssignedCell<C::Base, C::Base>,
+        ),
+        bitstring: &Self::Bitstring,
+    ) -> Result<
+        (
+            AssignedCell<C::Base, C::Base>,
+            AssignedCell<C::Base, C::Base>,
+        ),
+        Error,
+    >;
+
+    /// Computes endoscalar (Alg 2) mapping to a variable-length bitstring using
+    /// the endoscaling algorithm.
+    ///
+    /// The bitstring is decomposed into windows using a running sum outside of
+    /// this gadget.
+    ///
+    /// # Panics
+    /// Panics if `NUM_BITS` is larger than `MAX_BITSTRING_LENGTH`.
+    /// Panics if `NUM_BITS` is an odd number.
+    fn endoscale_scalar<L: Layouter<C::Base>, const NUM_BITS: usize, const NUM_WINDOWS: usize>(
+        &self,
+        layouter: L,
+        bitstring: &Self::Bitstring,
+    ) -> Result<AssignedCell<C::Base, C::Base>, Error>;
+
+    /// Check that a witnessed bitstring corresponds to a range of endoscalars
+    /// provided as public inputs.
+    ///
+    /// # Panics
+    /// Panics if `NUM_BITS` is larger than `MAX_BITSTRING_LENGTH`.
+    /// Panics if `NUM_BITS` is an odd number.
+    fn recover_bitstring<L: Layouter<C::Base>, const NUM_BITS: usize, const NUM_WINDOWS: usize>(
+        &self,
+        layouter: L,
+        bitstring: &Self::Bitstring,
+        pub_input_rows: [usize; NUM_WINDOWS],
+    ) -> Result<(), Error>;
+}
--- a/halo2/src/recursion/endoscale.rs
+++ b/halo2/src/recursion/endoscale.rs
@@ -1,0 +1,3 @@
+//! Gadget for endoscaling.
+
+pub mod primitive;
--- a/halo2/src/recursion/endoscale/chip.rs
+++ b/halo2/src/recursion/endoscale/chip.rs
@@ -817,3 +817,326 @@ where
         )
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use crate::recursion::endoscale::primitive::endoscale_scalar;
+
+    use super::{EndoscaleConfig, EndoscaleInstructions, TableConfig};
+    use ff::PrimeFieldBits;
+    use halo2_gadgets::utilities::{lebs2ip, UtilitiesInstructions};
+    use halo2_proofs::{
+        arithmetic::CurveAffine,
+        circuit::{AssignedCell, Layouter, SimpleFloorPlanner},
+        plonk::{Circuit, ConstraintSystem, Error},
+        poly::commitment::Params,
+    };
+    use pasta_curves::{pallas, vesta};
+
+    use std::{convert::TryInto, marker::PhantomData};
+
+    #[derive(Default)]
+    struct BaseCircuit<
+        C: CurveAffine,
+        const NUM_BITS: usize,
+        const NUM_BITS_DIV2: usize,
+        const NUM_BITS_DIV_K_CEIL: usize,
+    >
+    where
+        C::Base: PrimeFieldBits,
+    {
+        bitstring: Option<[bool; NUM_BITS]>,
+        pub_input_rows: Option<[usize; NUM_BITS_DIV_K_CEIL]>,
+        _marker: PhantomData<C>,
+    }
+
+    impl<
+            C: CurveAffine,
+            const NUM_BITS: usize,
+            const NUM_BITS_DIV2: usize,
+            const NUM_BITS_DIV_K_CEIL: usize,
+        > UtilitiesInstructions<C::Base>
+        for BaseCircuit<C, NUM_BITS, NUM_BITS_DIV2, NUM_BITS_DIV_K_CEIL>
+    where
+        C::Base: PrimeFieldBits,
+    {
+        type Var = AssignedCell<C::Base, C::Base>;
+    }
+
+    impl<
+            C: CurveAffine,
+            const NUM_BITS: usize,
+            const NUM_BITS_DIV2: usize,
+            const NUM_BITS_DIV_K_CEIL: usize,
+        > Circuit<C::Base> for BaseCircuit<C, NUM_BITS, NUM_BITS_DIV2, NUM_BITS_DIV_K_CEIL>
+    where
+        C::Base: PrimeFieldBits,
+    {
+        type Config = EndoscaleConfig<C, 10>;
+        type FloorPlanner = SimpleFloorPlanner;
+
+        fn without_witnesses(&self) -> Self {
+            Self::default()
+        }
+
+        fn configure(meta: &mut ConstraintSystem<C::Base>) -> Self::Config {
+            let table_config = TableConfig::configure(meta);
+
+            let constants = meta.fixed_column();
+            meta.enable_constant(constants);
+            let endoscalars = meta.instance_column();
+            let endoscalars_copy = meta.advice_column();
+            let acc = (meta.advice_column(), meta.advice_column());
+            let point = (meta.advice_column(), meta.advice_column());
+            let base = (meta.advice_column(), meta.advice_column());
+            let running_sum = meta.advice_column();
+            let pair = (meta.advice_column(), meta.advice_column());
+
+            EndoscaleConfig::<C, 10>::configure(
+                meta,
+                endoscalars,
+                endoscalars_copy,
+                acc,
+                point,
+                base,
+                running_sum,
+                pair,
+                table_config,
+            )
+        }
+
+        fn synthesize(
+            &self,
+            config: Self::Config,
+            mut layouter: impl Layouter<C::Base>,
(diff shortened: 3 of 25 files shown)


halo2/src/recursion/endoscale.rs
```diff
@@ -0,0 +1,102 @@
...
+    fn endoscale_fixed_base<L: Layouter<C::Base>, const NUM_BITS: usize, const NUM_WINDOWS: usize>(
+        &self,
+        layouter: L,
+        base: C,
+        bitstring: &Self::Bitstring,
+    ) -> Result<
+        (
+            AssignedCell<C::Base, C::Base>,
+            AssignedCell<C::Base, C::Base>,
+        ),
```
ebfull, line 45: Right now we assume that the endoscaling algorithm doesn't return the point at infinity, but in theory there are some solutions for the bitstring that would cause this to happen. In the past we've dismissed this as unlikely because the endoscalars were just random challenges, but in other situations where the value is partially or fully controlled by an adversary we could have problems. Let's revisit this with @daira.

halo2/src/recursion/endoscale.rs
```diff
@@ -0,0 +1,102 @@
...
+    /// # Panics
+    /// Panics if `NUM_BITS` is larger than `MAX_BITSTRING_LENGTH`.
+    /// Panics if `NUM_BITS` is an odd number.
+    #[allow(clippy::type_complexity)]
+    fn endoscale_var_base<L: Layouter<C::Base>, const NUM_BITS: usize, const NUM_WINDOWS: usize>(
+        &self,
+        layouter: L,
+        base: (
+            AssignedCell<C::Base, C::Base>,
+            AssignedCell<C::Base, C::Base>,
```
ebfull, line 64: Document or type constrain that this needs to be a valid non-identity point on the curve.

halo2/src/recursion/endoscale.rs
```diff
@@ -0,0 +1,102 @@
...
+where
+    C::Base: PrimeFieldBits,
+{
+    /// A bitstring to be used in endoscaling.
+    type Bitstring: std::fmt::Debug + Clone;
+    /// The maximum number of bits that can be represented by [`Self::Bitstring`].
+    const MAX_BITSTRING_LENGTH: usize;
+
+    /// Witnesses a slice of bools as a vector of [`Self::Bitstring`]s.
+    fn witness_bitstring(bits: &[bool]) -> Vec<Self::Bitstring>;
```
ebfull, line 24: Should this take `&[Option<bool>]`?

halo2/src/recursion/endoscale.rs
```diff
@@ -0,0 +1,102 @@
...
+    /// Computes endoscalar (Alg 2) mapping to a variable-length bitstring using
+    /// the endoscaling algorithm.
+    ///
+    /// The bitstring is decomposed into windows using a running sum outside of
+    /// this gadget.
+    ///
+    /// # Panics
+    /// Panics if `NUM_BITS` is larger than `MAX_BITSTRING_LENGTH`.
+    /// Panics if `NUM_BITS` is an odd number.
+    fn endoscale_scalar<L: Layouter<C::Base>, const NUM_BITS: usize, const NUM_WINDOWS: usize>(
```
ebfull, line 84: `compute_endoscalar` would be a slightly better name I think.

halo2/src/recursion/endoscale/primitive.rs
```diff
@@ -0,0 +1,161 @@
...
+}
+
+/// Maps a K-bit bitstring to a scalar.
+///
+/// This corresponds to Algorithm 1 from [BGH2019], where `F` corresponds to $F_q$, the
+/// scalar field of $P$. Where Algorithm 1 computes $Acc = [scalar] P$, this function
+/// computes `scalar`.
+///
+/// [BGH2019]: https://eprint.iacr.org/2019/1021.pdf
+#[allow(dead_code)]
```
ebfull, line 27: Can this `dead_code` lint be restored?

halo2/src/recursion/endoscale/primitive.rs
```diff
@@ -0,0 +1,161 @@
...
+    for j in (0..(bits.len() / 2)).rev() {
+        let pair = [bits[2 * j], bits[2 * j + 1]];
+        let endo = endoscale_pair_scalar::<F>(pair);
+        acc = acc.double();
+        acc += endo;
+    }
+    acc
+}
+
+/// Maps a pair of bits to a multiple of a base using endoscaling.
```
ebfull, line 45: Document that the provided base must not be the identity.

halo2/src/recursion/endoscale/primitive.rs
```diff
@@ -0,0 +1,161 @@
...
+    }
+
+    C::from_xy(base.0, base.1)
+}
+
+/// Maps a K-bit bitstring to a multiple of a given base.
+///
+/// This is Algorithm 1 from [BGH2019](https://eprint.iacr.org/2019/1021.pdf).
+#[allow(dead_code)]
+pub(crate) fn endoscale<C: CurveAffine>(bits: &[bool], base: C) -> C {
```
ebfull, line 67: Again document that identities can't be passed in.

halo2/src/recursion/endoscale/chip.rs
```diff
@@ -0,0 +1,1142 @@
...
+                b_1 * (endo_x - zeta * p_x)
+            };
+
+            std::array::IntoIter::new([
+                ("b_0_check", b_0_check),
+                ("b_1_check", b_1_check),
+                ("decomposition_check", decomposition_check),
+                ("x_check", x_check),
+                ("y_check", y_check),
+            ])
```
ebfull, line 234: Switch to the new `Constraints` API (or whatever it's called)

ebfull (changes requested):
First pass (didn't cover the chip in detail)

halo2/src/recursion/endoscale.rs
str4d, line 24: This doesn't take a layouter, so it prevents us from asserting that `EndoscaleInstructions::Bitstring` is always correctly constrained. We should pass a layouter argument in here, and then even if the chip doesn't use it, the other parts of the instruction API (as used externally by gadgets) can transitively assume that every `Bitstring` is correctly constrained (even if the chips do so lazily).

halo2/src/recursion/endoscale.rs
```diff
@@ -0,0 +1,102 @@
...
+    const MAX_BITSTRING_LENGTH: usize;
+
+    /// Witnesses a slice of bools as a vector of [`Self::Bitstring`]s.
+    fn witness_bitstring(bits: &[bool]) -> Vec<Self::Bitstring>;
+
+    /// Computes commitment (Alg 1) to a variable-length bitstring using the endoscaling
+    /// algorithm. Uses a fixed base.
+    ///
+    /// The bitstring is decomposed into pairs of bits using a running sum outside of
+    /// this gadget.
```
str4d, line 30:
```suggestion
    /// The bitstring is decomposed into pairs of bits using a running sum outside of
    /// this instruction.
```
But I'd also argue that this is a boundary violation: the instruction user should not need to care how this is implemented, and a chip implementation should be able to decompose with running sum inside this instruction impl if it's better for their use case. If instead every avenue for constructing a `Bitstring` requires that the resulting type is (eventually) constrained, then this becomes an implementation detail.

halo2/src/recursion/endoscale.rs
```diff
@@ -0,0 +1,102 @@
...
+    ///
+    /// # Panics
+    /// Panics if `NUM_BITS` is larger than `MAX_BITSTRING_LENGTH`.
+    /// Panics if `NUM_BITS` is an odd number.
+    fn recover_bitstring<L: Layouter<C::Base>, const NUM_BITS: usize, const NUM_WINDOWS: usize>(
+        &self,
+        layouter: L,
+        bitstring: &Self::Bitstring,
+        pub_input_rows: [usize; NUM_WINDOWS],
+    ) -> Result<(), Error>;
```
str4d, line 101: Assuming that equality-constraining `Bitstring`s can generically be considered "efficient", it's probably more usable to have this return the recovered bitstring instead of requiring the caller to provide it, and then have a separate instruction for equality-constraining two bitstrings. Then a gadget that wants to take an existing bitstring and expose it as a public input can have a method that calls both of these instructions, while a different gadget that wants to grab a bitstring from public inputs and use it in some way can just do that without needing to go through the separate witnessing instruction.
```suggestion
    /// Check that a witnessed bitstring corresponds to a range of endoscalars
    /// provided as public inputs.
    ///
    /// # Panics
    /// Panics if `NUM_BITS` is larger than `MAX_BITSTRING_LENGTH`.
    /// Panics if `NUM_BITS` is an odd number.
    fn recover_bitstring<L: Layouter<C::Base>, const NUM_BITS: usize, const NUM_WINDOWS: usize>(
        &self,
        layouter: L,
        pub_input_rows: [usize; NUM_WINDOWS],
    ) -> Result<Self::Bitstring, Error>;
```

halo2_gadgets/src/utilities/lookup_range_check.rs
```diff
@@ -271,7 +271,7 @@ impl<F: FieldExt + PrimeFieldBits, const K: usize> LookupRangeCheckConfig<F, K>
     pub fn copy_short_check(
         &self,
         mut layouter: impl Layouter<F>,
-        element: AssignedCell<F, F>,
+        element: &AssignedCell<F, F>,
```
str4d, line 274: Needs documentation in the changelog.

halo2/src/recursion/endoscale/primitive.rs
```diff
@@ -0,0 +1,161 @@
...
+
+/// Maps a K-bit bitstring to a scalar.
+///
+/// This corresponds to Algorithm 1 from [BGH2019], where `F` corresponds to $F_q$, the
+/// scalar field of $P$. Where Algorithm 1 computes $Acc = [scalar] P$, this function
+/// computes `scalar`.
+///
+/// [BGH2019]: https://eprint.iacr.org/2019/1021.pdf
+#[allow(dead_code)]
+pub(crate) fn endoscale_scalar<F: FieldExt>(acc: Option<F>, bits: &[bool]) -> F {
```
str4d, line 28: This API takes an optional `acc`, which Algorithm 2 does not; this feels like the function is only partially implementing Algorithm 2 in that case. Document what is going on here when `acc.is_some()`.

EDIT: @therealyingtong explained that `acc` is used specifically for how the efficient chip implements this using a lookup table. That is implementation-specific, but this crate function doesn't live under `endoscale::chip` which is what was confusing me. Rename this function to clarify its purpose, and then have a separate `endoscale_scalar` function that is documented like this currently is (just Alg 2), and does not take `acc` as an argument.

halo2/src/recursion/endoscale/primitive.rs
```diff
@@ -0,0 +1,161 @@
...
+        let pair = [bits[2 * j], bits[2 * j + 1]];
+        let endo = endoscale_pair_scalar::<F>(pair);
+        acc = acc.double();
+        acc += endo;
+    }
+    acc
+}
+
+/// Maps a pair of bits to a multiple of a base using endoscaling.
+pub(crate) fn endoscale_pair<C: CurveAffine>(bits: [bool; 2], base: C) -> CtOption<C> {
```
str4d, line 46: We don't currently have a non-identity type in either `group` or `pasta_curves`. We do have similar wrappers in the `orchard` crate, but we should see if there's a not-horrible way to upstream this into the `group` type system (the initial way I'd imagine integrating this would require downstream curve impls to provide wrapper types equivalent to what we implemented manually in `orchard`).

halo2/src/recursion/endoscale/primitive.rs
```diff
@@ -0,0 +1,161 @@
...
+        let endo = endoscale_pair::<C>(pair, base);
+        acc = acc.double();
+        acc += endo.unwrap();
+    }
+
+    acc.to_affine()
+}
+
+#[allow(dead_code)]
+pub(crate) fn i2lebsp<const NUM_BITS: usize>(int: u64) -> [bool; NUM_BITS] {
```
str4d, line 84: Maybe this can go away once this is in `halo2_gadgets`? But we might also be able to simplify this now that we are on Rust 2021.

halo2/src/recursion/endoscale.rs
```diff
@@ -0,0 +1,102 @@
...
+    /// this gadget.
+    ///
+    /// # Panics
+    /// Panics if `NUM_BITS` is larger than `MAX_BITSTRING_LENGTH`.
+    /// Panics if `NUM_BITS` is an odd number.
+    #[allow(clippy::type_complexity)]
+    fn endoscale_fixed_base<L: Layouter<C::Base>, const NUM_BITS: usize, const NUM_WINDOWS: usize>(
+        &self,
+        layouter: L,
+        base: C,
```
str4d, line 39: Currently, there is no way to constrain in the circuit that multiple invocations of `endoscale_fixed_base` use the same fixed base (it can happen incidentally, but only if the caller passes the same value through to be constant-assigned; a mistake here couldn't be caught in an automated way). Is this intentional (i.e. are the bases supposed to be unconnected)? If not, then a separate associated type, and a corresponding "assign this type from a constant" would be good (and probably what we actually want is just #334).

halo2/src/recursion/endoscale.rs
daira: Search for "We have glossed over a complication." in the [Halo paper](https://eprint.iacr.org/2019/1021.pdf): we proved that incomplete addition can safely be used for a *single* encoscaling.

Note, however, that this argument does not apply when we transpose the loops over i and j in a multi-endoscaling as done here: https://github.com/zcash/halo2/pull/599/files#diff-a884e132e768c2b2d425500ae7976ae5669e9a0cd135159ece1217209a4ec68aR57-R69

If you want I can try to prove that finding an exceptional case in that algorithm entails finding a discrete log relation between the bases. That's plausible, but it will be some effort to do the proof, so before that I want to make sure that's the algorithm that we're actually going to be using.

We should also estimate the circuit cost of a plain multiscalar mul (i.e. not endoscaling), to check whether endoscaling still makes sense &mdash; because the rationale for including it when we were the Halo paper was partly based on the relative cost in Sonic-style arithmetization. (I feel bad that @therealyingtong has already done lots of work on endoscaling without us having done that already, but we were busy with NU5.)

halo2/src/recursion.rs
```diff
@@ -0,0 +1,3 @@
+//! Recursive infrastructure for halo2 circuits.
+
+pub mod endoscale;
```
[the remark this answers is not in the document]
daira: Indeed, there's nothing specific to recursion, or even to zk proofs, about it.

halo2_gadgets/src/endoscale/chip/alg_1.rs
```diff
@@ -0,0 +1,523 @@
...
+                // If the second bit is set, check that endo_x = ζ * P_x
+                let zeta = Expression::Constant(C::Base::ZETA);
+                let b1_set = b_1.clone() * (endo_x.clone() - zeta * p_x.clone());
+
+                // If the second bit is not set, check that endo_x = P_x
+                let not_b1 = Expression::Constant(C::Base::one()) - b_1;
+                let b1_not_set = not_b1 * (endo_x - p_x);
+
+                b1_set + b1_not_set
+            };
```
therealyingtong, line 212: Use `ternary` gate.

halo2_gadgets/src/lib.rs
```diff
@@ -22,6 +22,7 @@
 #![deny(unsafe_code)]
 
 pub mod ecc;
+pub mod endoscale;
```
str4d, line 25:
```suggestion
#[cfg(feature = "unstable-endoscale-gadget")]
pub mod endoscale;
```

book/src/SUMMARY.md
```diff
@@ -42,6 +42,7 @@
     - [Double-and-add](design/gadgets/double-and-add.md)
     - [SHA-256](design/gadgets/sha256.md)
       - [16-bit table chip](design/gadgets/sha256/table16.md)
+    - [Endoscaling](design/gadgets/endoscaling.md)
```
str4d, line 45:
```suggestion
    - Unstable
      - [SHA-256](design/gadgets/sha256.md)
        - [16-bit table chip](design/gadgets/sha256/table16.md)
      - [Endoscaling](design/gadgets/endoscaling.md)
```
I _think_ this would make them sub-sections of an "Unstable" section in the summary, while keeping their URLs stable.

book/src/design/gadgets/endoscaling.md
```diff
@@ -0,0 +1,368 @@
+# Endoscaling
```
str4d, line 1:
```suggestion
# Endoscaling

> Status: Nightly
> Feature flag: `unstable-endoscale-gadget`
```

halo2_gadgets/src/utilities.rs
```diff
@@ -238,6 +238,37 @@ pub fn i2lebsp<const NUM_BITS: usize>(int: u64) -> [bool; NUM_BITS] {
     gen_const_array(|mask: usize| (int & (1 << mask)) != 0)
 }
 
+/// The field element represented by an L-bit little-endian bitstring.
+///
+/// # Panics
+///
+/// Panics if the bitstring is longer than `F::NUM_BITS`.
+pub fn le_bits_to_field_elem<F: PrimeFieldBits>(bits: &[bool]) -> F {
```
str4d, line 246: Does this need to be `pub`? How is it used outside `halo2_gadgets`?

halo2_gadgets/src/utilities/decompose_running_sum.rs
```diff
@@ -32,7 +32,7 @@ use halo2_proofs::{
 use std::marker::PhantomData;
 
 /// The running sum $[z_0, ..., z_W]$. If created in strict mode, $z_W = 0$.
-#[derive(Debug)]
+#[derive(Clone, Debug)]
```
str4d, line 35: Technically should be listed as an addition in the changelog.

halo2_proofs/src/circuit.rs
```diff
@@ -136,6 +136,16 @@ impl<F: Field> AssignedCell<Assigned<F>, F> {
     }
 }
 
+impl<F: Field> From<AssignedCell<F, F>> for AssignedCell<Assigned<F>, F> {
```
str4d, line 139: Document as an addition in the changelog.

halo2_proofs/src/poly/commitment.rs
```diff
@@ -160,6 +160,11 @@ impl<C: CurveAffine> Params<C> {
         self.g.clone()
     }
 
+    /// The `g_lagrange` bases
+    pub fn g_lagrange(&self) -> Vec<C> {
```
str4d, line 164: Document as an addition in the changelog. It would also be useful to document how or why this is useful. If it is only useful for endoscaling:
```suggestion
    #[cfg(feature = "unstable-endoscale-gadget")]
    pub fn g_lagrange(&self) -> Vec<C> {
```

str4d (changes requested):
Gave this a quick review with an eye to differentiating between what can go straight into stable, and what should be behind a nightly feature flag.

PR #537, opened 2022-04-02
Reduce depth of AST by special casing the application of Horner's rule

### LINKED ISSUE #536: Stack overflow with lots of circuits with lots of gates
Backtrace

```
[snip]
#5023 0x000055555693470e in halo2_proofs::poly::evaluator::Evaluator<E,F,B>::evaluate::recurse ()
#5024 0x0000555556934964 in halo2_proofs::poly::evaluator::Evaluator<E,F,B>::evaluate::recurse ()
#5025 0x000055555693470e in halo2_proofs::poly::evaluator::Evaluator<E,F,B>::evaluate::recurse ()
```

I think keeping the gates linearly independent across proofs is increasing the depth of the evaluation tree so much that it's triggering this.

ebfull:
The existing code will fold together a very deep AST that applies Horner's
rule to each gate in a proof -- which could include multiple circuits and
so for some applications will quickly grow such that when we recursively
descend later during evaluation the stack will easily overflow.

This change special cases the application of Horner's rule to a
"DistributePowers" AST node to keep the tree depth from exploding in size.

Closes #536

### DIFF
--- a/halo2_proofs/src/poly/evaluator.rs
+++ b/halo2_proofs/src/poly/evaluator.rs
@@ -152,9 +152,8 @@ impl<E, F: Field, B: Basis> Evaluator<E, F, B> {
                 Ast::Scale(a, _) => collect_rotations(a),
                 Ast::DistributePowers(terms, _) => terms
                     .iter()
-                    .map(|term| collect_rotations(term))
-                    .reduce(|a, b| a.union(&b).cloned().collect())
-                    .unwrap_or_default(),
+                    .flat_map(|term| collect_rotations(term).into_iter())
+                    .collect(),
                 Ast::LinearTerm(_) | Ast::ConstantTerm(_) => HashSet::default(),
             }
         }
@@ -201,17 +200,6 @@ impl<E, F: Field, B: Basis> Evaluator<E, F, B> {
             leaves: &'a HashMap<AstLeaf<E, B>, &'a [F]>,
         }
 
-        impl<'a, E, F: FieldExt, B: Basis> AstContext<'a, E, F, B> {
-            /// Returns the actual size of the chunk we're operating on in this
-            /// context, which may be smaller than `chunk_size`.
-            fn local_chunk_size(&self) -> usize {
-                cmp::min(
-                    self.chunk_size,
-                    self.poly_len - self.chunk_size * self.chunk_index,
-                )
-            }
-        }
-
         fn recurse<E, F: FieldExt, B: BasisOps>(
             ast: &Ast<E, F, B>,
             ctx: &AstContext<'_, E, F, B>,
@@ -241,19 +229,17 @@ impl<E, F: Field, B: Basis> Evaluator<E, F, B> {
                     }
                     lhs
                 }
-                Ast::DistributePowers(terms, base) => {
-                    let mut acc = vec![F::zero(); ctx.local_chunk_size()];
-
-                    for term in terms.iter() {
+                Ast::DistributePowers(terms, base) => terms.iter().fold(
+                    B::constant_term(ctx.poly_len, ctx.chunk_size, ctx.chunk_index, F::zero()),
+                    |mut acc, term| {
                         let term = recurse(term, ctx);
                         for (acc, term) in acc.iter_mut().zip(term) {
                             *acc *= base;
                             *acc += term;
                         }
-                    }
-
-                    acc
-                }
+                        acc
+                    },
+                ),
                 Ast::LinearTerm(scalar) => B::linear_term(
                     ctx.domain,
                     ctx.poly_len,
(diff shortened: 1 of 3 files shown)


daira (approved):
utACK

halo2_proofs/src/poly/evaluator.rs
```diff
@@ -150,6 +150,11 @@ impl<E, F: Field, B: Basis> Evaluator<E, F, B> {
                     lhs.union(&rhs).cloned().collect()
                 }
                 Ast::Scale(a, _) => collect_rotations(a),
+                Ast::DistributePowers(terms, _) => terms
+                    .iter()
+                    .map(|term| collect_rotations(term))
+                    .reduce(|a, b| a.union(&b).cloned().collect())
+                    .unwrap_or_default(),
```
str4d, line 157: This currently allocates multiple times per term, with what looks like poor worst-case behaviour (the `.cloned()` means we'll clone `AstLeaf`s (which could each be 12 or 16 bytes depending on alignment) inside the first term `N` times, and the `.collect()` means we allocate a new `HashSet<AstLeaf>` `N` times with increasing length each time). Instead, we can simplify this by taking advantage of `impl FromIterator<T> for HashSet` (which will automatically produce the union of an iterator that contains duplicates):
```suggestion
                    .flat_map(|term| collect_rotations(term).into_iter())
                    .collect(),
```

halo2_proofs/src/poly/evaluator.rs
```diff
@@ -225,6 +241,19 @@ impl<E, F: Field, B: Basis> Evaluator<E, F, B> {
                     }
                     lhs
                 }
+                Ast::DistributePowers(terms, base) => {
+                    let mut acc = vec![F::zero(); ctx.local_chunk_size()];
```
str4d, line 245: I'd prefer to use the existing basis method to ensure we remain basis-agnostic here (even though this happens to be equivalent for all _current_ bases):
```suggestion
                    let mut acc = B::constant_term(ctx.poly_len, ctx.chunk_size, ctx.chunk_index, F::zero());
```

halo2_proofs/src/poly/evaluator.rs
```diff
@@ -196,6 +201,17 @@ impl<E, F: Field, B: Basis> Evaluator<E, F, B> {
...
+            /// Returns the actual size of the chunk we're operating on in this
+            /// context, which may be smaller than `chunk_size`.
+            fn local_chunk_size(&self) -> usize {
+                cmp::min(
+                    self.chunk_size,
+                    self.poly_len - self.chunk_size * self.chunk_index,
+                )
+            }
+        }
+
```
str4d, line 214: After my other suggestion, this becomes unused. Let's remove it from this PR, but keep it in mind as a nice refactor for a future PR (we could simplify the basis trait APIs by refactoring `AstContext` to have a sub-struct storing just the fields necessary for this calculation, which is currently repeated in several places).

halo2_proofs/src/poly/evaluator.rs
```diff
@@ -225,6 +241,19 @@ impl<E, F: Field, B: Basis> Evaluator<E, F, B> {
...
+                    for term in terms.iter() {
+                        let term = recurse(term, ctx);
+                        for (acc, term) in acc.iter_mut().zip(term) {
+                            *acc *= base;
+                            *acc += term;
+                        }
+                    }
+
+                    acc
+                }
```
str4d, line 256: I would slightly prefer this to use `fold`:
```suggestion
                Ast::DistributePowers(terms, base) => terms.iter().fold(
                    B::constant_term(ctx.poly_len, ctx.chunk_size, ctx.chunk_index, F::zero()),
                    |mut acc, term| {
                        let term = recurse(term, ctx);
                        for (acc, term) in acc.iter_mut().zip(term) {
                            *acc *= base;
                            *acc += term;
                        }
                        acc
                    },
                ),
```
This more closely matches the `reduce` we were originally using (except taking into account the fact that we only have an immutable reference to the AST here, so a `fold` is easier to work with).

halo2_proofs/src/poly/evaluator.rs
```diff
@@ -225,6 +241,19 @@ impl<E, F: Field, B: Basis> Evaluator<E, F, B> {
...
                 }
+                Ast::DistributePowers(terms, base) => {
+                    let mut acc = vec![F::zero(); ctx.local_chunk_size()];
+
+                    for term in terms.iter() {
+                        let term = recurse(term, ctx);
+                        for (acc, term) in acc.iter_mut().zip(term) {
+                            *acc *= base;
+                            *acc += term;
+                        }
```
str4d, line 252: I confirmed that this is equivalent to the previous nested stack of `Ast::Scale` and `Ast::Add`.

halo2_proofs/src/poly/evaluator.rs
```diff
@@ -296,13 +328,24 @@ pub(crate) enum Ast<E, F: Field, B: Basis> {
     ConstantTerm(F),
 }
 
+impl<E, F: Field, B: Basis> Ast<E, F, B> {
+    pub fn distribute_powers<I: IntoIterator<Item = Self>>(i: I, base: F) -> Self {
+        Ast::DistributePowers(Arc::new(i.into_iter().collect()), base)
```
str4d, line 333: At some point I might try and see if I can find a way to avoid this allocation, but it's going to require a bunch of extra lifetime annotations, so only worthwhile if we actually would see a performance benefit.

str4d (changes requested):
The documentation fix is blocking; otherwise ACK with suggestions. I verified that this has no noticeable effect on the Orchard circuit benchmarks (which go up to 4-instance proofs).

str4d (approved):
ACK

daira (approved):
utACK

PR #543, opened 2022-04-08
Optimize Fft

NoCtrlZ:
# Fft Optimization
Hi there.
I optimized `best_fft` algorithm and replaced with more efficient `rayon` method.

## What I did
`n` = 2^k
### precompute twiddle factors
To precompute the `twiddle factors`, reuse them and it reduced `n` times multiplication.

### avoid multiplication when twiddle factor = 1
Prepare specific case when twiddle factor = 1.
To avoid unnecessary `twiddle factor * element` and it reduced `2n` times multiplication.
It slowed down when I implement it as `if` condition so I hardcode that case.

### `scope` to `par_iter`
Replace `scope` with `par_iter`.
The official document says that the `par_iter` is the most efficient basically.
https://docs.rs/rayon/latest/rayon/#how-to-use-rayon
I benched them and the speed is the `scope > par_iter > join` but the `join` uses a lot of memory.
https://github.com/appliedzkp/halo2/pull/55#issue-1195660345

### other
Avoid `unshuffle` and multiple times bit reverse, it reduced `n` times substitution.
Merge `serial_fft`, `parallel_fft` and `log2_floor` to one function.

## Bench
The `prover` has a [bottleneck](https://github.com/appliedzkp/halo2/pull/36#issuecomment-1085456038) so it doesn't reduce time so much.
|k = `16`|keygen|prover|verifier|
|---|---|---|---|
|default|77.531 s 77.569 s 77.615 s|13.259 s 13.278 s 13.297 s|384.52 ms 384.98 ms 385.71 ms|
|fft|63.543 s 63.631 s 63.764 s|13.015 s 13.025 s 13.035 s|210.18 ms 210.40 ms 210.67 ms|

I would appreciate it if you would confirm.
Thank you.

### DIFF
--- a/halo2_proofs/src/arithmetic.rs
+++ b/halo2_proofs/src/arithmetic.rs
@@ -191,12 +191,13 @@ pub fn best_fft<G: Group>(a: &mut [G], omega: G::Scalar, log_n: u32) {
     }
 
     // precompute twiddle factors
-    let mut w = G::Scalar::one();
-    let mut twiddles = vec![G::Scalar::one(); (n / 2) as usize];
-    for tw in twiddles.iter_mut() {
-        *tw = w;
-        w.group_scale(&omega);
-    }
+    let twiddles: Vec<_> = (0..(n / 2) as usize)
+        .scan(G::Scalar::one(), |w, _| {
+            let tw = *w;
+            w.group_scale(&omega);
+            Some(tw)
+        })
+        .collect();
 
     if log_n <= log_threads {
         let mut chunk = 2_usize;
--- a/halo2_proofs/src/arithmetic.rs
+++ b/halo2_proofs/src/arithmetic.rs
@@ -169,18 +169,7 @@ pub fn best_multiexp<C: CurveAffine>(coeffs: &[C::Scalar], bases: &[C]) -> C::Cu
 ///
 /// This will use multithreading if beneficial.
 pub fn best_fft<G: Group>(a: &mut [G], omega: G::Scalar, log_n: u32) {
-    let threads = multicore::current_num_threads();
-    let log_threads = log2_floor(threads);
-
-    if log_n <= log_threads {
-        serial_fft(a, omega, log_n);
-    } else {
-        parallel_fft(a, omega, log_n, log_threads);
-    }
-}
-
-fn serial_fft<G: Group>(a: &mut [G], omega: G::Scalar, log_n: u32) {
-    fn bitreverse(mut n: u32, l: u32) -> u32 {
+    fn bitreverse(mut n: usize, l: usize) -> usize {
         let mut r = 0;
         for _ in 0..l {
             r = (r << 1) | (n & 1);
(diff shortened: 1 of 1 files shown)


str4d:
Thanks for the PR @NoCtrlZ! We'll review it this week.

ebfull:
We added a benchmark in #552 that I ran locally on this PR to assess the impact.

```
Gnuplot not found, using plotters backend
fft/k/3                 time:   [106.66 us 106.87 us 107.08 us]                    
                        change: [+5603.2% +5625.0% +5651.3%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 10 outliers among 100 measurements (10.00%)
  2 (2.00%) low mild
  2 (2.00%) high mild
  6 (6.00%) high severe
fft/k/4                 time:   [300.40 us 301.84 us 303.38 us]                    
                        change: [+8298.5% +8345.8% +8400.4%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 5 outliers among 100 measurements (5.00%)
  1 (1.00%) low mild
  4 (4.00%) high mild
fft/k/5                 time:   [424.64 us 427.30 us 430.03 us]                    
                        change: [+5875.0% +5930.8% +5987.5%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 4 outliers among 100 measurements (4.00%)
  1 (1.00%) low mild
  3 (3.00%) high mild
fft/k/6                 time:   [227.02 us 228.89 us 230.83 us]                    
                        change: [+431.68% +449.11% +463.92%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 6 outliers among 100 measurements (6.00%)
  1 (1.00%) low severe
  1 (1.00%) low mild
  3 (3.00%) high mild
  1 (1.00%) high severe
fft/k/7                 time:   [182.95 us 184.51 us 186.17 us]                    
                        change: [+189.29% +200.70% +210.54%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 3 outliers among 100 measurements (3.00%)
  2 (2.00%) low mild
  1 (1.00%) high mild
fft/k/8                 time:   [259.16 us 262.10 us 265.02 us]                    
                        change: [+177.92% +186.59% +195.19%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 12 outliers among 100 measurements (12.00%)
  2 (2.00%) low mild
  9 (9.00%) high mild
  1 (1.00%) high severe
fft/k/9                 time:   [365.51 us 369.22 us 373.13 us]                    
                        change: [+128.93% +136.59% +144.39%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 4 outliers among 100 measurements (4.00%)
  3 (3.00%) high mild
  1 (1.00%) high severe
fft/k/10                time:   [561.13 us 568.10 us 575.54 us]                     
                        change: [+95.243% +99.895% +104.69%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 4 outliers among 100 measurements (4.00%)
  2 (2.00%) low mild
  2 (2.00%) high mild
fft/k/11                time:   [848.05 us 852.88 us 857.78 us]                     
                        change: [+62.124% +64.790% +67.511%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 10 outliers among 100 measurements (10.00%)
  2 (2.00%) low mild
  5 (5.00%) high mild
  3 (3.00%) high severe
Benchmarking fft/k/12: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 6.1s, enable flat sampling, or reduce sample count to 60.
fft/k/12                time:   [1.1982 ms 1.2063 ms 1.2155 ms]                      
                        change: [+31.052% +32.531% +34.144%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 13 outliers among 100 measurements (13.00%)
  5 (5.00%) low mild
  7 (7.00%) high mild
  1 (1.00%) high severe
Benchmarking fft/k/13: Warming up for 3.0000 s
Warning: Unable to complete 100 samples in 5.0s. You may wish to increase target time to 8.9s, enable flat sampling, or reduce sample count to 50.
fft/k/13                time:   [1.7562 ms 1.7664 ms 1.7769 ms]                      
                        change: [+7.8407% +9.4387% +11.060%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 3 outliers among 100 measurements (3.00%)
  1 (1.00%) high mild
  2 (2.00%) high severe
fft/k/14                time:   [2.6159 ms 2.6315 ms 2.6479 ms]                      
                        change: [-17.393% -16.100% -14.828%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 2 outliers among 100 measurements (2.00%)
  2 (2.00%) high mild
fft/k/15                time:   [4.0539 ms 4.0786 ms 4.1046 ms]                      
                        change: [-28.847% -27.828% -26.788%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild

```

These results suggest two things:

1. at low versions of `k` it is worth having a single thread do the work, as it currently does, due to the overhead of multithreading
2. at high versions of `k` (on my system, `14` or more) it is beneficial to use your patch

There seem to be some tradeoffs in the algorithm that we need to address before we merge this PR.

NoCtrlZ:
@ebfull 
Thank you for feedback!
I am going to address.

NoCtrlZ:
Hi @ebfull 
I found the bottlenecks and fixed.

What I did is separated low degree butterfly arithmetic with condition branch and replaced `par_iter` with `join` because `par_iter` takes overhead as you mentioned.
The following is experiment.
https://github.com/appliedzkp/halo2/pull/55#issue-1195660345

The optimized fft is faster for all k now.
I benched it locally with thread num = 8 and following is the result.

| k | default | recursive |
| ---- | ---- | ---- |
|3| 1.8787 us 1.8841 us 1.8898 us | 510.74 ns 513.45 ns 516.32 ns |
|4| 27.871 us 29.831 us 32.287 us | 13.491 us 13.600 us 13.712 us |
|5| 40.885 us 44.149 us 47.757 us | 17.910 us 17.960 us 18.015 us |
|6| 49.677 us 52.739 us 56.196 us | 29.320 us 29.443 us 29.582 us |
|7| 51.373 us 52.009 us 52.890 us | 41.120 us 41.502 us 41.985 us |
|8| 100.27 us 106.84 us 114.88 us | 54.564 us 54.695 us 54.837 us |
|9| 198.94 us 227.62 us 259.97 us | 83.676 us 84.190 us 84.777 us |
|10| 277.42 us 283.34 us 290.65 us | 144.14 us 145.33 us 146.65 us |
|11| 506.49 us 511.16 us 516.39 us | 269.93 us 272.76 us 275.41 us |
|12| 1.0648 ms 1.0964 ms 1.1335 ms | 534.74 us 539.85 us 544.27 us |
|13| 2.0286 ms 2.0736 ms 2.1223 ms | 1.0254 ms 1.0290 ms 1.0332 ms |
|14| 5.6022 ms 5.8867 ms 6.2248 ms | 2.0972 ms 2.1071 ms 2.1186 ms |
|15| 8.2442 ms 8.4402 ms 8.6559 ms | 4.3206 ms 4.3327 ms 4.3451 ms | 
|16| 19.105 ms 19.842 ms 20.658 ms | 9.1073 ms 9.1606 ms 9.2234 ms |
|17| 37.900 ms 39.631 ms 41.765 ms | 23.751 ms 24.573 ms 25.512 ms |
|18| 73.686 ms 76.518 ms 80.451 ms | 48.620 ms 50.373 ms 52.601 ms |

I would appreciate it if you confirm.
Thank you!

therealyingtong:
I've locally verified this performance improvement.

halo2_proofs/src/arithmetic.rs
```diff
@@ -189,79 +178,107 @@ fn serial_fft<G: Group>(a: &mut [G], omega: G::Scalar, log_n: u32) {
         r
     }
 
-    let n = a.len() as u32;
+    fn log2_floor() -> u32 {
```
therealyingtong, line 181:
```suggestion
    fn log2_floor_threads() -> u32 {
```

therealyingtong, line 181: Was there a particular reason behind changing the original `log2_floor(num: usize)`?

NoCtrlZ: Yes.
`log2_floor` is not referred from other function so it should be local function as the same with `bitreverse`.
And the thread number can get where ever and it's not used in best_fft function so it should be local variable.
I think that makes more sense.
What do you think?

daira: Even though `log2_floor` is not used elsewhere, making it a standalone function would allow it to be tested separately, and IMHO is more understandable for review and maintenance. (In general, if you have a large complicated function like `*_fft` then moving the code of another function into it is going in the wrong direction.)

If `log2_floor` were kept as it is in this PR then it would need to be renamed, because it is no longer calculating the log2_floor of a general argument. But I prefer to revert this change.

daira: I would also prefer to move out `bitreverse` actually, but that should not be done in this PR.

halo2_proofs/src/arithmetic.rs
```diff
@@ -189,79 +178,107 @@ fn serial_fft<G: Group>(a: &mut [G], omega: G::Scalar, log_n: u32) {
...
-                        t.group_scale(&elt);
-                        tmp.group_add(&t);
-                        elt *= &omega_step;
-                    }
-                    elt *= &omega_j;
-                }
-
-                // Perform sub-FFT
-                serial_fft(tmp, new_omega, log_new_n);
+    if log_n <= log_threads {
```
daira, line 211: This uses a non-recursive implementation for the whole FFT when log_n ≤ log_threads. If it is not, then the recursive butterfly recurses all the way down to n = 2. My intuition would be that it's best to bottom out with the non-recursive implementation at a higher n, i.e. to switch from recursive to non-recursive when n becomes small enough within a large FFT.

halo2_proofs/src/arithmetic.rs
```diff
@@ -189,79 +178,107 @@ fn serial_fft<G: Group>(a: &mut [G], omega: G::Scalar, log_n: u32) {
...
-    for (idx, a) in a.iter_mut().enumerate() {
-        *a = tmp[idx & mask][idx >> log_threads];
+/// This perform recursive butterfly arithmetic
+pub fn recursive_butterfly_arithmetic<G: Group>(
+    a: &mut [G],
+    n: usize,
+    twiddle_chunk: usize,
+    twiddles: &[G::Scalar],
+) {
+    if n == 2 {
```
daira, line 252: See comment above.

daira (commented):
Reviewed with comments.

halo2_proofs/src/arithmetic.rs
NoCtrlZ: > If log2_floor were kept as it is in this PR then it would need to be renamed
making it a standalone function allows it to be unit-tested separately
I would also prefer to move out bitreverse actually, but that should not be done in this PR.

That makes sense.
I am going to revert.

halo2_proofs/src/arithmetic.rs
NoCtrlZ: > My intuition would be that it's best to bottom out at a higher n

Yes.
The computation doesn't cost much so the less multiplicity, the faster butterfly performs.
> to switch from recursive to non-recursive when n becomes small enough within a large FFT

I think all recursive is the best option because of rayon specification.
We use `join` for recursive and `scope` for for loop.
The `join` allocates stack and `scope` allocates heap.
We should use `join` as possible as we can as the [experiment](https://github.com/appliedzkp/halo2/pull/55#issue-1195660345) suggests.

I think we can improve more to implement following.

- implement 2 layer butterfly arithmetic

As you suggest, the less multiplicity, the faster butterfly performs.
If we perform 2 layer butterfly arithmetic, it would improve performance and be able to user intermediate value.
But code will be mess as following.
https://github.com/appliedzkp/halo2/blob/fft-optimization/halo2_proofs/src/arithmetic.rs#L216
It's not so hard just adding `if n == 4` condition and divide by 4.
Should I implement it?

- change `par_iter` when the n is large enough

I use `left.iter_mut` when we perform butterfly arithmetic in `recursive_butterfly`.
https://github.com/zcash/halo2/pull/543/files#diff-e03420d624b59ef5f1ea16419344db88889e1864d7b089db7653ae943b4c7809R272
We can replace with `par_iter_mut` but it was bottleneck because the overhead of multithread costs more than speed up.
If we replace `iter_mut` with `par_iter_mut` at proper n, it would improve performance but we shouldn't use `if` condition because it slows down.
But we need some experiments.

NoCtrlZ:
@daira 
Thank you for the review.
I am going to revert about `log2_floor`.
We also can improve more as I commented but it's a bit complicated and makes code mess.
I have other ideas about optimizations of halo2 which improves performance so much.
We can optimize deeply after major improvement in my opinion.
What do you think?

halo2_proofs/src/arithmetic.rs
daira: I don't find this code too bad: https://github.com/appliedzkp/halo2/blob/fft-optimization/halo2_proofs/src/arithmetic.rs#L216
It seems feasibly reviewable and would be worth the complexity if it gave, say, another 5% improvement or more.

NoCtrlZ: Copy that.

daira (approved):
utACK

str4d:
Yes, the commits should be squashed (or otherwise cleaned up).

NoCtrlZ:
Thank you for confirm.
I modified according to your reviews.

ebfull (approved):

ebfull:
Thank you for the PR!

PR #549, opened 2022-04-14
cost-model.rs: Correct lookup required degree calculation.

### LINKED ISSUE #539: Cost model miscalculates the degree of lookup arguments
https://github.com/zcash/halo2/blob/95df0af86dab803ca5dea5586dae15d9b2f75394/halo2_proofs/examples/cost-model.rs#L127-L129

does not match

https://github.com/zcash/halo2/blob/95df0af86dab803ca5dea5586dae15d9b2f75394/halo2_proofs/src/plonk/lookup.rs#L65-L70

therealyingtong:
Closes #539

### DIFF
--- a/halo2_proofs/examples/cost-model.rs
+++ b/halo2_proofs/examples/cost-model.rs
@@ -125,7 +125,7 @@ impl FromStr for Lookup {
 
 impl Lookup {
     fn required_degree(&self) -> usize {
-        1 + cmp::max(1, self.input_deg) + cmp::max(1, self.table_deg)
+        2 + cmp::max(1, self.input_deg) + cmp::max(1, self.table_deg)
     }
 
     fn queries(&self) -> impl Iterator<Item = Poly> {


halo2_proofs/examples/cost-model.rs
```diff
@@ -125,7 +125,7 @@ impl FromStr for Lookup {
 
 impl Lookup {
     fn required_degree(&self) -> usize {
-        1 + cmp::max(1, self.input_deg) + cmp::max(1, self.table_deg)
+        2 + cmp::max(1, self.input_deg) + cmp::max(1, self.table_deg)
```
str4d, line 128: I confirmed that the reason for this bug was that this cost model example was implemented before the lookup argument was made perfectly zero-knowledge: https://github.com/zcash/halo2/commit/1f1b705fa662e0ab8cb35e4b5e0fc7fb9cc20282

I also agree that this happens to ensure the lowest returned degree is 4, matching the real code.

str4d (approved):
ACK

Issue #482, opened 2022-02-01
Raise MSRV to 1.56.1


str4d:
Rationale:
- 2021 edition support:
  - v2 resolver.
  - `rust-version` in `Cargo.toml`
- `bumpalo 3.9.0` raises its MSRV to 1.54.0
- `uint 0.9.2` raises its MSRV to 1.56.1 (not 1.56.0).
- `ff` will be raising its MSRV to 1.56.0, in line with `bitvec 1.0.0`

PR #556, opened 2022-04-24
halo2_gadgets: Migrate chip gates to `Constraints::with_selector`

str4d:
Only one gate couldn't be migrated without altering the Orchard circuit.

### DIFF
--- a/halo2_gadgets/src/poseidon/pow5.rs
+++ b/halo2_gadgets/src/poseidon/pow5.rs
@@ -4,7 +4,9 @@ use std::iter;
 use halo2_proofs::{
     arithmetic::FieldExt,
     circuit::{AssignedCell, Cell, Chip, Layouter, Region},
-    plonk::{Advice, Any, Column, ConstraintSystem, Error, Expression, Fixed, Selector},
+    plonk::{
+        Advice, Any, Column, ConstraintSystem, Constraints, Error, Expression, Fixed, Selector,
+    },
     poly::Rotation,
 };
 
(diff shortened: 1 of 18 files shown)


str4d:
Ignore the beta Clippy lints about `array::IntoInter::new`; they are because `beta` is 1.61, while we are currently pinned to MSRV 1.51 which doesn't have the APIs the beta lints refer to. The lints will be fixed once #482 is addressed.

halo2_gadgets/src/poseidon/pow5.rs
```diff
@@ -94,20 +96,23 @@ impl<F: FieldExt, const WIDTH: usize, const RATE: usize> Pow5Chip<F, WIDTH, RATE
...
+                            .map(|idx| {
+                                let state_cur = meta.query_advice(state[idx], Rotation::cur());
+                                let rc_a = meta.query_fixed(rc_a[idx], Rotation::cur());
+                                pow_5(state_cur + rc_a) * m_reg[next_idx][idx]
+                            })
+                            .reduce(|acc, term| acc + term)
+                            .expect("WIDTH > 0");
+                        expr - state_next
+                    })
+                    .collect::<Vec<_>>(),
```
daira, line 114: Collecting into a Vec shouldn't be necessary here, I think.
```suggestion
```

str4d: It is necessary due to lifetimes. Believe me, I tried 🙂

daira (approved):
utACK

ebfull (approved):

PR #557, opened 2022-04-24
Fix Halo 2 book deployment

str4d:


str4d:
Self-ACK, only alters the book CI workflow.

PR #558, opened 2022-04-24
CI: Pass the correct `katex-header.html` path to rustdoc

str4d:
The header was moved to fix docs.rs deployment, but the CI workflow wasn't updated to account for this.


str4d:
Self-ACK, only touches the book workflow.

PR #565, opened 2022-04-29
CI: Fix Rust version used for benchmarks

str4d:


str4d (commented):
Trivial self-ACK.

daira (commented):
Post-hoc ACK

PR #525, opened 2022-03-18
[halo2_gadgets] Bitstring util

therealyingtong:

PR #568, opened 2022-05-02
Changes from Orchard circuit review

str4d:
Part of https://github.com/zcash/orchard/issues/83.

### DIFF
--- a/halo2_gadgets/src/utilities.rs
+++ b/halo2_gadgets/src/utilities.rs
@@ -99,8 +99,8 @@ impl<F: Field, T: FieldValue<F>> RangeConstrained<F, T> {
 }
 
 impl<F: PrimeFieldBits> RangeConstrained<F, Option<F>> {
-    /// Constructs a `RangeConstrained<Option<F>>` as a subset of the given value.
-    pub fn subset_of(value: Option<&F>, bitrange: Range<usize>) -> Self {
+    /// Constructs a `RangeConstrained<Option<F>>` as a bitrange of the given value.
+    pub fn bitrange_of(value: Option<&F>, bitrange: Range<usize>) -> Self {
         let num_bits = bitrange.len();
         Self {
             inner: value.map(|value| bitrange_subset(value, bitrange)),
--- a/halo2_gadgets/src/utilities.rs
+++ b/halo2_gadgets/src/utilities.rs
@@ -1,16 +1,35 @@
 //! Utility gadgets.
 
-use ff::PrimeFieldBits;
+use ff::{Field, PrimeFieldBits};
 use halo2_proofs::{
     circuit::{AssignedCell, Cell, Layouter},
     plonk::{Advice, Column, Error, Expression},
 };
 use pasta_curves::arithmetic::FieldExt;
+use std::marker::PhantomData;
 use std::ops::Range;
 
 pub mod cond_swap;
 pub mod decompose_running_sum;
 pub mod lookup_range_check;
+
+/// A type that has a value at either keygen or proving time.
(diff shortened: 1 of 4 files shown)


halo2_gadgets/src/utilities.rs
```diff
@@ -59,6 +78,62 @@ pub trait UtilitiesInstructions<F: FieldExt> {
...
+        }
+    }
+}
+
+impl<F: Field> RangeConstrained<F, AssignedCell<F, F>> {
+    /// Constructs a `RangeConstrained<AssignedCell<F, F>>` without verifying that the
+    /// cell is correctly range constrained.
+    ///
+    /// This API only exists to ease with integrating this type into existing circuits,
+    /// and will likely be removed in future.
```
ebfull, line 118: Unsure if it's appropriate here to set `#[doc(hidden)]` if it's intended to be removed.

str4d: My plan is to finish going through the circuit and finding all the ways we want to construct this type, and then see how we can better integrate it into the existing APIs.

halo2_gadgets/src/utilities.rs
```diff
@@ -59,6 +78,62 @@ pub trait UtilitiesInstructions<F: FieldExt> {
...
+
+    /// Returns the number of bits to which this cell is constrained.
+    pub fn num_bits(&self) -> usize {
+        self.num_bits
+    }
+}
+
+impl<F: PrimeFieldBits> RangeConstrained<F, Option<F>> {
+    /// Constructs a `RangeConstrained<Option<F>>` as a subset of the given value.
+    pub fn subset_of(value: Option<&F>, bitrange: Range<usize>) -> Self {
```
str4d, line 103:
```suggestion
    /// Constructs a `RangeConstrained<Option<F>>` as a bitrange of the given value.
    pub fn bitrange_of(value: Option<&F>, bitrange: Range<usize>) -> Self {
```

halo2_gadgets/src/utilities.rs
str4d: I'm going to come back to this in a second PR going over the `halo2_gadgets` circuit impls.

halo2_gadgets/src/utilities.rs
```diff
@@ -59,6 +78,62 @@ pub trait UtilitiesInstructions<F: FieldExt> {
     }
 }
 
+/// A type representing a range-constrained field element.
+#[derive(Clone, Copy, Debug)]
+pub struct RangeConstrained<F: Field, T: FieldValue<F>> {
```
therealyingtong, line 83: Could this be something like 

```rust
struct RangeConstrained<F: Field, NUM_BITS: usize>(AssignedCell<Bits<NUM_BITS>, F)
```

Where `Bits<NUM_BITS: usize>: Into<Assigned<F>>`.

therealyingtong: Something like in https://github.com/zcash/halo2/pull/525/commits/c4cebc675129ccb578170d4febb80a1586a70cfa.

ebfull (approved):

halo2_gadgets/src/utilities.rs
str4d: That would not work for the way the `MessagePiece::from_subpieces` API works, as we can't yet have an iterator that is generic over a type parameter. And we can't do type-system concatenation until Rust implements const generics expressions, so we'd instead need to use something like `generic_array`. It might work if we also rework the Sinsemilla APIs, but that will have to wait until after `halo2_gadgets 0.1.0` (so won't ever be used by the Orchard circuit).

PR #574, opened 2022-05-06
Halo2 Proof Fixes

defuse:
This fixes three problems:

- The informal description of zero knowledge should worry about what the verifier learns, not the prover.
- $\in$ is undefined for the transcript strings.
- $\tau$ is appended to the transcript after $\tau$ has already been checked to be a prefix of the transcript, which doesn't seem right.

### DIFF
--- a/book/src/design/protocol.md
+++ b/book/src/design/protocol.md
@@ -83,7 +83,7 @@ notions.
   correct, does the prover actually possess ("know") a valid witness? We refer
   to the probability that a cheating prover falsely convinces the verifier of
   this knowledge as the _knowledge error_.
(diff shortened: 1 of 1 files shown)


defuse:
Reviewers should check my understanding of the $\tau$ thing, I'm not 100% sure (another interpretation would be that $\in$ is a substring but if that's the case the encoding is ambiguous I think).

defuse:
I also haven't figured out how to render the latex locally so please double-check that or tell me how

str4d:
> I also haven't figured out how to render the latex locally so please double-check that or tell me how

```
cargo install mdbook mdbook-katex
mdbook build book
```

PR #551, opened 2022-04-18
[book] Add description of selector combining optimization

daira:
fixes #466

PR #580, opened 2022-05-10
[book] Add 'Selector combining' to SUMMARY.md

therealyingtong:
This was added in https://github.com/zcash/halo2/pull/551 but may have been undone somewhere.


daira (approved):
ACK. Doc-only change

PR #581, opened 2022-05-10
Release `halo2_proofs` and `halo2_gadgets` 0.1.0

str4d:


ebfull (changes requested):
Still need to remove the "we're in beta and not audited" thing from the `halo2_gadgets` repo (you've just removed it from `halo2_proofs`)

str4d:
The only outdated dependency is `pprof`, which is less than a day old and is only a dev-dependency.

ebfull (approved):

daira (commented):
Post-hoc utACK

PR #586, opened 2022-05-19
Remove `UtilitiesInstructions` trait

therealyingtong:
The UtilitiesInstructions trait had an associated type Var, and a
helper method load_private().

In practice, Var is always AssignedCell, and all instances of
Self::Var have now been replaced with AssignedCell.

The load_private() method was only ever used in tests, and has
been flagged off as a standalone method under #cfg[(test)].


daira (changes requested):
This changes public API so needs a changelog entry.

PR #593, opened 2022-06-01
Support multi-phase circuit

### LINKED ISSUE #527: Potential usages if `halo2` has `meta.randomness()`
It would be nice to have an API like `meta.randomness()` to allocate randomness for custom gates, which guarantees the allocated randomness is squeezed after all previous allocated columns are committed and written into transcript.

Some potential usages in our case ([appliedzkp/zkevm-circuits](https://github.com/appliedzkp/zkevm-circuits)) are:

1. To provide shuffle argument when we want the witness trace to be verified in a different order. One case is to shuffle the access record of RAM to another friendly order for consistency check. With such API, the custom gate would be implemented like:

    ```rust
    let [q_shuffle, q_first, q_last] = [(); 3].map(|_| meta.selector());
    let [a0, a1, b0, b1] = [(); 4].map(|_| meta.advice_column());
    let [theta, beta] = [(); 2].map(|_| meta.randomness());
    let z = meta.advice_column();

    meta.create_gate(|meta| {
        let [q_shuffle, q_first, q_last]  = [q_shuffle, q_first, q_last].map(|q| meta.query_selector(q));
        let [a0, a1, b0, b1] = [a0, a1, b0, b1].map(|a| meta.query_advice(a, Rotation::cur()));
        let [z, zw] = [Rotation::cur(), Rotation::next()].map(|rotation| meta.query_advice(z, rotation));

        // compress
        let a = a0 * theta + a1;
        let b = b0 * theta + b1;

        vec![
            q_shuffle * (z * (a + beta) - zw * (b + beta)),
            (q_first + q_last) * (1 - z),
        ]
    });
    ```

    Although I guess this could also be achieved by modularizing the shuffle argument used in subset argument, and provide API like `meta.shuffle()`, then the code could be much simpler like:

    ```rust
    let q_shuffle = meta.selector();
    let [a0, a1, b0, b1] = [(); 4].map(|_| meta.advice_column());

    meta.shuffle(|meta| {
        let q_shuffle  = meta.query_selector(q_shuffle);
        let [a0, a1, b0, b1] = [a0, a1, b0, b1].map(|a| meta.query_advice(a, Rotation::cur()));

        [(a0, b0), (a1, b1)].map(|(lhs, rhs)| (q_shuffle * lhs, q_shuffle * rhs))
    });
    ```

2. Another usage is to copy dynamic-size vector of witness to different regions, to avoid sacrificing capacity due to rare but worst case. One of case is to do keccak256 on dynamic size bytes, although the keccak256 is designed to be repeating the keccak-f to accept dynamic size input, but passing input becomes an issue because we can't use permutation argument for that. One approach to do is make the caller to reorder the witness to align with the one keccak256 expects, but it makes the gate complicated (need to be aware of which round it is, and maintain a buffer of bytes keccak256 expects).

    So instead passing byte by byte, we'd like to try to use random linear combination <code>Σ<sub>i</sub>(b<sub>i</sub> * θ<sup>i</sup>)</code> to accumulate dynamic size bytes into a single witness, and use lookup (or shuffle) to make sure caller and keccak256 accumulate to same value at specific row, with its corresponding output. Then caller doesn't need to be aware of which keccak-f round it is, and only need to maintain a single witness as all bytes, and no buffer is required anymore.

But there are also some issues come to my mind:

1. Because some witness value would depend on randomness, `synthesize` might need to be called multiple times, which seems confusing to most users that don't need extra randomness.
2. In above pseudo code, I assume an extra round is implicitly added when `meta.randomness` is called after any advice columns are allocated, but such implicit behavior might also be confusing. Another explicit API approach is to require user to specify which round of each advice column and randomness is in, but it seems too annoying for most user.
3. From `Expression` it seems hard to tell if the randomness is dependency of other column's witness, so users need to be aware themselves. But it seems fine because we would find some value couldn't be calculated because it depends on future randomness, then we know we have done a wrong configuration.

Also even without this API, we can still slightly modify the `create_proof` and `verify_proof` in a fork of `halo2`, to call `synthesize` multiple times and mock the randomness with a "constant instance column", thanks for the extremely flexible API that already exists! But still like to explore the possibility to allocate randomness while configuration.

han0110:
This PR aims to support multi-phase circuit (closes #527). The main changes include:

1. [`commit@794d370`](https://github.com/zcash/halo2/commit/794d370a308bd6b0c723ba616783088a42e235e4)
    - Add an enum `Phase`, which has 3 variant `First`, `Second` and `Third` (3 phases should be enough for most cases in my imagination).
    - Add a field `phase` to struct `Advice`, variant `Any::Advice` and variant `Expression::Advice`.
2. [`commit@a330f52`](https://github.com/zcash/halo2/commit/a330f520086e204f9ca2dfef0ec46802a2003f82)
    - Add a struct `Challenge` and can be allocated by `ConstraintSystem::challenge_in(phase)`, and it's also
        1. Used in `VirtualCells::query_challenge(challenge)` to get `Expression::Challenge`.
        2. Used in `Layouter::get_challenge(challenge)` to get `Option<F>` (`None` if current phase is earlier than requested `challenge`).
    - In `create_proof`, `ConcreteCircuit::FloorPlanner::synthesize` will be call as many time as the phases used in the circuit, and `WitnessCollection` will ignore the advice column assignment if phase doesn't match.

Also it adds a new example [`shuffle.rs`](https://github.com/zcash/halo2/blob/d50a81e7a36bbbfa1892ae140159264f032fa07f/halo2_proofs/examples/shuffle.rs) to show how to use multi-phase to roll our own vector shuffle argument.

### Handling advice column assignment in wrong phase

Instead of returning an error, currently `WitnessCollection` ignores the advice column assignment if phase doesn't match (like [this](https://github.com/zcash/halo2/blob/a330f520086e204f9ca2dfef0ec46802a2003f82/halo2_proofs/src/plonk/prover.rs#L195-L198)). The reason is to make sure the regions measured by `V1` planner to be same in different phase (To me it's like in key generation (phase 0) we still need to call `assign_advice`).

One approach is to call and check the value closure returns an `Err(Error::UnmatchedPhase)`, but not sure if this is desired, because it breaks the principle that all value closures passed to any `Region::assign_*` method will be called at most once.

### `PinnedConstraintSystem` is updated, so as `PinnedVerificationKey`

Since a new field is added to `Advice`, and several fields `num_challenges`, `advice_column_phase` and `challenge_phase` are added to `ConstraintSystem`, the `PinnedConstraintSystem` is also updated to carry these information. So this PR makes `VerifyingKey::hash_into` to have different result, and might break some existing prover/verifier that relies on serialized pk/vk (a breaking change to `halo2`).

### Effect to small circuit that doesn't need multi-phase

The memory usage of `Column<Advice>` is increased to `12` or `16` bytes, and several fields are added `ConstraintSystem` which also increases memory usage. This might affect small circuit's performance.

The benchmark or `orchard` circuit seems to be unstable on my mac (Intel Core i5), but here is the result:

<details>
<summary>As-is <a href="https://github.com/zcash/halo2/tree/c0db68aa0535cf4071ee58f5b72c8211f888b5a3"><code>halo2_proof@c0db68a</code></a></summary>
<br>

```
proving/bundle/1        time:   [4.3719 s 4.4765 s 4.5679 s]
proving/bundle/2        time:   [4.3741 s 4.4825 s 4.5911 s]
proving/bundle/3        time:   [5.9627 s 6.0780 s 6.1929 s]
proving/bundle/4        time:   [7.5456 s 7.9638 s 8.4966 s]

verifying/bundle/1      time:   [43.363 ms 45.014 ms 46.810 ms]
verifying/bundle/2      time:   [40.755 ms 43.009 ms 46.204 ms]
verifying/bundle/3      time:   [55.697 ms 60.355 ms 65.751 ms]
verifying/bundle/4      time:   [48.969 ms 49.327 ms 49.727 ms]
```

</details>

<details>
<summary>With this PR <a href="https://github.com/zcash/halo2/tree/1e96c837d9d5475531ce36bdcf4f79c4e1c3a99f"><code>halo2_proof@1e96c83</code></a></summary>
<br>

```
proving/bundle/1        time:   [3.9909 s 4.0380 s 4.1140 s]
                        change: [-12.039% -9.7957% -7.2186%] (p = 0.00 < 0.05)
                        Performance has improved.
proving/bundle/2        time:   [3.9906 s 4.0130 s 4.0369 s]
                        change: [-12.653% -10.474% -8.1569%] (p = 0.00 < 0.05)
                        Performance has improved.
proving/bundle/3        time:   [5.8532 s 6.1311 s 6.4735 s]
                        change: [-4.0383% +0.8745% +6.0771%] (p = 0.79 > 0.05)
                        No change in performance detected.
proving/bundle/4        time:   [7.7428 s 8.2447 s 8.7707 s]
                        change: [-5.3507% +3.5264% +12.740%] (p = 0.47 > 0.05)
                        No change in performance detected.

verifying/bundle/1      time:   [39.497 ms 39.886 ms 40.390 ms]
                        change: [-14.938% -11.391% -7.8738%] (p = 0.00 < 0.05)
                        Performance has improved.
verifying/bundle/2      time:   [44.617 ms 49.110 ms 55.597 ms]
                        change: [+1.1374% +14.185% +30.292%] (p = 0.04 < 0.05)
                        Performance has regressed.
verifying/bundle/3      time:   [44.147 ms 44.392 ms 44.656 ms]
                        change: [-32.543% -26.449% -20.237%] (p = 0.00 < 0.05)
                        Performance has improved.
verifying/bundle/4      time:   [49.890 ms 50.577 ms 51.344 ms]
                        change: [+0.8118% +2.5337% +4.2452%] (p = 0.00 < 0.05)
```

</details>


halo2_proofs/src/plonk/circuit.rs
```diff
@@ -291,6 +337,25 @@ impl TableColumn {
...
+/// A challenge squeezed from transcript after advice columns at the phase have been committed.
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
+pub struct Challenge {
+    index: usize,
+    phase: Phase,
+}
+
+impl Challenge {
+    /// The index of this challenge.
+    pub fn index(&self) -> usize {
```
str4d, line 349: This should not be public:
```suggestion
    pub(crate) fn index(&self) -> usize {
```

halo2_proofs/src/plonk/circuit.rs
```diff
@@ -291,6 +337,25 @@ impl TableColumn {
...
+        self.index
+    }
+
+    /// The phase where this challenge will be squeezed in the end.
+    pub fn phase(&self) -> Phase {
```
str4d, line 354: This should not be public:
```suggestion
    pub(crate) fn phase(&self) -> Phase {
```

halo2_proofs/src/plonk/circuit.rs
```diff
@@ -376,6 +441,10 @@ pub trait Assignment<F: Field> {
         to: Option<Assigned<F>>,
     ) -> Result<(), Error>;
 
+    /// Get challenge value. If current phase is still earlier than requested one, a `None`
+    /// will be returned.
+    fn get_challenge(&mut self, challenge: Challenge) -> Option<F>;
```
str4d, line 446: This needs a changelog entry.

halo2_proofs/src/plonk/circuit.rs
```diff
@@ -376,6 +441,10 @@ pub trait Assignment<F: Field> {
         to: Option<Assigned<F>>,
     ) -> Result<(), Error>;
 
+    /// Get challenge value. If current phase is still earlier than requested one, a `None`
+    /// will be returned.
```
str4d, line 445:
```suggestion
    /// Queries the value of the given challenge.
    ///
    /// Returns `None` if the current synthesis phase is before the challenge can be queried.
```

halo2_proofs/src/plonk/circuit.rs
```diff
@@ -475,6 +546,13 @@ pub enum Expression<F> {
         /// Rotation of this query
         rotation: Rotation,
     },
+    /// This is a challenge
+    Challenge {
+        /// Challenge index
+        index: usize,
+        /// Challenge phase
+        phase: Phase,
+    },
```
str4d, line 555:
```suggestion
    Challenge(Challenge),
```

halo2_proofs/src/plonk/circuit.rs
```diff
@@ -465,6 +534,8 @@ pub enum Expression<F> {
         column_index: usize,
         /// Rotation of this query
         rotation: Rotation,
+        /// Phase of this advice column
+        phase: Phase,
```
str4d, line 538: This shouldn't be part of the public API (otherwise someone could mess with the queried expression to break the safety property we are trying to enforce).

We should instead make this entire enum struct opaque (which probably should have been the case for all of `Expression`, but it's easy to forget that expression structs and tuples have the visibility of the enum). This is probably best done in a separate PR first.

halo2_proofs/src/plonk/circuit.rs
```diff
@@ -493,8 +571,9 @@ impl<F: Field> Expression<F> {
         constant: &impl Fn(F) -> T,
         selector_column: &impl Fn(Selector) -> T,
         fixed_column: &impl Fn(usize, usize, Rotation) -> T,
-        advice_column: &impl Fn(usize, usize, Rotation) -> T,
+        advice_column: &impl Fn(usize, usize, Rotation, Phase) -> T,
         instance_column: &impl Fn(usize, usize, Rotation) -> T,
+        challenge: &impl Fn(usize, Phase) -> T,
```
str4d, line 576:
```suggestion
        challenge: &impl Fn(Challenge) -> T,
```

halo2_proofs/src/plonk/circuit.rs
```diff
@@ -493,8 +571,9 @@ impl<F: Field> Expression<F> {
         constant: &impl Fn(F) -> T,
         selector_column: &impl Fn(Selector) -> T,
         fixed_column: &impl Fn(usize, usize, Rotation) -> T,
-        advice_column: &impl Fn(usize, usize, Rotation) -> T,
+        advice_column: &impl Fn(usize, usize, Rotation, Phase) -> T,
```
str4d, line 574: These would then take the new `AdviceQuery` struct or whatever it gets called.

halo2_proofs/src/plonk/circuit.rs
```diff
@@ -1351,6 +1466,17 @@ impl<F: Field> ConstraintSystem<F> {
         tmp
     }
 
+    /// Allocate a new challenge in given phase
+    pub fn challenge_in(&mut self, phase: Phase) -> Challenge {
```
str4d, line 1470: Per my earlier comment, the challenge should not be interpreted as allocated "in" `phase`, but instead between `phase` and `phase + 1`. I don't have any great succinct naming suggestions right now, so maybe we should just be verbose:
```suggestion
    /// Requests a challenge that is usable after the given phase.
    pub fn challenge_usable_after(&mut self, phase: Phase) -> Challenge {
```
This is also where I think using type markers instead of an enum would give a better public API, so the `Phase` enum could be entirely private:
```suggestion
    /// Requests a challenge that is usable after the given phase.
    pub fn challenge_usable_after<P: Phase>(&mut self, phase: P) -> Challenge {
```
and then
```rust
trait SealedPhase {
    fn to_enum(self) -> PhaseEnum;
}
pub trait Phase: SealedPhase {}

enum FirstPhase {}
impl SealedPhase for FirstPhase {
    fn to_enum(self) -> PhaseEnum {
        PhaseEnum::First
    }
}
impl Phase for FirstPhase {}
```
`Challenge` could still store `PhaseEnum` internally to stop the `Phase` trait bound from propagating further than this API (which it has no need to, as phases don't need to appear anywhere else).

halo2_proofs/src/dev.rs
```diff
@@ -513,6 +520,10 @@ impl<F: FieldExt> MockProver<F> {
         let permutation = permutation::keygen::Assembly::new(n, &cs.permutation);
         let constants = cs.constants.clone();
 
+        let challenges = iter::repeat_with(|| F::random(OsRng))
+            .take(cs.num_challenges)
+            .collect();
```
str4d, line 525: This helps ensure the rest of `MockProver` continues to work, but it adds non-determinism to `MockProver` (that isn't present in the real prover). I'd like to avoid that if possible...

halo2_proofs/src/plonk/circuit.rs
han0110: Updated in [`902965c`](https://github.com/zcash/halo2/pull/593/commits/902965c385833ada61def929905c7431ff36e78b).

halo2_proofs/src/plonk/circuit.rs
han0110: Updated in [`902965c`](https://github.com/zcash/halo2/pull/593/commits/902965c385833ada61def929905c7431ff36e78b).

halo2_proofs/src/plonk/circuit.rs
han0110: Updated in [`315d0f0`](https://github.com/zcash/halo2/pull/593/commits/315d0f0e7842d6082d7fdd5a66de30c096dd2b49).

halo2_proofs/src/plonk/circuit.rs
han0110: Updated in [`902965c`](https://github.com/zcash/halo2/pull/593/commits/902965c385833ada61def929905c7431ff36e78b).

halo2_proofs/src/plonk/circuit.rs
han0110: Updated in [`902965c`](https://github.com/zcash/halo2/pull/593/commits/902965c385833ada61def929905c7431ff36e78b).

halo2_proofs/tests/plonk_api.rs
```diff
@@ -598,6 +598,15 @@ fn plonk_api() {
...
         num_selectors: 0,
+        num_challenges: 0,
+        advice_column_phase: [
+            First,
+            First,
+            First,
+            First,
+            First,
+        ],
+        challenge_phase: [],
```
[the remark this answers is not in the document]
han0110: To make this serialized vk unaltered, I implement `std::fmt::Debug` for `Advice`, `Any`, `Expression` and `PinnedConstraintSystem` in [`9e3c7d2`](https://github.com/zcash/halo2/pull/593/commits/9e3c7d2761f47fca6e229d4d5acc8e35233eb4fd), tho it seems to make the codebase a little bit messy :(

halo2_proofs/tests/plonk_api.rs
```diff
@@ -739,7 +756,9 @@ fn plonk_api() {
             (
                 Column {
                     index: 1,
-                    column_type: Advice,
+                    column_type: Advice {
+                        phase: First,
+                    },
```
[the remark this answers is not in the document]
han0110: Updated in [`9e3c7d2`](https://github.com/zcash/halo2/pull/593/commits/9e3c7d2761f47fca6e229d4d5acc8e35233eb4fd).

halo2_proofs/src/dev.rs
```diff
@@ -7,13 +7,14 @@ use std::iter;
 use std::ops::{Add, Mul, Neg, Range};
 
 use ff::Field;
+use rand_core::{block::BlockRng, OsRng, SeedableRng};
```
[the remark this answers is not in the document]
han0110: Removed in [`8b844b7`](https://github.com/zcash/halo2/pull/593/commits/8b844b759f1be814778d3f187ad1d201b5589772).

str4d:
> ### Handling advice column assignment in wrong phase
> 
> Instead of returning an error, currently `WitnessCollection` ignores the advice column assignment if phase doesn't match (like [this](https://github.com/zcash/halo2/blob/a330f520086e204f9ca2dfef0ec46802a2003f82/halo2_proofs/src/plonk/prover.rs#L195-L198)). The reason is to make sure the regions measured by `V1` planner to be same in different phase (To me it's like in key generation (phase 0) we still need to call `assign_advice`).
> 
> One approach is to call and check the value closure returns an `Err(Error::UnmatchedPhase)`, but not sure if this is desired, because it breaks the principle that all value closures passed to any `Region::assign_*` method will be called at most once.

Ignoring off-phase assignments is the correct behaviour. There is no way to construct a column with an arbitrary phase other than during `Circuit::configure`, meaning that by the time we reach `Circuit::synthesize` the backend knows precisely what columns and phases it can expect. Combined with the way challenges are exposed incrementally, we have exactly the desired safety.

What we could improve however is error handling. If a user accidentally calls `Region::assign_advice` with the wrong `(column, value)` pair, the error they get is likely not particularly helpful (namely `Error::Synthesis`), because during proving we currently have no way to propagate phase through values. This could change if we had an opaque `Value` type (see https://github.com/zcash/halo2/issues/509#issuecomment-1146025689), but we could also try to add a `MockProver` pass that figures out more debugging info.

halo2_proofs/src/plonk/circuit.rs
han0110: While I was trying sealed pattern for this, I found that `PhaseEnum` might still need to be public because `Advice` and `Any::Advice` has field of it, and some failure test cases need to create `VirtualCell` to be matched, which then needs to specify advice column's phase (even it doesn't use multi-phase feature).

Perhaps we might need to create `VirtualCell` in a way that it doesn't need `PhaseEnum`?

halo2_proofs/src/plonk/circuit.rs
han0110: Added in [`7786bf6`](https://github.com/zcash/halo2/pull/593/commits/7786bf60ec4641bf5593938a87f46048a79d2a97).

halo2_proofs/src/plonk/prover.rs
```diff
@@ -127,209 +122,272 @@ pub fn create_proof<
...
+            usable_rows: ..unusable_rows_start,
+            _marker: std::marker::PhantomData,
+        };
+
+        let mut advice_values = vec![None; meta.num_advice_columns];
+        let mut advice_polys = vec![None; meta.num_advice_columns];
+        let mut advice_cosets = vec![None; meta.num_advice_columns];
+        let mut advice_blinds = vec![None; meta.num_advice_columns];
+
+        for current_phase in [Phase::First, Phase::Second, Phase::Third] {
```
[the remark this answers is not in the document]
han0110: Updated in [`997cc01`](https://github.com/zcash/halo2/pull/593/commits/997cc012d87949c32056a8d12a702aadbeaa4b61).

halo2_proofs/src/plonk/prover.rs
```diff
@@ -127,209 +122,272 @@ pub fn create_proof<
...
-                    NR: Into<String>,
-                    N: FnOnce() -> NR,
-                {
-                    // Do nothing; we don't care about namespaces in this context.
+        fn get_challenge(&mut self, challenge: Challenge) -> Option<F> {
+            if self.current_phase > challenge.phase() {
+                Some(self.challenges[challenge.index()])
+            } else {
+                None
+            }
```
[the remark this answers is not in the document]
han0110: Updated in [`ecd7081`](https://github.com/zcash/halo2/pull/593/commits/ecd70819c571d44a960f29969c9d751811d91263).

halo2_proofs/examples/shuffle.rs
```diff
@@ -0,0 +1,336 @@
...
+}
+
+impl<const W: usize> MyConfig<W> {
+    fn configure<F: FieldExt>(meta: &mut ConstraintSystem<F>) -> Self {
+        // Phase 0
+        let [q_shuffle, q_first, q_last] = [(); 3].map(|_| meta.selector());
+        // Phase 1
+        let original = [(); W].map(|_| meta.advice_column_in(First));
+        let shuffled = [(); W].map(|_| meta.advice_column_in(First));
+        let [theta, gamma] = [(); 2].map(|_| meta.challenge_in(First));
```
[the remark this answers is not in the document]
han0110: Updated in [`997cc01`](https://github.com/zcash/halo2/pull/593/commits/997cc012d87949c32056a8d12a702aadbeaa4b61).

str4d:
> This could change if we had an opaque `Value` type

We now have this in #598.

halo2_proofs/src/dev/gates.rs
```diff
@@ -120,8 +120,9 @@ impl CircuitGates {
                             &util::format_value,
                             &|selector| format!("S{}", selector.0),
                             &|_, column, rotation| format!("F{}@{}", column, rotation.0),
-                            &|_, column, rotation| format!("A{}@{}", column, rotation.0),
+                            &|_, column, rotation, _| format!("A{}@{}", column, rotation.0),
```
daira, line 123: Perhaps include the phase if it is not phase 1.

halo2_proofs/src/dev/gates.rs
```diff
@@ -158,7 +159,7 @@ impl CircuitGates {
                                     .into_iter()
                                     .collect()
                             },
-                            &|_, column, rotation| {
+                            &|_, column, rotation, _| {
```
daira, line 162: Ditto.

halo2_proofs/src/plonk/circuit.rs
```diff
@@ -668,6 +777,60 @@ impl<F: Field> Expression<F> {
     }
 }
 
+impl<F: std::fmt::Debug> std::fmt::Debug for Expression<F> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        match self {
+            Expression::Constant(scalar) => f.debug_tuple("Constant").field(scalar).finish(),
+            Expression::Selector(_) => panic!("virtual selectors are removed during optimization"),
```
str4d, line 784: This panic is incorrect; there is no guarantee we won't use the `Debug` impl on `Expression` before optimization. This needs to be implemented.

halo2_proofs/src/plonk/circuit.rs
```diff
@@ -59,9 +59,51 @@ impl<C: ColumnType> PartialOrd for Column<C> {
     }
 }
 
+/// Phase of advice column
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Hash)]
+pub enum Phase {
```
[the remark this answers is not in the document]
daira: I think I also prefer the marker type approach.

halo2_proofs/src/plonk/circuit.rs
str4d: Yeah, this should be handled by changing the way `VirtualCell` works. If the dev tooling is the only thing placing this requirement on the public API, it's the dev tooling we should change.

str4d: I suspect the easiest change here is to add a `VirtualColumn` that allows setting the phase for advice, and then have an `impl From<Any> for VirtualColumn` etc. We could also keep the existing `From` impls, so fixed and instance checks wouldn't have to change.

halo2_proofs/src/dev.rs
str4d: Yeah, the hash chain looks fine to me. What precisely we do here is an implementation detail that can be altered at any time.

halo2_proofs/CHANGELOG.md
```diff
@@ -10,6 +10,18 @@ and this project adheres to Rust's notion of
 - `halo2_proofs::circuit::Value`, a more usable and type-safe replacement for
   `Option<V>` in circuit synthesis.
 - `impl Mul<F: Field> for &Assigned<F>`
+- `halo2_proofs::plonk`:
```
str4d, line 13: Now that we've released `halo2_proofs 0.2.0`, this PR needs to be rebased (because this section of the changelog is now part of `0.2.0`. Additionally, the commits should be squashed together so that the hard-coded proof binary is unaltered by the branch (rather than having multiple changes across several commits that together are a no-op but cost commit size).

halo2_proofs/src/circuit.rs
```diff
@@ -441,6 +443,11 @@ pub trait Layouter<F: Field> {
         row: usize,
     ) -> Result<(), Error>;
 
+    /// Queries the value of the given challenge.
+    ///
+    /// Returns `Value::unknown()` if the current synthesis phase is before the challenge can be queried.
+    fn get_challenge(&mut self, challenge: Challenge) -> Value<F>;
```
str4d, line 449: I don't see any reason for this to be mutable.
```suggestion
    fn get_challenge(&self, challenge: Challenge) -> Value<F>;
```

halo2_proofs/src/circuit.rs
```diff
@@ -506,6 +513,10 @@ impl<'a, F: Field, L: Layouter<F> + 'a> Layouter<F> for NamespacedLayouter<'a, F
         self.0.constrain_instance(cell, column, row)
     }
 
+    fn get_challenge(&mut self, challenge: Challenge) -> Value<F> {
```
str4d, line 516:
```suggestion
    fn get_challenge(&self, challenge: Challenge) -> Value<F> {
```

halo2_proofs/src/circuit/floor_planner/single_pass.rs
```diff
@@ -214,6 +214,10 @@ impl<'a, F: Field, CS: Assignment<F> + 'a> Layouter<F> for SingleChipLayouter<'a
         )
     }
 
+    fn get_challenge(&mut self, challenge: Challenge) -> Value<F> {
```
str4d, line 217:
```suggestion
    fn get_challenge(&self, challenge: Challenge) -> Value<F> {
```

halo2_proofs/src/circuit/floor_planner/v1.rs
```diff
@@ -199,6 +199,13 @@ impl<'p, 'a, F: Field, CS: Assignment<F> + 'a> Layouter<F> for V1Pass<'p, 'a, F,
         }
     }
 
+    fn get_challenge(&mut self, challenge: Challenge) -> Value<F> {
```
str4d, line 202:
```suggestion
    fn get_challenge(&self, challenge: Challenge) -> Value<F> {
```

halo2_proofs/src/circuit/floor_planner/v1.rs
```diff
@@ -199,6 +199,13 @@ impl<'p, 'a, F: Field, CS: Assignment<F> + 'a> Layouter<F> for V1Pass<'p, 'a, F,
...
+        match &mut self.0 {
```
str4d, line 203:
```suggestion
    fn get_challenge(&self, challenge: Challenge) -> Value<F> {
        match &self.0 {
```

halo2_proofs/src/plonk/keygen.rs
```diff
@@ -172,6 +172,10 @@ impl<F: Field> Assignment<F> for Assembly<F> {
         Ok(())
     }
 
+    fn get_challenge(&mut self, _: Challenge) -> Value<F> {
```
str4d, line 175:
```suggestion
    fn get_challenge(&self, _: Challenge) -> Value<F> {
```

halo2_proofs/src/dev/graph/layout.rs
```diff
@@ -490,6 +490,10 @@ impl<F: Field> Assignment<F> for Layout {
         Ok(())
     }
 
+    fn get_challenge(&mut self, _: Challenge) -> Value<F> {
```
str4d, line 493:
```suggestion
    fn get_challenge(&self, _: Challenge) -> Value<F> {
```

halo2_proofs/src/dev/graph.rs
```diff
@@ -157,6 +157,10 @@ impl<F: Field> Assignment<F> for Graph {
         Ok(())
     }
 
+    fn get_challenge(&mut self, _: Challenge) -> Value<F> {
```
str4d, line 160:
```suggestion
    fn get_challenge(&self, _: Challenge) -> Value<F> {
```

halo2_proofs/src/plonk/prover.rs
```diff
@@ -128,209 +127,274 @@ pub fn create_proof<
...
-                    _: V,
-                ) -> Result<(), Error>
-                where
-                    V: FnOnce() -> Value<VR>,
-                    VR: Into<Assigned<F>>,
-                    A: FnOnce() -> AR,
-                    AR: Into<String>,
-                {
-                    // We only care about advice columns here
+        fn get_challenge(&mut self, challenge: Challenge) -> Value<F> {
```
str4d, line 245:
```suggestion
        fn get_challenge(&self, challenge: Challenge) -> Value<F> {
```

halo2_gadgets/src/ecc/chip/mul_fixed/short.rs
```diff
@@ -580,7 +580,7 @@ pub mod tests {
                                     offset: 1,
                                 },
                                 cell_values: vec![(
-                                    ((Any::Advice, 5).into(), 0).into(),
+                                    ((Any::Advice { phase: First }, 5).into(), 0).into(),
```
str4d, line 583: These would become something like:
```suggestion
                                    ((VirtualColumn::Advice(First), 5).into(), 0).into(),
```

halo2_proofs/src/dev.rs
```diff
@@ -451,6 +454,10 @@ impl<F: Field + Group> Assignment<F> for MockProver<F> {
         Ok(())
     }
 
+    fn get_challenge(&mut self, challenge: Challenge) -> circuit::Value<F> {
```
str4d, line 457:
```suggestion
    fn get_challenge(&self, challenge: Challenge) -> circuit::Value<F> {
```

halo2_proofs/src/dev/cost.rs
```diff
@@ -118,6 +118,10 @@ impl<F: Field> Assignment<F> for Assembly {
         Ok(())
     }
 
+    fn get_challenge(&mut self, _: Challenge) -> Value<F> {
```
str4d, line 121:
```suggestion
    fn get_challenge(&self, _: Challenge) -> Value<F> {
```

halo2_proofs/src/plonk/circuit.rs
```diff
@@ -382,6 +470,11 @@ pub trait Assignment<F: Field> {
         to: Value<Assigned<F>>,
     ) -> Result<(), Error>;
 
+    /// Queries the value of the given challenge.
+    ///
+    /// Returns `Value::unknown()` if the current synthesis phase is before the challenge can be queried.
+    fn get_challenge(&mut self, challenge: Challenge) -> Value<F>;
```
str4d, line 476:
```suggestion
    fn get_challenge(&self, challenge: Challenge) -> Value<F>;
```

halo2_proofs/src/plonk/circuit.rs
```diff
@@ -1336,17 +1545,24 @@ impl<F: Field> ConstraintSystem<F> {
...
         self.num_advice_queries.push(0);
+        self.advice_column_phase.push(phase);
         tmp
     }
 
+    /// Allocate a new advice column at `First` phase
+    pub fn advice_column(&mut self) -> Column<Advice> {
+        self.advice_column_in(Phase::First)
+    }
+
```
str4d, line 1565: Nit: placing this above `advice_column_in` would reduce the diff size.

halo2_proofs/src/plonk/circuit.rs
```diff
@@ -1357,6 +1573,38 @@ impl<F: Field> ConstraintSystem<F> {
...
+        tmp
+    }
+
+    /// Assert phase-aware resources are allocated in order, and avoids any
+    /// phase to be skipped accidentally to cause unexpected issue in the future.
+    fn assert_no_phase_skipped(&self, phase: Phase, resource: &'static str) {
+        if let Some(previous_phase) = match phase {
+            Phase::Second => Some(Phase::First),
+            Phase::Third => Some(Phase::Second),
+            _ => None,
```
str4d, line 1594: If a new phase is added later, this will silently cause it to not be checked. Use an explicit match here instead:
```suggestion
            Phase::First => None,
            Phase::Second => Some(Phase::First),
            Phase::Third => Some(Phase::Second),
```

halo2_proofs/src/plonk/circuit.rs
```diff
@@ -1357,6 +1573,38 @@ impl<F: Field> ConstraintSystem<F> {
         tmp
     }
 
+    /// Requests a challenge that is usable after the given phase.
+    pub fn challenge_usable_after(&mut self, phase: Phase) -> Challenge {
+        self.assert_no_phase_skipped(phase, "Challenge");
```
str4d, line 1578: If we're requiring that phases are assigned in order, then we should be checking here that `phase` exists, not `previous_phase` (since this challenge only makes sense if there are columns in `phase`).

halo2_proofs/src/plonk/circuit.rs
han0110: I see, it makes sense to me, will try to add `VirtualColumn`.

The other issue I have just encountered is in `Expression::evaluate`, if we have `PhaseEnum` as private enum, what should we pass to the function to evaluate advice column (we can't pass `PhaseEnum` because `Expression::evaluate` is public)?

halo2_proofs/src/plonk/circuit.rs
han0110: Ah I didn't think about that. The assertion helper to check if specified phase exists is refactored in [`6051345`](https://github.com/zcash/halo2/pull/593/commits/6051345b999acca2c5110dce01ce3cfa58e321b7#diff-9ee6d800f38a8bebc2692cd214db8ed87d3a800e9f42c15a4c53f6f6fac9877cR1592-R1622).

halo2_proofs/CHANGELOG.md
han0110: Make sense, I have rebased and squashed all commits into 6 ones with suggestions and fixes above (and I resolved some minor comments, only leave those I'm not sure how to resolve for further discussion), hope it look cleaner now.

halo2_proofs/src/dev/gates.rs
han0110: Both updated in [`c87ba7a`](https://github.com/zcash/halo2/pull/593/commits/c87ba7a4a403d8c6f201134fece8f0af5375f8d4#diff-801c054c192c4d61fb5811b798f45a71d3dde952c5716ece6787f79c32d4f27d).

halo2_proofs/src/plonk/circuit.rs
str4d: Yes, I think introducing `*Query` structs is the way to go here; then future changes like this will be simpler, the evaluate interface will be simpler, and we can control what is exposed.

han0110: I see, I will try to work on these `*Query` structs first in a separate PR, and then get back to resolve this one.

halo2_proofs/examples/shuffle.rs
```diff
@@ -0,0 +1,351 @@
...
+impl<const W: usize> MyConfig<W> {
+    fn configure<F: FieldExt>(meta: &mut ConstraintSystem<F>) -> Self {
+        let [q_shuffle, q_first, q_last] = [(); 3].map(|_| meta.selector());
+        // Fist phase
+        let original = [(); W].map(|_| meta.advice_column_in(First));
+        let shuffled = [(); W].map(|_| meta.advice_column_in(First));
+        let [theta, gamma] = [(); 2].map(|_| meta.challenge_usable_after(First));
+        // Any phase skipped will cause a panic
+        assert!(catch_unwind_silent(|| meta.challenge_usable_after(Second)).is_err());
+        assert!(catch_unwind_silent(|| meta.advice_column_in(Third)).is_err());
```
daira, line 68: This would be appropriate for a test, but examples should just show the normal usage of the API, not demonstrate error conditions (especially error conditions we've chosen to expose as panics).

halo2_proofs/src/plonk/circuit.rs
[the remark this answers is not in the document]
daira: I'm also dubious about restricting this to three phases. Why three?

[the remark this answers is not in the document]
han0110: I thought about expose phase as `struct Phase(u8)` like rotation, but I was afraid it might look too flexible, so I choose to use enum, and set it to have 3 phases from my naive imagination.

But since we are going to expose the phases with exhaustive types, we might be able to define internal phase as `struct Phase(u8)` now? Then it should make it much easier to add more phases.

halo2_proofs/src/plonk/prover.rs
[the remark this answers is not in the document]
levs57: I suggest also adding Expression::Postprocess to ensure we can apply any black-box computations to challenges.

It is partially implemented here: https://github.com/levs57/halo2 , though it probably should be refactored a bit, because right now the fact that Expression:Postprocess is formed only from Expression::Constant, Expression::Challenge or Expression::Postprocess is checked in runtime, while we could achieve the same using wrapper struct PublicExpression with private field Expression.

Usage example here: https://github.com/levs57/halo2-liam-eagen-msm/blob/master/src/testing_stuff/challenge_postprocessing_test.rs

PR #597, opened 2022-06-08
Small performance improvements

str4d:


daira (approved):
utACK. I did not double-check the point addition formulae.

therealyingtong (approved):
utACK

ebfull (commented):
post-utACK

PR #610, opened 2022-06-23
Rework `halo2_proofs::plonk::BatchVerifier`

str4d:
`BatchVerifier` now manages the entire batch verification process. Individual proofs are verified on a threadpool, and the resulting MSMs are then batch-checked as before. The addition of parallelism here couples with zcash/halo2#608 to make parallelism less fine-grained and reduce the overhead of multi-threading.


ebfull (approved):

str4d:
[Force-pushed](https://github.com/zcash/halo2/compare/e38bdc444f1e5aab49898f94f1874a6c135ad3f6..a3ecf71909ea675dd0a86612d1bb254825d704d7) to fix a test.

daira (approved):
utACK

daira (approved):
re-utACK

PR #622, opened 2022-07-05
[MockProver] Check for instance values in gate queries.

### LINKED ISSUE #621: [MockProver] bug: `query_instance` always triggers `VerifyFailure::CellNotAssigned`.
In the current `MockProver`, `query_instance` always results in a `VerifyFailure::CellNotAssigned` when used in a gate with a selector. This is because we only look for assigned cells in the region where the gate was activated, and instance cells are not in any region.

https://github.com/zcash/halo2/blob/96d9bde905a20117b4350ffba0b0a6479aa63f0a/halo2_proofs/src/dev.rs#L577-L593

therealyingtong:
Closes #621.


daira (approved):
utACK

daira (approved):
utACK 76c658b321587da84d19e3e25072f08469ff044e

daira:
The clippy failures are not related to the PR.

PR #623, opened 2022-07-06
Allow `enable_equality` on `TableColumn`

therealyingtong:
The `TableColumn` should not be used in the `enable_constant` API, since that would introduce global constants not in the logical lookup table. However, there is no reason to disallow copy constraints on already-assigned table cells.


daira (approved):
utACK

ebfull (approved):

PR #625, opened 2022-07-07
halo2_proofs: Introduce `RegionLayouter::instance_value` method.

therealyingtong:
This allows us to access instance column values within a region. Previously, this was done only through `RegionLayouter::assign_advice_from_instance`.


daira (approved):
utACK. I can't see the logs of the failures on wasi, but there is nothing in this PR that could specifically affect wasi.

str4d:
I was reminded of this PR while trying to implement #738; given the existence of `Region` I appear to be blocked on this PR being seemingly incomplete. I've opened #740 for investigating.

PR #631, opened 2022-07-09
Removes `window_num_bits` < 8 in `decompose_word`

Janmajayamall:


halo2_gadgets/src/utilities.rs
```diff
@@ -186,8 +186,6 @@ pub fn decompose_word<F: PrimeFieldBits>(
     word_num_bits: usize,
     window_num_bits: usize,
 ) -> Vec<u8> {
```
therealyingtong, line 188: If we're returning `Vec<u8>`, I think we would have to at least check that `window_num_bits <= (1 << 8)`. However, in future I would like this function to return

```suggestion
) -> Vec<[bool; K]> {
```

therealyingtong (changes requested):

halo2_gadgets/src/utilities.rs
Janmajayamall: I am not sure how to do this since [array lengths don't support generics](https://github.com/rust-lang/rust/issues/43408). 

Shall we add the following assert statement for now 
`assert(window_num_bits <= (1 << 8))` 
Or should we try making it generic?

therealyingtong: We could add a const generic to

```
decompose_word<F: PrimeFieldBits, const K: usize>(
```

Janmajayamall: So something like the following? - 

`pub fn decompose_word<F: PrimeFieldBits, const K: usize>(
    word: &F,
    word_num_bits: usize,
) -> Vec<[bool; K]> {
    // Pad bits to multiple of window_num_bits
    let padding = (K - (word_num_bits % K)) % K;
    let bits: Vec<bool> = word
        .to_le_bits()
        .into_iter()
        .take(word_num_bits)
        .chain(std::iter::repeat(false).take(padding))
        .collect();
    assert_eq!(bits.len(), word_num_bits + padding);

    bits.chunks_exact(K)
        .map(|x| {
            let chunks = [false; K];
            chunks.copy_from_slice(x);
            chunks
        })
        .collect()
}`

PR #633, opened 2022-07-14
Only allow fixed queries at `Rotation::cur`

### LINKED ISSUE #585: Prohibit queries of fixed columns at non-cur rotations.
Rationale: There's [a technique](https://hackmd.io/@dJO3Nbl4RTirkR2uDM6eOA/BkoOdpwFF) that lets us make fixed columns almost free inside of recursion, which works ideally when fixed columns are not queried at non-cur rotations. I can't think of a situation where we would query a fixed column at an offset though? If we prohibited these queries entirely in the API, we could assume within recursion that we can deploy this trick.

therealyingtong:
Closes #585.

### DIFF
--- a/halo2_proofs/src/plonk/circuit.rs
+++ b/halo2_proofs/src/plonk/circuit.rs
@@ -1169,17 +1169,18 @@ impl<F: Field> ConstraintSystem<F> {
         panic!("get_instance_query_index called for non-existent query");
     }
 
-    pub(crate) fn get_any_query_index(&self, column: Column<Any>, at: Rotation) -> usize {
+    pub(crate) fn get_any_query_index(&self, column: Column<Any>) -> usize {
         match column.column_type() {
-            Any::Advice => {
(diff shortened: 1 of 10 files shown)


daira (approved):
utACK

halo2_proofs/src/plonk/circuit.rs
```diff
@@ -1169,17 +1169,18 @@ impl<F: Field> ConstraintSystem<F> {
         panic!("get_instance_query_index called for non-existent query");
     }
 
-    pub(crate) fn get_any_query_index(&self, column: Column<Any>, at: Rotation) -> usize {
+    pub(crate) fn get_any_query_index(&self, column: Column<Any>) -> usize {
```
str4d, line 1172: Is this really only used to query at `Rotation::cur()`? If so, we should document it as such.

str4d (approved):
ACK 41c87eac0f9766dc36af94291ae8537581b1272b. I confirmed that the `orchard` circuit is unaltered.

ebfull (approved):

PR #642, opened 2022-08-23
halo2_proofs: Reduce memory usage of `poly::Evaluator`

str4d:
`poly::Evaluator` stores all of the polynomials registered with it in memory for the duration of its existence. When evaluating an AST, it additionally caches the rotated polynomials in memory, and then chunks all of the rotated polynomials for parallel evaluation.

Previously, it stored a polynomial for every unique AST leaf, regardless of whether that leaf required a rotation or not. This resulted in the unrotated polynomials being stored twice in memory. However, the chunks simply refer to slices over cached polynomials, so we can reference the unrotated polynomials stored in `poly::Evaluator` instead of copies of them stored in the rotated polynomial `HashMap`. This strictly reduces memory usage during proving with no effect on correctness.

### DIFF
--- a/halo2_proofs/src/poly/evaluator.rs
+++ b/halo2_proofs/src/poly/evaluator.rs
@@ -159,14 +159,20 @@ impl<E, F: Field, B: Basis> Evaluator<E, F, B> {
         }
         let leaves = collect_rotations(ast);
 
-        // Produce the rotated polynomials.
+        // Produce a map from each leaf to the rotated polynomial it corresponds to, if
+        // any (or None if the leaf uses an unrotated polynomial).
         let rotated: HashMap<_, _> = leaves
             .iter()
             .cloned()
             .map(|leaf| {
                 (
                     leaf,
-                    B::rotate(domain, &self.polys[leaf.index], leaf.rotation),
+                    if leaf.rotation == Rotation::cur() {
+                        // We can use the polynomial as-is for this leaf.
+                        None
+                    } else {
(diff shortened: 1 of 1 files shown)


str4d:
Running `heaptrack` on the Orchard test `circuit::tests::round_trip` (which runs `MockProver` and then creates and verifies a single proof), I get the following memory usage measurements before and after this change:
```
❯ heaptrack --analyze heaptrack.circuit.before.zst | tail -n 7
total runtime: 5.17s.
bytes allocated in total (ignoring deallocations): 2.99GB (578.10MB/s)
calls to allocation functions: 1125475 (217693/s)
temporary memory allocations: 267839 (51806/s)
peak heap memory consumption: 132.75MB
peak RSS (including heaptrack overhead): 939.88MB
total memory leaked: 637.82KB

❯ heaptrack --analyze heaptrack.circuit.after.zst | tail -n 7
total runtime: 5.20s.
bytes allocated in total (ignoring deallocations): 2.95GB (567.03MB/s)
calls to allocation functions: 1125346 (216412/s)
temporary memory allocations: 266765 (51300/s)
peak heap memory consumption: 96.05MB
peak RSS (including heaptrack overhead): 680.80MB
total memory leaked: 684.67KB
```

So for the Orchard circuit, this is a 27.6% reduction in both peak heap memory consumption and peak RSS.

daira (approved):
utACK

therealyingtong (approved):
utACK

Issue #643, opened 2022-08-24
Create separate `VerifyingKey` structures for pre- and post-optimisation circuits


str4d:
The current `VerifyingKey` represents the post-optimisation circuit; however it does so in a way that isn't reproducible without access to the circuit. This causes issues with e.g. defining a stable serialization format (#449).

Separately, we also need a well-defined `VerifyingKey` suitable for passing into a circuit gadget for verifying a proof (#435, #436).

We should refactor the `VerifyingKey` structure to separate pre- and post-optimisation concerns, to make it very clear what is needed in the latter case and enable us to start defining more concretely how we want to encode it in the various consuming contexts.

str4d:
We made some exploratory code changes in Halo 2 Office Hours today, to figure out how to approach teh post-optimisation `VerifyingKey` structure. The resulting changes are in https://github.com/zcash/halo2/pull/781, along with the conclusions we drew from them about how the real refactor should work.

daira:
The verification key serialization should include the number of public input elements.

PR #661, opened 2022-09-16
VerifyingKey Serialisation

### LINKED ISSUE #449: `ViewingKey` serialization is not round-trip compatible

nalinbhardwaj:
Closes #449 by storing selectors (the input to `compress_selectors`) as part of the serialised vkey. On reads, reruns the selector compression on the `cs` object. This also partly completes #643 (by creating the "post' version as the default, afaict no place _needs_ serialisation for the pre part).

Tested using the circuits from https://github.com/nalinbhardwaj/zordle and https://github.com/icemelon/halo2-examples/tree/master/src/fibonacci. This is also extremely useful for faster wasm provers!


halo2_proofs/src/plonk.rs
```diff
@@ -47,15 +48,70 @@ pub struct VerifyingKey<C: CurveAffine> {
...
+    }
+
+    /// Reads a verification key from a buffer.
+    pub fn read<R: io::Read, ConcreteCircuit: Circuit<C::Scalar>>(
+        reader: &mut R,
+        params: &Params<C>,
+    ) -> io::Result<Self> {
+        let (domain, cs, _) = keygen::create_domain::<C, ConcreteCircuit>(params);
+
+        let fixed_commitments: Vec<_> = (0..cs.num_fixed_columns)
```
jonathanpwang, line 83: Unfortunately I believe this may not work because the `num_fixed_columns` of the constraint system is changed by the `compress_selectors` function. I think you need to store the number of fixed columns as a separate parameter within the vkey to fix this.

nalinbhardwaj: Cherry-picked @jonathanpwang's fixing commit to this PR

parazyd:
Are there any plans to merge this?

halo2_proofs/src/plonk.rs
```diff
@@ -47,14 +48,82 @@ pub struct VerifyingKey<C: CurveAffine> {
     cs_degree: usize,
     /// The representative of this `VerifyingKey` in transcripts.
     transcript_repr: C::Scalar,
+    selectors: Vec<Vec<bool>>,
 }
 
 impl<C: CurveAffine> VerifyingKey<C> {
+    /// Writes a verifying key to a buffer.
+    pub fn write<W: io::Write>(&self, writer: &mut W) -> io::Result<()> {
+        writer.write_all(&(self.fixed_commitments.len() as u32).to_be_bytes())?;
```
daira, line 57: Please write this as little-endian.

halo2_proofs/src/plonk.rs
```diff
@@ -47,14 +48,82 @@ pub struct VerifyingKey<C: CurveAffine> {
     cs_degree: usize,
     /// The representative of this `VerifyingKey` in transcripts.
     transcript_repr: C::Scalar,
+    selectors: Vec<Vec<bool>>,
 }
 
 impl<C: CurveAffine> VerifyingKey<C> {
+    /// Writes a verifying key to a buffer.
+    pub fn write<W: io::Write>(&self, writer: &mut W) -> io::Result<()> {
```
daira, line 56: Please write a zero version byte first. On read, error if this byte is not zero.

halo2_proofs/src/plonk.rs
```diff
@@ -47,14 +48,82 @@ pub struct VerifyingKey<C: CurveAffine> {
     cs_degree: usize,
     /// The representative of this `VerifyingKey` in transcripts.
     transcript_repr: C::Scalar,
+    selectors: Vec<Vec<bool>>,
```
daira, line 51: I think this should be the compressed selectors. There's no reason to keep the uncompressed selectors in the verifying key, and doing so makes the representation dependent on the selector compression algorithm when it needn't be.

halo2_proofs/src/plonk.rs
```diff
@@ -47,14 +48,82 @@ pub struct VerifyingKey<C: CurveAffine> {
...
+                for (i, selector_idx) in selector.iter_mut().enumerate() {
+                    let byte_index = i / 8;
+                    let bit_index = i % 8;
+                    *selector_idx = (selector_bytes[byte_index] >> bit_index) & 1 == 1;
+                }
+                Ok(selector)
+            })
+            .collect::<Result<Vec<Vec<bool>>, &str>>()
+            .unwrap();
+        let (cs, _) = cs.compress_selectors(selectors.clone());
```
daira, line 110: Instead of compressing the selectors here, read the compressed ones.

halo2_proofs/src/plonk.rs
```diff
@@ -47,14 +48,82 @@ pub struct VerifyingKey<C: CurveAffine> {
...
+    pub fn write<W: io::Write>(&self, writer: &mut W) -> io::Result<()> {
+        writer.write_all(&(self.fixed_commitments.len() as u32).to_be_bytes())?;
+        for commitment in &self.fixed_commitments {
+            writer.write_all(commitment.to_bytes().as_ref())?;
+        }
+        self.permutation.write(writer)?;
+
+        // write self.selectors
+        for selector in &self.selectors {
+            let mut selector_bytes = vec![0u8; selector.len() / 8 + 1];
```
daira, line 65:
```suggestion
            let mut selector_bytes = vec![0u8; (selector.len()+7) / 8];
```

halo2_proofs/src/plonk.rs
```diff
@@ -47,14 +48,82 @@ pub struct VerifyingKey<C: CurveAffine> {
...
+            .map(|_| C::read(reader))
+            .collect::<Result<_, _>>()?;
+
+        let permutation = permutation::VerifyingKey::read(reader, &cs.permutation)?;
+
+        // read selectors
+        let selectors: Vec<Vec<bool>> = vec![vec![false; params.n as usize]; cs.num_selectors]
+            .into_iter()
+            .map(|mut selector| {
+                let mut selector_bytes = vec![0u8; selector.len() / 8 + 1];
```
daira, line 97:
```suggestion
                let mut selector_bytes = vec![0u8; (selector.len() + 7) / 8];
```

halo2_proofs/src/plonk.rs
```diff
@@ -47,14 +48,82 @@ pub struct VerifyingKey<C: CurveAffine> {
...
+
+    /// Reads a verification key from a buffer.
+    pub fn read<R: io::Read, ConcreteCircuit: Circuit<C::Scalar>>(
+        reader: &mut R,
+        params: &Params<C>,
+    ) -> io::Result<Self> {
+        let (domain, cs, _) = keygen::create_domain::<C, ConcreteCircuit>(params);
+        let mut num_fixed_columns_be_bytes = [0u8; 4];
+        reader.read_exact(&mut num_fixed_columns_be_bytes)?;
+        let num_fixed_columns = u32::from_be_bytes(num_fixed_columns_be_bytes);
```
daira, line 85:
```suggestion
        let mut num_fixed_columns_le_bytes = [0u8; 4];
        reader.read_exact(&mut num_fixed_columns_le_bytes)?;
        let num_fixed_columns = u32::from_le_bytes(num_fixed_columns_le_bytes);
```

halo2_proofs/src/plonk.rs
```diff
@@ -47,14 +48,82 @@ pub struct VerifyingKey<C: CurveAffine> {
...
+
+        let fixed_commitments: Vec<_> = (0..num_fixed_columns)
+            .map(|_| C::read(reader))
+            .collect::<Result<_, _>>()?;
+
+        let permutation = permutation::VerifyingKey::read(reader, &cs.permutation)?;
+
+        // read selectors
```
daira, line 93:
```suggestion
        // read selectors
        let mut num_selectors_le_bytes = [0u8; 4];
        reader.read_exact(&mut num_selectors_le_bytes)?;
        if cs.num_selectors != u32::from_le_bytes(num_selectors_le_bytes) {
            return io::Error::new(io::ErrorKind::InvalidData, "unexpected number of selectors");
        }
```

halo2_proofs/src/plonk.rs
```diff
@@ -47,14 +48,82 @@ pub struct VerifyingKey<C: CurveAffine> {
...
 impl<C: CurveAffine> VerifyingKey<C> {
+    /// Writes a verifying key to a buffer.
+    pub fn write<W: io::Write>(&self, writer: &mut W) -> io::Result<()> {
+        writer.write_all(&(self.fixed_commitments.len() as u32).to_be_bytes())?;
+        for commitment in &self.fixed_commitments {
+            writer.write_all(commitment.to_bytes().as_ref())?;
+        }
+        self.permutation.write(writer)?;
+
+        // write self.selectors
```
daira, line 63:
```suggestion
        // write self.selectors
        writer.write_all(&(u32::try_from(self.selectors.len()).unwrap()).to_le_bytes())?;
```

halo2_proofs/src/plonk.rs
```diff
@@ -47,14 +48,82 @@ pub struct VerifyingKey<C: CurveAffine> {
...
+        let permutation = permutation::VerifyingKey::read(reader, &cs.permutation)?;
+
+        // read selectors
+        let selectors: Vec<Vec<bool>> = vec![vec![false; params.n as usize]; cs.num_selectors]
+            .into_iter()
+            .map(|mut selector| {
+                let mut selector_bytes = vec![0u8; selector.len() / 8 + 1];
+                reader
+                    .read_exact(&mut selector_bytes)
+                    .expect("unable to read selector bytes");
```
daira, line 100:
```suggestion
                reader.read_exact(&mut selector_bytes)?;
```

halo2_proofs/src/plonk/permutation.rs
```diff
@@ -75,6 +78,23 @@ pub(crate) struct VerifyingKey<C: CurveAffine> {
     commitments: Vec<C>,
 }
 
+impl<C: CurveAffine> VerifyingKey<C> {
+    pub(crate) fn write<W: io::Write>(&self, writer: &mut W) -> io::Result<()> {
```
daira, line 82:
```suggestion
    pub(crate) fn write<W: io::Write>(&self, writer: &mut W) -> io::Result<()> {
        writer.write_all(&(u32::try_from(self.commitments.len()).unwrap()).to_le_bytes())?;
```

halo2_proofs/src/plonk/permutation.rs
```diff
@@ -75,6 +78,23 @@ pub(crate) struct VerifyingKey<C: CurveAffine> {
...
+        for commitment in &self.commitments {
+            writer.write_all(commitment.to_bytes().as_ref())?;
+        }
+
+        Ok(())
+    }
+
+    pub(crate) fn read<R: io::Read>(reader: &mut R, argument: &Argument) -> io::Result<Self> {
```
daira, line 90:
```suggestion
    pub(crate) fn read<R: io::Read>(reader: &mut R, argument: &Argument) -> io::Result<Self> {
    ```suggestion
        let mut num_commitments_le_bytes = [0u8; 4];
        reader.read_exact(&mut num_commitments_le_bytes)?;
        if argument.columns.len() != u32::from_le_bytes(num_commitments_le_bytes) {
            return io::Error::new(io::ErrorKind::InvalidData, "unexpected number of column commitments");
        }
```
```

File: halo2_proofs/src/plonk.rs [hunk 14]
```diff
@@ -47,14 +48,82 @@ pub struct VerifyingKey<C: CurveAffine> {
...
+                for (i, selector_idx) in selector.iter_mut().enumerate() {
+                    let byte_index = i / 8;
+                    let bit_index = i % 8;
+                    *selector_idx = (selector_bytes[byte_index] >> bit_index) & 1 == 1;
+                }
+                Ok(selector)
+            })
+            .collect::<Result<Vec<Vec<bool>>, &str>>()
+            .unwrap();
```
daira, line 109:
```suggestion
            .collect::<io::Result<_>>()?
```

halo2_proofs/src/plonk/permutation.rs
```diff
@@ -75,6 +78,23 @@ pub(crate) struct VerifyingKey<C: CurveAffine> {
...
+        let commitments = (0..argument.columns.len())
+            .map(|_| C::read(reader))
+            .collect::<Result<Vec<_>, _>>()?;
```
daira, line 93:
```suggestion
        let commitments: Vec<_> = (0..argument.columns.len())
            .map(|_| C::read(reader))
            .collect::<io::Result<_>>()?;
```

halo2_proofs/src/plonk.rs
```diff
@@ -47,14 +48,82 @@ pub struct VerifyingKey<C: CurveAffine> {
...
+        params: &Params<C>,
+    ) -> io::Result<Self> {
+        let (domain, cs, _) = keygen::create_domain::<C, ConcreteCircuit>(params);
+        let mut num_fixed_columns_be_bytes = [0u8; 4];
+        reader.read_exact(&mut num_fixed_columns_be_bytes)?;
+        let num_fixed_columns = u32::from_be_bytes(num_fixed_columns_be_bytes);
+
+        let fixed_commitments: Vec<_> = (0..num_fixed_columns)
+            .map(|_| C::read(reader))
+            .collect::<Result<_, _>>()?;
```
daira, line 89:
```suggestion
            .collect::<io::Result<_>>()?;
```

daira (changes requested):
At least the change to little-endian for the length of fixed commitments is blocking. [Done.]

halo2_proofs/src/plonk.rs
```diff
@@ -47,14 +48,82 @@ pub struct VerifyingKey<C: CurveAffine> {
...
 }
 
 impl<C: CurveAffine> VerifyingKey<C> {
+    /// Writes a verifying key to a buffer.
+    pub fn write<W: io::Write>(&self, writer: &mut W) -> io::Result<()> {
+        writer.write_all(&(self.fixed_commitments.len() as u32).to_be_bytes())?;
+        for commitment in &self.fixed_commitments {
+            writer.write_all(commitment.to_bytes().as_ref())?;
+        }
+        self.permutation.write(writer)?;
```
daira, line 61: This should also write a length field.

daira (dismissed):
Reviewed; rebased to fix conflicts and add my changes in the last commit. This will need tests.

halo2_proofs/src/plonk.rs
daira: I see that I said the same thing in https://github.com/zcash/halo2/issues/449#issuecomment-1031035632 . I'm nothing if not consistent, even though I'd completely forgotten making that comment!

jonathanpwang: I agree with you. In retrospect serialization should definitely use the compressed selectors. Ideally `VerifyingKey::read` should not need to know the type of `ConcreteCircuit` or even be supplied with `params` (the latter is only used to generate `domain, cs` and get `k, n` I believe). 

Right now I have to create a `ConcreteCircuit` ahead of time and then trick Rust to do some type inference to call `read`.

jonathanpwang (commented):
LGTM!

I leave it up to you whether you want to make the change to compressed selectors before merging.

str4d:
We discussed this PR in Halo 2 Office Hours today, and decided that we will not merge it as-is. The core issue is that the serialization format being defined here is tightly coupled to the current implementation, and will be very difficult to retain support for after the planned optimisation / `FloorPlanner` refactorings occur.

For example, during keygen we extend the set of fixed column commitments to include commitments to the post-selector-combining selector columns:
https://github.com/zcash/halo2/blob/5678a506cbec60d593a21ff618d09bed48abf7f2/halo2_proofs/src/plonk/keygen.rs#L222-L227

Then in `VerifyingKey::read` we read the post-selector-combining fixed commitments, read the pre-selector-combining selector definitions, re-compress the selector columns, and _throw away the resulting selector polynomials_ (without checking that the effect of the compression on the constraint system's gates was correct or consistent with the parsed fixed commitments):
https://github.com/zcash/halo2/blob/58e797da937c5193ce43ad333f22ead664f1ba61/halo2_proofs/src/plonk.rs#L101-L103
https://github.com/zcash/halo2/blob/58e797da937c5193ce43ad333f22ead664f1ba61/halo2_proofs/src/plonk.rs#L129

So as-is, the implicit serialization format is cutting half-way through the selector-combining optimisation, instead of picking a side. This specific issue could for instance be averted by not serializing the fixed commitments for the selectors, and always re-deriving them afterwards. Alternatively we could re-derive them and check they match, but then we're doing the same computational work during parsing with no benefit to encoding size. Solving this specific issue is a blocker on this PR (and it would be good to check for any other similar specific issues).

But if we _did_ make that fix, we are then in a situation where the implicit serialization format is still cutting part-way through the logic that we want the `FloorPlanner` to be responsible for. Maintaining compatibility with this format would place strong constraints on how we designed the new `FloorPlanner` APIs, and probably require special-casing the selector-combining optimisation near-indefinitely (rather than it becoming an implementation detail of specific floor planners).

The most productive route forward would I think be to work on creating a post-optimisation `VerifyingKey` struct (the second half of #643), and then defining an encoding format for it. This can be done without blocking on the `FloorPlanner` refactoring, as it's about defining the type that would exist on the output side of the `FloorPlanner` rather than anything within its boundaries. It would require defining canonical encodings for polynomials as well as the polynomial ASTs we use to represent gates; we probably want to look at https://crates.io/crates/postcard for actually producing the encoding format.

jonathanpwang:
I'm curious - how do you feel about just using `#[derive(Serialize, Deserialize)]`? That would avoid all of this. I have yet to benchmark but I would guess the `bincode` serialization/deserialization will be faster than what's currently implemented. This would involve the `PrimeField` to also implement `Serialize, Deserialize`.

str4d:
> I'm curious - how do you feel about just using `#[derive(Serialize, Deserialize)]`? That would avoid all of this.

That would not solve the above problem, and in fact would make the implicit encoding even more tightly coupled to the current implementation.

Once the types issue is addressed however, and a `VerifyingKey` type exists that contains precisely what should be encoded in precisely the internal form required for circuit correctness, then that could be an approach taken (using `postcard` as I suggested above).

str4d:
> The most productive route forward would I think be to work on creating a post-optimisation VerifyingKey struct (the second half of https://github.com/zcash/halo2/issues/643), and then defining an encoding format for it.

I'm going to try working on this some time next week.

str4d:
> > The most productive route forward would I think be to work on creating a post-optimisation VerifyingKey struct (the second half of #643), and then defining an encoding format for it.
> 
> I'm going to try working on this some time next week.

We did the first phase of this in Halo 2 Office Hours today. The resulting exploratory changes are in #781, along with the conclusions we drew from them about how the real refactor should work.

parazyd:
@str4d For retaining compatibility and more comfortable API usage for users which do not use `serde`, why not define the encoding and decoding format explicitly in the `halo2` crate?

It seems like it would give you finer-grained control over the format, and it is also low overhead for people that use different serialization methods (i.e. they only have to wrap the defined halo2 encoding format in whatever trait their serialization library needs).

hanzhichao:
any update about the provingkey and verifyingkey serialization?

themighty1:
@hanzhichao , PSE fork supports key serialization https://github.com/privacy-scaling-explorations/halo2/blob/main/book/src/user/experimental-features.md#provingkey--verifyingkey-deserialization-and-serdeformat

Issue #614, opened 2022-06-24
Initialise unassigned cells to random values (instead of all-zeros)


therealyingtong:
Currently, we initialise unassigned cells to zero. (Note: we do not CONSTRAIN them to be zero.) This has a few undesirable effects on developer experience:

- when some cell is required to be zero, the developer may simply choose to leave it unassigned, and forget to actually constrain it to be zero;
- currently, unassigned cells will pass all lookups with a `0` value in its table. This obscures errors where the lookup argument is accidentally enabled on an unintended row.

In witness collection:
https://github.com/zcash/halo2/blob/96d9bde905a20117b4350ffba0b0a6479aa63f0a/halo2_proofs/src/plonk/prover.rs#L267-L277

In keygen:
https://github.com/zcash/halo2/blob/main/halo2_proofs/src/plonk/keygen.rs#L203-L218

Let's randomise these values instead.

greenhat:
I'd like to take this one.

greenhat:
@therealyingtong I took a shot at it (#663), and I'm stuck with two questions. I'll copy them from #663:
1. for `WitnessCollection::advice`, there is what looks like a similar operation already implemented later in  
https://github.com/greenhat/halo2/blob/fd8a54a96dca405209588a6d5aa3ac8bbfc74eaf/halo2_proofs/src/plonk/prover.rs#L299-L304 . If so, we probably don't need both of them, right?
2. Assigning random values to unassigned cells in `Assembly::fixed` breaks the `plonk_api` test. I'm having trouble figuring out why it fails. Any ideas?

therealyingtong:
@greenhat 

1. Yes, unusable advice cells are already set to random values;
2. We should only initialise unassigned **advice** cells to random values. Fixed cells' values are set in keygen, and encode the circuit logic: changing fixed cell values would change the circuit definition.

greenhat:
> @greenhat
> 
>     1. Yes, unusable advice cells are already set to random values;

Ok, so I'm scraping my changes to `WitnessCollection`.
> 
>     2. We should only initialise unassigned **advice** cells to random values. Fixed cells' values are set in keygen, and encode the circuit logic: changing fixed cell values would change the circuit definition.

Thanks for the explanation. I initially changed the `Assembly::fixed` because you mentioned it in `In keygen: ...` in the PR description. I'm scraping my changes to keygen as well. So my PR can be closed, right?
[...]

PR #663, opened 2022-09-18
Initialise unassigned cells to random values #614;

greenhat:
This PR is WIP on #614

This PR adds `Group::empty_lagrange_assigned_with_unused_random()` and used it in `WitnessCollection::advice` and `Assembly::fixed` in `plonk::keygen::keygen_vk`.

Questions:
- for `WitnessCollection::advice`, there is what looks like a similar operation already implemented later in  
https://github.com/greenhat/halo2/blob/fd8a54a96dca405209588a6d5aa3ac8bbfc74eaf/halo2_proofs/src/plonk/prover.rs#L299-L304
- Assigning random values to unassigned cells in `Assembly::fixed` breaks the `plonk_api` test. I'm having trouble figuring out why it fails.


halo2_proofs/src/poly/domain.rs
```diff
@@ -201,6 +202,30 @@ impl<G: Group> EvaluationDomain<G> {
...
+                    if i < unusable_rows_start as u64 {
+                        G::group_zero().into()
+                    } else {
+                        G::random(&mut rng).into()
+                    }
+                })
+                .collect(),
+            _marker: PhantomData,
+        }
+    }
```
therealyingtong, line 227: This should initialise the polynomial to random values on all rows (not just unusable rows). Assigning a cell would overwrite the initial random value; unassigned cells will remain set to the initial random value.

```suggestion
    /// Returns a random polynomial in the Lagrange coefficient basis,
    /// with deferred inversions.
    pub(crate) fn random_lagrange_assigned(
        &self,
        mut rng: impl RngCore,
    ) -> Polynomial<Assigned<G>, LagrangeCoeff>
    where
        G: Field,
    {
        Polynomial {
            values: (0..self.n)
                .map(|_| G::random(&mut rng).into())
                .collect(),
            _marker: PhantomData,
        }
    }
```

therealyingtong (changes requested):

therealyingtong (approved):
LGTM

greenhat:
@therealyingtong I fixed code formatting to make CI happy. Could you please rerun the CI?

therealyingtong:
It just occurred to me that the unusable advice rows are now "double-blinded": once in initialisation, and again after synthesis:

https://github.com/greenhat/halo2/blob/fd8a54a96dca405209588a6d5aa3ac8bbfc74eaf/halo2_proofs/src/plonk/prover.rs#L299-L304

This has no effects on soundness and should have negligible effects on performance. Do we want to remove this redundant randomisation, or leave it + document it?

moreSocratic:
I think the `MockProver` also needs to be updated?
https://github.com/zcash/halo2/blob/6f692c0e53c50f18aae09ca731d144d4bc9e951b/halo2_proofs/src/dev.rs#L95

therealyingtong:
> I think the MockProver also needs to be updated?

You're right! @moreSocratic

therealyingtong (changes requested):
Corresponding updates to `MockProver` initialisation are needed.

greenhat:
> I think the `MockProver` also needs to be updated?
> 
> https://github.com/zcash/halo2/blob/6f692c0e53c50f18aae09ca731d144d4bc9e951b/halo2_proofs/src/dev.rs#L95

I tried a straight `CellValue::Unassigned => Value::Real(F::random(OsRng)),` but this makes a few tests fail. I found that this `From` implementation is used not only on `MockProver::advice` but also on `MockProver::fixed` cells, which I think might be the cause of tests failing. I'm thinking of "limiting" the randomization only to `MockProver::advice` cells. Am I on the right track?

PR #665, opened 2022-09-20
Add halo_proof as a dependency

murksombra:
Just an addition to Cargo.toml to add:
```
 halo2_proofs = "0.2.0"
```


str4d:
We can't add `halo2_proofs` as a dependency of itself.

PR #667, opened 2022-09-23
Don't use `include_bytes!` for `plonk_api` test

L-as:
The current working directory for a test always seems to be the root of the crate, hence `./tests/plonk_api_proof.bin` is always the correct path.


therealyingtong (approved):
utACK

PR #670, opened 2022-09-27
End to End example from Halo2 Learning Group

AtHeartEngineer:


therealyingtong (approved):
ACK, I rendered the book locally and ran `cargo run --example e2e-example 2 3`.

PR #671, opened 2022-09-28
`lookup::prover`: Compress expressions and cosets in `commit_permuted`

therealyingtong:
The `theta` challenge is used in the lookup argument to compress the input and table expressions. This compression can be done fully in the `commit_permuted` step, after which the original uncompressed expressions are no longer needed.


daira (approved):
utACK modulo clippy lints.

PR #685, opened 2022-10-21
[MockProver] failure::FailureLocation: Handle zero-area regions.

### LINKED ISSUE #684: MockProver.verify() panics
(Using halo2 main branch with the most recent changes) 
I made a trivial change to my circuit and MockProver.verify() now panics in my tests.

I made a repro branch with 2 top-level commits - before the last commit `cargo test test_circuit` succeeded. After a small change to the circuit `cargo test test_circuit` causes a panic from inside MockProver.verify() 
https://github.com/themighty1/tlsn/commits/mockprover_panic

The test which triggers the panic is located here:
https://github.com/themighty1/tlsn/blob/abdcbb98d7c66f61326ee8a755530eba4845b875/authdecode/src/halo2_backend/prover.rs#L239

The test feeds the wrong public inputs to the circuit and expects that verification fails, not panics.

therealyingtong:
Closes #684

### DIFF
--- a/halo2_proofs/src/dev/failure.rs
+++ b/halo2_proofs/src/dev/failure.rs
@@ -87,13 +87,17 @@ impl FailureLocation {
             .iter()
             .enumerate()
             .find(|(_, r)| {
-                let (start, end) = r.rows.unwrap();
-                // We match the region if any input columns overlap, rather than all of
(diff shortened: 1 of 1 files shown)


daira (approved):
utACK

str4d (approved):

PR #687, opened 2022-10-26
Optimize multiopen verifier for many columns

naure:
The multiopen verification time is supposed to be linear in the number of commitments, but the implementation is quadratic. This becomes significant in the 100+ columns / commitments.

The code uses Horner’s rule to merge the commitments, but this is quadratic in an MSM structure. See the functions `scale / append_term`.

[multiopen/verifier.rs](https://github.com/zcash/halo2/blob/ec9dcefe9103fc23c13e8195120419d4d2f232a6/halo2_proofs/src/poly/multiopen/verifier.rs#L52)

IPA Verifier quad work

The optimization is to scale individual terms by powers of `x_1` instead. See benchmark below ([code here](https://github.com/zcash/halo2/pull/687/commits/54f1e34d21eb8b8958dddb3e5f5cea90def380bf)).

image


halo2_proofs/Cargo.toml
```diff
@@ -63,6 +63,7 @@ criterion = "0.3"
 gumdrop = "0.8"
 proptest = "1"
 rand_core = { version = "0.6", default-features = false, features = ["getrandom"] }
+ark-std = { version = "0.3", features = ["print-trace"] }
```
naure, line 66: This is for demonstration and can be removed.

halo2_proofs/src/poly/multiopen.rs
```diff
@@ -360,6 +360,9 @@ fn test_roundtrip() {
     }
 }
 
+#[cfg(test)]
+mod bench;
```
naure, line 364: This is for demonstration and can be removed.

halo2_proofs/src/poly/multiopen/verifier.rs
```diff
@@ -39,34 +39,40 @@ where
...
+                *set_eval += (*eval) * (*x_1_power);
             }
+            *x_1_power *= *x_1;
         };
 
         // Each commitment corresponds to evaluations at a set of points.
         // For each set, we collapse each commitment's evals pointwise.
-        for commitment_data in commitment_map.into_iter() {
+        // Run in order of increasing x_1 powers.
+        for commitment_data in commitment_map.into_iter().rev() {
```
daira, line 75: Ah I see. The new definition of `accumulate` would multiply by powers in the opposite order, but then this `.rev()` compensates.

daira (approved):
utACK

str4d (approved):
ACK. I verified that the `orchard` tests pass after this change.

PR #689, opened 2022-11-13
rm useless code

zhiqiangxu:


halo2_proofs/src/plonk/prover.rs
```diff
@@ -97,7 +97,6 @@ pub fn create_proof<
             let mut instance_commitments =
                 vec![C::identity(); instance_commitments_projective.len()];
             C::Curve::batch_normalize(&instance_commitments_projective, &mut instance_commitments);
-            let instance_commitments = instance_commitments;
```
str4d, line 100: This isn't quite useless; it shadows the old `instance_commitments` binding to make it immutable. An alternative change that preserves this behaviour would be:
```suggestion
            let instance_commitments = {
                let mut normalized = vec![C::identity(); instance_commitments_projective.len()];
                C::Curve::batch_normalize(&instance_commitments_projective, &mut normalized);
                normalized
            };
```

zhiqiangxu:
Keep the rebinding there as a mark of immutable.

PR #693, opened 2022-11-23
Add secp/secq curve support

vb7401:
This PR is going to need a lot of careful review, so I'm starting it now. There's lots of specific constants involved in setting up these curves, and I was fairly sloppy with copying over stuff from SAGE and setting up a good test suite.


str4d:
Just to set expectations, we will not be merging this PR; `pasta_curves` was extracted from this repository and we are not going to merge it back in, and likewise we don't want to have another curve implementation in the workspace.

With that said, absolutely feel free to continue developing in this PR to figure out what needs to be changed to make `halo2_proofs` work with secp/secq curves! We are meanwhile working on removing the need for those curves to depend on `pasta_curves` (the first step of this is in #697), so hopefully everything converges and then this PR can be closed (and smaller targeted PRs opened for any `halo2_*` changes that are necessary).

PR #694, opened 2022-11-24
Simplify `EvaluationDomain` to only accept field elements

str4d:
We do not use `EvaluationDomain` with group elements, so there is no need to directly use the FFT `Group` abstraction trait here.

### DIFF
--- a/halo2_proofs/Cargo.toml
+++ b/halo2_proofs/Cargo.toml
@@ -70,6 +70,10 @@ getrandom = { version = "0.2", features = ["js"] }
 [features]
 default = ["batch"]
 dev-graph = ["plotters", "tabbycat"]
+test-dev-graph = [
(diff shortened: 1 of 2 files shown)


str4d:
[Rebased on `main`](https://github.com/zcash/halo2/compare/fd70b79c6f99403191493b9ba6f88cd4939903e8..93af730c6ca55502d5f1b8eeda8b0587cd8d9328) to include a CI change that I accidentally pushed to `main` (but is necessary to fix CI being broken by transitive dependencies).

str4d:
- Book CI failures are unrelated to this PR.
- Lint failures are unrelated, and I believe being fixed in #678.

therealyingtong (approved):

daira (approved):
utACK

PR #713, opened 2022-12-27
Book PDF generation

HollowMan6:
Resolve #692

Install dependencies:
```bash
pip3 install mdbook-pdf-outline
cargo install mdbook-katex mdbook-pdf
# Patched mdbook for fixing broken links in print.html
git clone https://github.com/HollowMan6/mdBook
cd mdBook && cargo install --path . --force
```

Ensure chrome / edge is available in your system, then run `mdbook build`.

Generated PDFs:

[book-without-linebreak.pdf](https://github.com/zcash/halo2/files/10310764/book-without-linebreak.pdf)
[book-with-linebreak.pdf](https://github.com/zcash/halo2/files/10310765/book-with-linebreak.pdf)

Signed-off-by: Hollow Man <hollowman@opensuse.org>

PR #718, opened 2023-01-08
Fix book deploying CI and add pdf uploading

HollowMan6:
Continuation of #713 

Hi! Sorry I didn't notice that there's a CI for book deploying.

This PR should fix book deploying CI. I've also added the pdf uploading functionality, so that readers now can download the pdf books directly from the GitHub Actions Artifacts.

Signed-off-by: Hollow Man <hollowman@opensuse.org>

### DIFF
--- a/.github/workflows/book.yml
+++ b/.github/workflows/book.yml
@@ -15,16 +15,27 @@ jobs:
           toolchain: nightly
           override: true
 
-      - name: Setup mdBook
-        uses: peaceiris/actions-mdbook@v1
-        with:
-          mdbook-version: 'latest'
+      # - name: Setup mdBook
+      #   uses: peaceiris/actions-mdbook@v1
+      #   with:
+      #     mdbook-version: 'latest'
 
-      - name: Install mdbook-katex
+      # mdBook for mdbook-pdf compatibility by fixing broken links in print.html
+      # Waiting for https://github.com/rust-lang/mdBook/pull/1738 to get merged
+      - name: Clone patched mdbook source code
+        run: cd .. && git clone https://github.com/HollowMan6/mdBook
+
+      - name: Install patched mdbook
+        run: cd ../mdBook && cargo install --path . --force
+
+      - name: Install mdbook-katex and mdbook-pdf
         uses: actions-rs/cargo@v1
         with:
           command: install
-          args: mdbook-katex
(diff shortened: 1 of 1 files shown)


HollowMan6:
https://github.com/HollowMan6/halo2/actions/runs/3868639867

.github/workflows/book.yml
```diff
@@ -15,16 +15,27 @@ jobs:
...
-          mdbook-version: 'latest'
+      # - name: Setup mdBook
+      #   uses: peaceiris/actions-mdbook@v1
+      #   with:
+      #     mdbook-version: 'latest'
+
+      # mdBook for mdbook-pdf compatibility by fixing broken links in print.html
+      # Waiting for https://github.com/rust-lang/mdBook/pull/1738 to get merged
+      - name: Clone patched mdbook source code
+        run: cd .. && git clone https://github.com/HollowMan6/mdBook
```
str4d, line 26: This is unsafe for us, as it enables you to subsequently alter your `master` branch to run whatever you want on our CI 🙂

I'd be fine with this if instead it pins a specific revision corresponding to the mdBook PR.

.github/workflows/book.yml
```diff
@@ -15,16 +15,27 @@ jobs:
...
+
+      - name: Install patched mdbook
+        run: cd ../mdBook && cargo install --path . --force
```
str4d, line 29: And in fact the manual clone is unnecessary. Instead, you can install directly from a git path:
```suggestion
        run: cargo install --git https://github.com/HollowMan6/mdBook.git --rev 62e01b34c23b957579c04ee1b24b57814ed8a4d5
```

str4d (changes requested):

daira (approved):
utACK with @str4d's suggestion.

.github/workflows/book.yml
HollowMan6: OK. I didn't mean to do something bad here... But what you mentioned is reasonable.

.github/workflows/book.yml
HollowMan6: OKay!

str4d (approved):
utACK. I glanced through https://github.com/rust-lang/mdBook/pull/1738 after confirming it contains the commit pinned here, and didn't see anything unusual.

PR #721, opened 2023-01-17
Update the status of support for nested web workers in Safari

daira:


str4d (approved):

daira:
According to the [Safari 16.4 release notes](https://developer.apple.com/documentation/safari-release-notes/safari-16_4-release-notes#Web-API), support for loading scripts for, and termination of nested web workers was added in 16.4, so it is likely that version is needed for full support. It hasn't been tested, though.

PR #723, opened 2023-01-19
bug fix - wasm pack compile error

ImmanuelSegol:
wasm-pack throws an error during compile time when unable to infer the type.

image

### DIFF
--- a/halo2_gadgets/src/poseidon/primitives/mds.rs
+++ b/halo2_gadgets/src/poseidon/primitives/mds.rs
@@ -81,8 +81,13 @@ pub(super) fn generate_mds<F: FromUniformBytes<64> + Ord, const T: usize>(
             if m == j {
                 acc
             } else {
+                // We hard-code the type, to avoid spurious "cannot infer type" rustc errors.
+                let denominator: F = x_j - x_m;
+
                 // We can invert freely; by construction, the elements of xs are distinct.
-                acc * (x - x_m) * (x_j - x_m).invert().unwrap()
+                let denominator_inverted: F = denominator.invert().unwrap();
+
+                acc * (x - x_m) * denominator_inverted
             }
         })
     };


halo2_gadgets/src/poseidon/primitives/mds.rs
```diff
@@ -82,7 +82,12 @@ pub(super) fn generate_mds<F: FromUniformBytes<64> + Ord, const T: usize>(
                 acc
             } else {
                 // We can invert freely; by construction, the elements of xs are distinct.
-                acc * (x - x_m) * (x_j - x_m).invert().unwrap()
+                
+                // We hard-code the type, to avoid wasm-pack compiler error of: cannot infer type. 
+                let denominator: F = x_j - x_m;
+                denominator.invert().unwrap();
+                
+                acc * (x - x_m) * denominator
```
str4d, line 90: This fails tests because [`Field::invert`](https://docs.rs/ff/0.13.0/ff/trait.Field.html#tymethod.invert) is not a mutating method, so the result of `denominator.invert()` is being dropped. Instead, you need to use the output of `denominator.invert()` in the original equation:
```suggestion
                // We hard-code the type, to avoid wasm-pack compiler error of: cannot infer type. 
                let denominator: F = x_j - x_m;
                acc * (x - x_m) * denominator.invert().unwrap()
```

str4d (changes requested):
Thanks for the PR!

halo2_gadgets/src/poseidon/primitives/mds.rs
ImmanuelSegol: Thx will fix asap

str4d (approved):
utACK

halo2_gadgets/src/poseidon/primitives/mds.rs
```diff
@@ -82,7 +82,12 @@ pub(super) fn generate_mds<F: FromUniformBytes<64> + Ord, const T: usize>(
                 acc
             } else {
                 // We can invert freely; by construction, the elements of xs are distinct.
-                acc * (x - x_m) * (x_j - x_m).invert().unwrap()
+                
+                // We hard-code the type, to avoid wasm-pack compiler error of: cannot infer type. 
+                let denominator: F = x_j - x_m;
+                let denominator_inverted: F = denominator.invert().unwrap();
+                
+                acc * (x - x_m) * denominator_inverted
```
str4d, line 90:
```suggestion
                // We hard-code the type, to avoid wasm-pack compiler error of: cannot infer type. 
                let denominator: F = x_j - x_m;

                // We can invert freely; by construction, the elements of xs are distinct.
                let denominator_inverted: F = denominator.invert().unwrap();

                acc * (x - x_m) * denominator_inverted
```

halo2_gadgets/src/poseidon/primitives/mds.rs
```diff
@@ -82,7 +82,12 @@ pub(super) fn generate_mds<F: FromUniformBytes<64> + Ord, const T: usize>(
                 acc
             } else {
                 // We can invert freely; by construction, the elements of xs are distinct.
```
str4d, line 84:
```suggestion
```

str4d (approved):
CI needed some rustfmt changes.

halo2_gadgets/src/poseidon/primitives/mds.rs
```diff
@@ -81,8 +81,13 @@ pub(super) fn generate_mds<F: FromUniformBytes<64> + Ord, const T: usize>(
             if m == j {
                 acc
             } else {
+                // We hard-code the type, to avoid wasm-pack compiler error of: cannot infer type.
```
str4d, line 84:
```suggestion
                // We hard-code the type, to avoid spurious "cannot infer type" rustc errors.
```

PR #728, opened 2023-01-29
  Add feature "multicore"  and you can disable rayon by disabling the  "multicore" feature. 

### LINKED ISSUE #648: Ability to disable rayon
When compiling for wasm, we have to rely on [wasm-bindgen-rayon](https://github.com/GoogleChromeLabs/wasm-bindgen-rayon#building-rust-code) which requires a Rust nightly.
We would like to stay on the Rust stable.
To that end, would it be possible to have a single-threaded mode in halo2 with rayon disabled?
Seems like the change is trivial and was already attempted earlier https://github.com/adria0/halo2/commit/e39100bb73d8122dd16fad191f4923f8e43f9b5a

If this single-threaded mode is ACKed, I could work on a PR.

nagatoism:
Close #648 All test has been passed with/without rayon.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,8 +14,8 @@ minor version bump.
 `halo2` currently uses [rayon](https://github.com/rayon-rs/rayon) for parallel computation.
 The `RAYON_NUM_THREADS` environment variable can be used to set the number of threads.
 
-You can disable [rayon] by disabling the ["multicore"] feature.
-Warning! Halo2 will lose access to parallelism if you disable the ["multicore"] feature.
+You can disable `rayon` by disabling the `"multicore"` feature.
+Warning! Halo2 will lose access to parallelism if you disable the `"multicore"` feature.
 This will significantly degrade performance.
 
 ## License
--- a/README.md
+++ b/README.md
@@ -14,6 +14,10 @@ minor version bump.
 `halo2` currently uses [rayon](https://github.com/rayon-rs/rayon) for parallel computation.
 The `RAYON_NUM_THREADS` environment variable can be used to set the number of threads.
 
+You can disable [rayon] by disabling the ["multicore"] feature.
+Warning! Halo2 will lose access to parallelism if you disable the ["multicore"] feature.
+This will significantly degrade performance.
+
 ## License
 
 Licensed under either of
--- a/halo2_proofs/Cargo.toml
+++ b/halo2_proofs/Cargo.toml
@@ -45,13 +45,13 @@ harness = false
 
 [dependencies]
 backtrace = { version = "0.3", optional = true }
-rayon = "1.5.1"
 ff = "0.13"
 group = "0.13"
 pasta_curves = "0.5"
 rand_core = { version = "0.6", default-features = false }
 tracing = "0.1"
 blake2b_simd = "1"
+maybe-rayon = {version = "0.1.0", default-features = false}
 
 # Developer tooling dependencies
 plotters = { version = "0.3.0", default-features = false, optional = true }
@@ -68,7 +68,8 @@ rand_core = { version = "0.6", default-features = false, features = ["getrandom"
 getrandom = { version = "0.2", features = ["js"] }
 
 [features]
-default = ["batch"]
+default = ["batch", "multicore"]
+multicore = ["maybe-rayon/threads"]
 dev-graph = ["plotters", "tabbycat"]
 test-dev-graph = [
     "dev-graph",
--- a/halo2_proofs/README.md
+++ b/halo2_proofs/README.md
@@ -15,6 +15,10 @@ minor version bump.
 computation. The `RAYON_NUM_THREADS` environment variable can be used to set the number of
 threads.
 
+You can disable [rayon] by disabling the ["multicore"] feature.
+Warning! Halo2 will lose access to parallelism if you disable the ["multicore"] feature.
+This will significantly degrade performance.
+
 ## License
 
 Licensed under either of
--- a/halo2_proofs/README.md
+++ b/halo2_proofs/README.md
@@ -15,8 +15,8 @@ minor version bump.
 computation. The `RAYON_NUM_THREADS` environment variable can be used to set the number of
 threads.
 
-You can disable [rayon] by disabling the ["multicore"] feature.
-Warning! Halo2 will lose access to parallelism if you disable the ["multicore"] feature.
+You can disable `rayon` by disabling the `"multicore"` feature.
+Warning! Halo2 will lose access to parallelism if you disable the `"multicore"` feature.
 This will significantly degrade performance.
 
 ## License
--- a/halo2_proofs/src/multicore.rs
+++ b/halo2_proofs/src/multicore.rs
@@ -1,9 +1,61 @@
 //! An interface for dealing with the kinds of parallel computations involved in
-//! `halo2`. It's currently just a (very!) thin wrapper around [`rayon`] but may
+//! `halo2`. It's currently just a (very!) thin wrapper around `rayon` but may
 //! be extended in the future to allow for various parallelism strategies.
 
-pub use rayon::{
-    current_num_threads,
-    iter::{IndexedParallelIterator, IntoParallelIterator, ParallelIterator},
+pub use maybe_rayon::{
+    iter::{IntoParallelIterator, ParallelIterator},
     join, scope, Scope,
 };
+
+#[cfg(feature = "multicore")]
+pub use maybe_rayon::{current_num_threads, iter::IndexedParallelIterator};
+
+#[cfg(not(feature = "multicore"))]
+pub fn current_num_threads() -> usize {
+    1
+}
+
+#[cfg(not(feature = "multicore"))]
+pub trait IndexedParallelIterator: std::iter::Iterator {}
+
+pub trait TryFoldAndReduce<T, E> {
+    /// Implements `iter.try_fold().try_reduce()` for `rayon::iter::ParallelIterator`,
+    /// falling back on `Iterator::try_fold` when the `multicore` feature flag is
+    /// disabled.
+    fn try_fold_and_reduce(
+        self,
+        identity: impl Fn() -> T + Send + Sync,
+        fold_op: impl Fn(Result<T, E>, Result<T, E>) -> Result<T, E> + Send + Sync,
+    ) -> Result<T, E>;
+}
+
+#[cfg(feature = "multicore")]
+impl<T, E, I> TryFoldAndReduce<T, E> for I
+where
+    T: Send + Sync,
+    E: Send + Sync,
+    I: maybe_rayon::iter::ParallelIterator<Item = Result<T, E>>,
+{
+    fn try_fold_and_reduce(
+        self,
+        identity: impl Fn() -> T + Send + Sync,
+        fold_op: impl Fn(Result<T, E>, Result<T, E>) -> Result<T, E> + Send + Sync,
+    ) -> Result<T, E> {
+        self.try_fold(&identity, |msm, res| fold_op(Ok(msm), res))
+            .try_reduce(&identity, |a, b| fold_op(Ok(a), Ok(b)))
+    }
+}
+
+#[cfg(not(feature = "multicore"))]
+impl<T, E, I> TryFoldAndReduce<T, E> for I
+where
+    I: std::iter::Iterator<Item = Result<T, E>>,
+{
+    fn try_fold_and_reduce(
+        mut self,
+        identity: impl Fn() -> T + Send + Sync,
+        fold_op: impl Fn(Result<T, E>, Result<T, E>) -> Result<T, E> + Send + Sync,
+    ) -> Result<T, E> {
+        self.try_fold(identity(), |acc, res| fold_op(Ok(acc), res))
+    }
+}
--- a/halo2_proofs/src/multicore.rs
+++ b/halo2_proofs/src/multicore.rs
@@ -22,9 +22,9 @@ pub trait TryFoldAndReduce<T, E> {
     /// Implements `iter.try_fold().try_reduce()` for `rayon::iter::ParallelIterator`,
     /// falling back on `Iterator::try_fold` when the `multicore` feature flag is
     /// disabled.
-    /// try_fold_and_reduce function can only be called by a iter with
-    /// Result<T,E> item type because the folder_op Fn<T,T>->T must
-    /// meet both trait bound of rayon's try_fold and try_reduce```   
+    /// The `try_fold_and_reduce` function can only be called by a iter with
+    /// `Result<T, E>` item type because the `fold_op` must meet the trait
+    /// bounds of both `try_fold` and `try_reduce` from rayon.   
     fn try_fold_and_reduce(
         self,
         identity: impl Fn() -> T + Send + Sync,
--- a/halo2_proofs/src/multicore.rs
+++ b/halo2_proofs/src/multicore.rs
@@ -22,6 +22,9 @@ pub trait TryFoldAndReduce<T, E> {
     /// Implements `iter.try_fold().try_reduce()` for `rayon::iter::ParallelIterator`,
     /// falling back on `Iterator::try_fold` when the `multicore` feature flag is
     /// disabled.
+    /// try_fold_and_reduce function can only be called by a iter with
+    /// Result<T,E> item type because the folder_op Fn<T,T>->T must
+    /// meet both trait bound of rayon's try_fold and try_reduce```   
     fn try_fold_and_reduce(
         self,
         identity: impl Fn() -> T + Send + Sync,
--- a/halo2_proofs/src/multicore.rs
+++ b/halo2_proofs/src/multicore.rs
@@ -2,6 +2,15 @@
 //! `halo2`. It's currently just a (very!) thin wrapper around `rayon` but may
 //! be extended in the future to allow for various parallelism strategies.
 
+#[cfg(all(
+    feature = "multicore",
+    target_arch = "wasm32",
+    not(target_feature = "atomics")
+))]
+compile_error!(
+    "The multicore feature flag is not supported on wasm32 architectures without atomics"
+);
+
 pub use maybe_rayon::{
     iter::{IntoParallelIterator, ParallelIterator},
     join, scope, Scope,
--- a/halo2_proofs/src/multicore.rs
+++ b/halo2_proofs/src/multicore.rs
@@ -28,7 +28,7 @@ pub trait TryFoldAndReduce<T, E> {
     fn try_fold_and_reduce(
         self,
         identity: impl Fn() -> T + Send + Sync,
-        fold_op: impl Fn(Result<T, E>, Result<T, E>) -> Result<T, E> + Send + Sync,
+        fold_op: impl Fn(T, Result<T, E>) -> Result<T, E> + Send + Sync,
     ) -> Result<T, E>;
 }
 
@@ -42,10 +42,10 @@ where
     fn try_fold_and_reduce(
         self,
         identity: impl Fn() -> T + Send + Sync,
-        fold_op: impl Fn(Result<T, E>, Result<T, E>) -> Result<T, E> + Send + Sync,
+        fold_op: impl Fn(T, Result<T, E>) -> Result<T, E> + Send + Sync,
     ) -> Result<T, E> {
-        self.try_fold(&identity, |msm, res| fold_op(Ok(msm), res))
-            .try_reduce(&identity, |a, b| fold_op(Ok(a), Ok(b)))
+        self.try_fold(&identity, &fold_op)
+            .try_reduce(&identity, |a, b| fold_op(a, Ok(b)))
     }
 }
 
@@ -57,8 +57,8 @@ where
     fn try_fold_and_reduce(
         mut self,
         identity: impl Fn() -> T + Send + Sync,
-        fold_op: impl Fn(Result<T, E>, Result<T, E>) -> Result<T, E> + Send + Sync,
+        fold_op: impl Fn(T, Result<T, E>) -> Result<T, E> + Send + Sync,
     ) -> Result<T, E> {
-        self.try_fold(identity(), |acc, res| fold_op(Ok(acc), res))
+        self.try_fold(identity(), fold_op)
     }
 }
--- a/halo2_proofs/src/plonk/verifier/batch.rs
+++ b/halo2_proofs/src/plonk/verifier/batch.rs
@@ -4,11 +4,14 @@ use rand_core::OsRng;
 
 use super::{verify_proof, VerificationStrategy};
 use crate::{
-    multicore::{IndexedParallelIterator, IntoParallelIterator, ParallelIterator},
+    multicore::{IntoParallelIterator, TryFoldAndReduce},
     plonk::{Error, VerifyingKey},
     poly::commitment::{Guard, Params, MSM},
     transcript::{Blake2bRead, EncodedChallenge},
 };
+
+#[cfg(feature = "multicore")]
+use crate::multicore::{IndexedParallelIterator, ParallelIterator};
 
 /// A proof verification strategy that returns the proof's MSM.
 ///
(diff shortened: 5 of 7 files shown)


CPerezz (commented):
Looks good.

Only two comments:
1) I wonder if `cfg_if!` could abbreviate some of the repeated chunks of code here. Maybe worth looking at it.
2) I think this doesn't need to go through the `nightly` feature first. As it doesn't include anything new. But worth checking with @str4d .

For the rest, looks good!

README.md
```diff
@@ -14,6 +14,9 @@ minor version bump.
 `halo2` currently uses [rayon](https://github.com/rayon-rs/rayon) for parallel computation.
 The `RAYON_NUM_THREADS` environment variable can be used to set the number of threads.
 
+You can disable [rayon] by disabling the ["multicore"] feature.
+Warning! Halo2 will lose access to parallelism if you disable the ["multicore"] feature. This will significantly downgrade performance.
```
daira, line 18:
```suggestion
Warning! Halo2 will lose access to parallelism if you disable the ["multicore"] feature.
This will significantly degrade performance.
```

halo2_proofs/Cargo.toml
```diff
@@ -45,7 +45,7 @@ harness = false
 
 [dependencies]
 backtrace = { version = "0.3", optional = true }
-rayon = "1.5.1"
+rayon = {version = "1.5.1",optional = true}
```
daira, line 48:
```suggestion
rayon = {version = "1.5.1", optional = true}
```

halo2_proofs/Cargo.toml
```diff
@@ -68,7 +68,8 @@ rand_core = { version = "0.6", default-features = false, features = ["getrandom"
 getrandom = { version = "0.2", features = ["js"] }
 
 [features]
-default = ["batch"]
+default = ["batch","multicore"]
```
daira, line 71:
```suggestion
default = ["batch", "multicore"]
```

halo2_proofs/README.md
```diff
@@ -15,6 +15,9 @@ minor version bump.
 computation. The `RAYON_NUM_THREADS` environment variable can be used to set the number of
 threads.
 
+You can disable [rayon] by disabling the ["multicore"] feature.
+Warning! Halo2 will lose access to parallelism if you disable the ["multicore"] feature. This will significantly downgrade the performance.
```
daira, line 19:
```suggestion
Warning! Halo2 will lose access to parallelism if you disable the ["multicore"] feature.
This will significantly degrade the performance. 
```

halo2_proofs/src/arithmetic.rs
```diff
@@ -152,19 +152,36 @@ pub fn best_multiexp<C: CurveAffine>(coeffs: &[C::Scalar], bases: &[C]) -> C::Cu
...
                 .zip(bases.chunks(chunk))
                 .zip(results.iter_mut())
             {
-                scope.spawn(move |_| {
-                    multiexp_serial(coeffs, bases, acc);
-                });
+                multiexp_serial(coeffs, bases, acc);
             }
-        });
+        };
```
daira, line 183:
```suggestion
        }
```

halo2_proofs/src/arithmetic.rs
```diff
@@ -262,10 +279,20 @@ pub fn recursive_butterfly_arithmetic<Scalar: Field, G: FftGroup<Scalar>>(
...
-        );
+
+        #[cfg(feature = "multicore")]
+        {
+            multicore::join(
+                || recursive_butterfly_arithmetic(left, n / 2, twiddle_chunk * 2, twiddles),
+                || recursive_butterfly_arithmetic(right, n / 2, twiddle_chunk * 2, twiddles),
+            );
+        }
+
```
daira, line 290:
```suggestion
```

halo2_proofs/src/arithmetic.rs
```diff
@@ -337,22 +364,30 @@ where
...
+        multicore::scope(|scope| {
+            for (chunk_num, v) in v.chunks_mut(chunk).enumerate() {
+                let f = f.clone();
+                scope.spawn(move |_| {
+                    let start = chunk_num * chunk;
+                    f(v, start);
+                });
+            }
+        });
     }
```
daira, line 386:
```suggestion
```

halo2_proofs/src/multicore.rs
```diff
@@ -1,9 +1,15 @@
...
 //! be extended in the future to allow for various parallelism strategies.
-
+#[cfg(feature = "multicore")]
 pub use rayon::{
     current_num_threads,
     iter::{IndexedParallelIterator, IntoParallelIterator, ParallelIterator},
     join, scope, Scope,
 };
+
+
```
daira, line 11:
```suggestion
```

halo2_proofs/src/multicore.rs
```diff
@@ -1,9 +1,15 @@
...
+#[cfg(not(feature = "multicore"))]
+pub fn current_num_threads() -> usize {
+    return 1;
```
daira, line 14:
```suggestion
    1
```

halo2_proofs/src/plonk/verifier/batch.rs
```diff
@@ -89,34 +92,70 @@ where
...
-                let instances: Vec<_> = instances.iter().map(|i| &i[..]).collect();
-
-                let strategy = BatchStrategy::new(params);
-                let mut transcript = Blake2bRead::init(&item.proof[..]);
-                verify_proof(params, vk, strategy, &instances, &mut transcript).map_err(|e| {
-                    tracing::debug!("Batch item {} failed verification: {}", i, e);
-                    e
+        #[cfg(feature = "multicore")]
+        {
+            let final_msm = self
```
daira, line 97:
```suggestion
        let final_msm = {
            self
```

halo2_proofs/src/plonk/verifier/batch.rs
```diff
@@ -89,34 +92,70 @@ where
...
-            Err(_) => false,
+                .try_fold(
+                    || params.empty_msm(),
+                    |msm, res| res.map(|proof_msm| accumulate_msm(msm, proof_msm)),
+                )
+                .try_reduce(|| params.empty_msm(), |a, b| Ok(accumulate_msm(a, b)));
+            match final_msm {
+                Ok(msm) => msm.eval(),
+                Err(_) => false,
+            }
```
daira, line 124:
```suggestion
```
(moved to common code below)

halo2_proofs/src/plonk/verifier/batch.rs
```diff
@@ -89,34 +92,70 @@ where
...
-            .try_reduce(|| params.empty_msm(), |a, b| Ok(accumulate_msm(a, b)));
-
-        match final_msm {
-            Ok(msm) => msm.eval(),
-            Err(_) => false,
+                .try_fold(
+                    || params.empty_msm(),
+                    |msm, res| res.map(|proof_msm| accumulate_msm(msm, proof_msm)),
+                )
+                .try_reduce(|| params.empty_msm(), |a, b| Ok(accumulate_msm(a, b)));
```
daira, line 120:
```suggestion
                .try_reduce(|| params.empty_msm(), |a, b| Ok(accumulate_msm(a, b)))
```

halo2_proofs/src/plonk/verifier/batch.rs
```diff
@@ -89,34 +92,70 @@ where
...
+            match final_msm {
+                Ok(msm) => msm.eval(),
+                Err(_) => false,
+            }
+        }
+        #[cfg(not(feature = "multicore"))]
+        {
+            let final_msm = self
```
daira, line 128:
```suggestion
        let final_msm = {
            self
```

halo2_proofs/src/plonk/verifier/batch.rs
```diff
@@ -89,34 +92,70 @@ where
...
+                    })
+                })
+                .try_fold(params.empty_msm(), |msm, res| {
+                    res.map(|proof_msm| accumulate_msm(msm, proof_msm))
+                });
+
+            let final_msm = match final_msm {
+                Ok(final_msm) => Ok(accumulate_msm(params.empty_msm(), final_msm)),
+                Err(err) => Err(err),
+            };
```
daira, line 154:
```suggestion
                })
                .map(|msm| accumulate_msm(params.empty_msm(), msm))
        }
```

halo2_proofs/src/plonk/verifier/batch.rs
```diff
@@ -89,34 +92,70 @@ where
...
+            match final_msm {
+                Ok(msm) => msm.eval(),
+                Err(_) => false,
+            }
         }
```
daira, line 159:
```suggestion
        match final_msm {
            Ok(msm) => msm.eval(),
            Err(_) => false,
        }
```

halo2_proofs/src/poly/evaluator.rs
```diff
@@ -210,20 +210,38 @@ impl<E, F: Field, B: Basis> Evaluator<E, F, B> {
...
+                            chunk_size,
+                            chunk_index,
+                            polys: &self.polys,
+                        };
+                        out.copy_from_slice(&recurse(ast, &ctx));
+                    });
+                }
+            });
+        }
+
```
daira, line 230:
```suggestion
```

daira (approved):
utACK with suggested changes.

str4d:
There is a lot of duplicated logic here, and the conditionals are spread out over the codebase. I'd prefer we take a similar approach to `bellman` and provide fake versions of the `rayon` types in the `halo2_proofs::multicore` module that just implement serial logic. Maybe we should move this "maybe rayon" logic into a crate?

And now that I go searching, someone else had the same idea a year ago: https://crates.io/crates/maybe-rayon

halo2_proofs/Cargo.toml
str4d, line 48:
```suggestion
rayon = { package = "maybe-rayon", version = "0.1", default-features = false }
```

halo2_proofs/Cargo.toml
```diff
@@ -68,7 +68,8 @@ rand_core = { version = "0.6", default-features = false, features = ["getrandom"
...
+multicore = ["rayon"]
```
str4d, line 72:
```suggestion
multicore = ["rayon/threads"]
```
(or `maybe-rayon/threads`, IDK if namespaced feature flags are resolved relative to the package name or the dependency name.)

halo2_proofs/src/arithmetic.rs
```diff
@@ -152,19 +152,36 @@ pub fn best_multiexp<C: CurveAffine>(coeffs: &[C::Scalar], bases: &[C]) -> C::Cu
         let chunk = coeffs.len() / num_threads;
         let num_chunks = coeffs.chunks(chunk).len();
         let mut results = vec![C::Curve::identity(); num_chunks];
-        multicore::scope(|scope| {
+
+        #[cfg(feature = "multicore")]
```
str4d, line 156: Undo these changes.

halo2_proofs/src/arithmetic.rs
```diff
@@ -262,10 +279,20 @@ pub fn recursive_butterfly_arithmetic<Scalar: Field, G: FftGroup<Scalar>>(
         a[1] -= &t;
     } else {
         let (left, right) = a.split_at_mut(n / 2);
-        multicore::join(
-            || recursive_butterfly_arithmetic(left, n / 2, twiddle_chunk * 2, twiddles),
-            || recursive_butterfly_arithmetic(right, n / 2, twiddle_chunk * 2, twiddles),
-        );
+
+        #[cfg(feature = "multicore")]
```
str4d, line 283: Undo these changes.

halo2_proofs/src/arithmetic.rs
```diff
@@ -337,22 +364,30 @@ where
 /// This simple utility function will parallelize an operation that is to be
 /// performed over a mutable slice.
 pub fn parallelize<T: Send, F: Fn(&mut [T], usize) + Send + Sync + Clone>(v: &mut [T], f: F) {
-    let n = v.len();
-    let num_threads = multicore::current_num_threads();
-    let mut chunk = n / num_threads;
-    if chunk < num_threads {
-        chunk = n;
+    #[cfg(feature = "multicore")]
```
str4d, line 367: Undo these changes.

halo2_proofs/src/multicore.rs
```diff
@@ -1,9 +1,15 @@
 //! An interface for dealing with the kinds of parallel computations involved in
 //! `halo2`. It's currently just a (very!) thin wrapper around [`rayon`] but may
 //! be extended in the future to allow for various parallelism strategies.
-
+#[cfg(feature = "multicore")]
```
str4d, line 4: Using `maybe-rayon`, I think this feature flag could just be removed and these imports left as unconditional (though we would need to keep `current_num_threads` separately conditional as I don't think `maybe-rayon` provides an equivalent).

halo2_proofs/src/plonk/verifier/batch.rs
```diff
@@ -89,34 +92,70 @@ where
...
-                    .map(|i| i.iter().map(|c| &c[..]).collect())
-                    .collect();
-                let instances: Vec<_> = instances.iter().map(|i| &i[..]).collect();
-
-                let strategy = BatchStrategy::new(params);
-                let mut transcript = Blake2bRead::init(&item.proof[..]);
-                verify_proof(params, vk, strategy, &instances, &mut transcript).map_err(|e| {
-                    tracing::debug!("Batch item {} failed verification: {}", i, e);
-                    e
+        #[cfg(feature = "multicore")]
```
str4d, line 95: Undo these changes.

halo2_proofs/src/poly/evaluator.rs
```diff
@@ -210,20 +210,38 @@ impl<E, F: Field, B: Basis> Evaluator<E, F, B> {
         // Apply `ast` to each chunk in parallel, writing the result into an output
         // polynomial.
         let mut result = B::empty_poly(domain);
-        multicore::scope(|scope| {
+        #[cfg(feature = "multicore")]
```
str4d, line 213: Undo these changes.

str4d (changes requested):
If we switch to using the https://crates.io/crates/maybe-rayon crate, I think that the only code changes in this PR should be to `halo2_proofs/src/multicore.rs`, which would greatly simplify it and make future maintenance simpler.

halo2_proofs/src/poly/evaluator.rs
nagatoism: Thanks for fixing the typos.  Will apply.

halo2_proofs/src/multicore.rs
```diff
@@ -2,8 +2,15 @@
 //! `halo2`. It's currently just a (very!) thin wrapper around [`rayon`] but may
```
daira, line 2:
```suggestion
//! `halo2`. It's currently just a (very!) thin wrapper around `rayon` but may
```
(This link broke because `rayon` is no longer a direct dependency.)

halo2_proofs/src/multicore.rs
```diff
@@ -2,8 +2,15 @@
...
 //! be extended in the future to allow for various parallelism strategies.
 
-pub use rayon::{
-    current_num_threads,
+pub use maybe_rayon::{
     iter::{IndexedParallelIterator, IntoParallelIterator, ParallelIterator},
```
daira, line 6: Please investigate the wasm32-wasi failure; I believe this was working on that platform before.

daira (changes requested):
Please investigate the wasm32-wasi failures. Otherwise this is looking pretty good modulo str4d's comments.

nagatoism:
Hi Found that IndexedParallelIter is not supported by maybe-rayon. The new code fixes this. 

The code is necessary because try_fold rayon and try_fold in rust std are different functions so I cannot reuse code. 
Here  https://docs.rs/rayon/latest/rayon/iter/trait.ParallelIterator.html#method.fold

PS pls help with the mut  on final_msm. No idea why the mut is needed.

str4d:
If `IndexedParallelIter` isn't supported upstream, then can we just write our own implementation? That would avoid the try_fold issue.

str4d:
CI is failing for an unrelated reason: `constant_time_eq` has raised its MSRV to 1.59. It is a dependency of `blake2b_simd`, so we will therefore need to similarly raise our MSRV (I'll do this in a separate PR).

README.md
```diff
@@ -14,6 +14,10 @@ minor version bump.
 `halo2` currently uses [rayon](https://github.com/rayon-rs/rayon) for parallel computation.
 The `RAYON_NUM_THREADS` environment variable can be used to set the number of threads.
 
+You can disable [rayon] by disabling the ["multicore"] feature.
+Warning! Halo2 will lose access to parallelism if you disable the ["multicore"] feature.
```
daira, line 18:
```suggestion
You can disable `rayon` by disabling the `"multicore"` feature.
Warning! Halo2 will lose access to parallelism if you disable the `"multicore"` feature.
```

halo2_proofs/README.md
```diff
@@ -15,6 +15,10 @@ minor version bump.
 computation. The `RAYON_NUM_THREADS` environment variable can be used to set the number of
 threads.
 
+You can disable [rayon] by disabling the ["multicore"] feature.
+Warning! Halo2 will lose access to parallelism if you disable the ["multicore"] feature.
```
daira, line 19:
```suggestion
You can disable `rayon` by disabling the `"multicore"` feature.
Warning! Halo2 will lose access to parallelism if you disable the `"multicore"` feature.
```

halo2_proofs/src/multicore.rs
```diff
@@ -1,9 +1,61 @@
...
+#[cfg(not(feature = "multicore"))]
+pub trait IndexedParallelIterator: std::iter::Iterator {}
+
+pub trait TryFoldAndReduce<T, E> {
+    /// Implements `iter.try_fold().try_reduce()` for `rayon::iter::ParallelIterator`,
+    /// falling back on `Iterator::try_fold` when the `multicore` feature flag is
+    /// disabled.
+    /// try_fold_and_reduce function can only be called by a iter with
+    /// Result<T,E> item type because the folder_op Fn<T,T>->T must
+    /// meet both trait bound of rayon's try_fold and try_reduce```
```
daira, line 27:
```suggestion
    /// The `try_fold_and_reduce` function can only be called by a iter with
    /// `Result<T, E>` item type because the `fold_op: Fn<T, T> -> T` must
    /// meet both the trait bounds of rayon's `try_fold` and `try_reduce`.   
```

halo2_proofs/src/multicore.rs
```diff
@@ -1,9 +1,61 @@
...
+        fold_op: impl Fn(Result<T, E>, Result<T, E>) -> Result<T, E> + Send + Sync,
+    ) -> Result<T, E>;
+}
+
+#[cfg(feature = "multicore")]
+impl<
+        T: Send + Sync,
+        E: Send + Sync,
+        I: maybe_rayon::iter::ParallelIterator<Item = Result<T, E>>,
+    > TryFoldAndReduce<T, E> for I
```
daira, line 40: I usually prefer to use `where` if the generic constraint would break across lines.

halo2_proofs/src/multicore.rs
```diff
@@ -1,9 +1,61 @@
...
+    /// Implements `iter.try_fold().try_reduce()` for `rayon::iter::ParallelIterator`,
+    /// falling back on `Iterator::try_fold` when the `multicore` feature flag is
+    /// disabled.
+    /// try_fold_and_reduce function can only be called by a iter with
+    /// Result<T,E> item type because the folder_op Fn<T,T>->T must
+    /// meet both trait bound of rayon's try_fold and try_reduce```   
+    fn try_fold_and_reduce(
+        self,
+        identity: impl Fn() -> T + Send + Sync,
+        fold_op: impl Fn(Result<T, E>, Result<T, E>) -> Result<T, E> + Send + Sync,
```
daira, line 31:
```suggestion
        fold_op: impl Fn(T, Result<T, E>) -> Result<T, E> + Send + Sync,
```
`fold_op` is always called with `Ok(...)` as its first argument in both the multicore and non-multicore implementations. This makes sense because the fold/reduce is only continued when the accumulator is ok.

halo2_proofs/src/multicore.rs
```diff
@@ -1,9 +1,61 @@
...
+        I: maybe_rayon::iter::ParallelIterator<Item = Result<T, E>>,
+    > TryFoldAndReduce<T, E> for I
+{
+    fn try_fold_and_reduce(
+        self,
+        identity: impl Fn() -> T + Send + Sync,
+        fold_op: impl Fn(Result<T, E>, Result<T, E>) -> Result<T, E> + Send + Sync,
+    ) -> Result<T, E> {
+        self.try_fold(&identity, |msm, res| fold_op(Ok(msm), res))
+            .try_reduce(&identity, |a, b| fold_op(Ok(a), Ok(b)))
```
daira, line 48:
```suggestion
        fold_op: impl Fn(T, Result<T, E>) -> Result<T, E> + Send + Sync,
    ) -> Result<T, E> {
        self.try_fold(&identity, fold_op).try_reduce(&identity, |a, b| fold_op(a, Ok(b)))
```

halo2_proofs/src/multicore.rs
```diff
@@ -1,9 +1,61 @@
...
+}
+
+#[cfg(not(feature = "multicore"))]
+impl<T, E, I: std::iter::Iterator<Item = Result<T, E>>> TryFoldAndReduce<T, E> for I {
+    fn try_fold_and_reduce(
+        mut self,
+        identity: impl Fn() -> T + Send + Sync,
+        fold_op: impl Fn(Result<T, E>, Result<T, E>) -> Result<T, E> + Send + Sync,
+    ) -> Result<T, E> {
+        self.try_fold(identity(), |acc, res| fold_op(Ok(acc), res))
```
daira, line 59:
```suggestion
        fold_op: impl Fn(T, Result<T, E>) -> Result<T, E> + Send + Sync,
    ) -> Result<T, E> {
        self.try_fold(identity(), fold_op)
```

halo2_proofs/src/plonk/verifier/batch.rs
```diff
@@ -108,11 +111,10 @@ where
                     e
                 })
             })
-            .try_fold(
+            .try_fold_and_reduce(
                 || params.empty_msm(),
-                |msm, res| res.map(|proof_msm| accumulate_msm(msm, proof_msm)),
-            )
-            .try_reduce(|| params.empty_msm(), |a, b| Ok(accumulate_msm(a, b)));
+                |msm, res| msm.and_then(|acc| res.map(|proof_msm| accumulate_msm(acc, proof_msm))),
```
daira, line 116:
```suggestion
                |acc, res| res.map(|proof_msm| accumulate_msm(acc, proof_msm)),
```

daira (approved):
utACK with a proposed refactoring.

str4d:
- [Force-pushed](https://github.com/zcash/halo2/compare/5d391f4584fac0c7574253d321f64b45ab26cde2..c9144b347c0beeae954fb9cc9718bbb5ceac73c1) to fix `rustfmt` and referencing errors after accepting a suggestion.
- [Rebased on `main`](https://github.com/zcash/halo2/compare/c9144b347c0beeae954fb9cc9718bbb5ceac73c1..b29306c6dea7bb6e30513d6b18b9124204293f75) to fix MSRV issue.

str4d:
[Force-pushed](https://github.com/zcash/halo2/compare/b29306c6dea7bb6e30513d6b18b9124204293f75..a6eef9e3cb0fb4fd480ff22717ef679f0d8b4100) to address @daira's comment.

str4d:
CI is currently failing because of an unrelated issue. Fixing that in #734, and then I'll rebase this PR so we can see it fail in the expected way :)

str4d:
[Rebased on `main`](https://github.com/zcash/halo2/compare/47e99978f5b100f95d43081df3a6a2a42814c6d3..2c9c105dc398e2662d63750cbc60b5c4ce31c585) to include #734.

str4d:
And after [another force-push to fix my last commit](https://github.com/zcash/halo2/compare/2c9c105dc398e2662d63750cbc60b5c4ce31c585..f95e495951c79e3fe88df7fdbef88ab69da70275), the WASM CI test is failing as expected: in addition to the expected compilation issues (due to `maybe-rayon` falling back to single-threading for non-atomics WASM if `threads` is enabled) we now see our compiler error that `multicore` is not supported on non-atomics WASM.

str4d (approved):
ACK. Thanks for the PR @nagatoism!

PR #730, opened 2023-02-10
Bump MSRV to 1.59

str4d:
`blake2b_simd 1.0.1` raised its MSRV to 1.59.


str4d:
Turns out MSRV of 1.59 is insufficient because `proptest 1.1.0` added a dependency on `unarray`, which has an implicit MSRV of 1.60. So let's just bump to 1.60 (the same MSRV as `zcash_*` crates currently have).

PR #731, opened 2023-02-10
Bump MSRV to 1.60

str4d:
- `blake2b_simd 1.0.1` raised its MSRV to 1.59.
- `proptest 1.1.0` added a dependency on `unarray` which has an implicit MSRV of 1.60.

Closes zcash/halo2#719.


daira (approved):
utACK

PR #734, opened 2023-02-14
Constrain `inferno` transitive dependency to our MSRV

str4d:
`inferno 0.11.5` raised its MSRV to 1.64.0. Adding this range constraint is fine because it's only a dev-dependency.

### DIFF
--- a/halo2_gadgets/Cargo.toml
+++ b/halo2_gadgets/Cargo.toml
@@ -42,6 +42,7 @@ criterion = "0.3"
 proptest = "1.0.0"
 
 [target.'cfg(unix)'.dev-dependencies]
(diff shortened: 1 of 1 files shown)


halo2_gadgets/Cargo.toml
```diff
@@ -42,6 +42,7 @@ criterion = "0.3"
 proptest = "1.0.0"
 
 [target.'cfg(unix)'.dev-dependencies]
+inferno = ">=0.11, <0.11.5" # MSRV 1.59
```
daira, line 45:
```suggestion
# we depend on inferno only via pprof
inferno = ">=0.11, <0.11.5" # MSRV 1.59
```

daira (approved):
utACK with suggested comment, non-blocking

PR #739, opened 2023-02-25
Fix a nondeterminism bug: we were depending on sort order

daira:


halo2_proofs/src/circuit/floor_planner/v1/strategy.rs
```diff
@@ -194,12 +194,31 @@ fn slot_in(
...
+where
+    F: FnMut(&T) -> K,
+    K: Ord,
+{
+    // https://github.com/rust-lang/rust/blob/1.56.1/library/core/src/slice/mod.rs#L2365-L2402
+    #[cfg(feature = "floor-planner-v1-legacy-pdqsort")]
+    halo2_legacy_pdqsort::sort::quicksort(v, |a, b| f(a).lt(&f(b)));
+
+    #[cfg(not(feature = "floor-planner-v1-legacy-pdqsort"))]
+    v.sort_by_key(f);
```
daira, line 210: This is probably not what we want. It might be necessary to put the conditional code in `slot_in_biggest_advice_first` rather than factoring it out, even if that results in a bit uglier code.

Cargo.toml
```diff
@@ -4,3 +4,6 @@ members = [
     "halo2_gadgets",
     "halo2_proofs",
 ]
+
+[patch.crates-io]
+pasta_curves = { path = "../pasta_curves" }
```
daira, line 9: I'll remove this from the PR.

str4d:
I did some tests. `sort_by_key` is slower than `sort_unstable_by_key` for the current non-unique keys (on the order of 80-100µs vs 105µs on my Ryzen 9 59850X). But once I change the sort key to `(advice_area, region_index)` (to tie-break on region index, thus making sort keys unique), `sort_by_key` becomes faster than `sort_unstable_by_key` (on the order of 150-165µs vs 140µs on my Ryzen 9 59850X). That makes sense to me, as `sort_unstable_by_key` leverages non-uniqueness of keys for speed.

I also tested `sort_by_cached_key`, and it is much faster in both cases (non-unique keys: 21-24µs, unique keys: 23-37µs).

Given that we are allocating anyway, using `sort_by_cached_key` with its temporary auxiliary allocation is the best option. So we should just use that with the existing non-unique keys (and rely on the original `region_shapes` order being sorted determinstically already).

halo2_proofs/src/circuit/floor_planner/v1/strategy.rs
```diff
@@ -194,12 +194,31 @@ fn slot_in(
...
+}
+
 /// Sorts the regions by advice area and then lays them out with the [`slot_in`] strategy.
 pub fn slot_in_biggest_advice_first(
     region_shapes: Vec<RegionShape>,
 ) -> (Vec<RegionStart>, CircuitAllocations) {
     let mut sorted_regions: Vec<_> = region_shapes.into_iter().collect();
```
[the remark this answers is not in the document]
str4d: Yep, it ends up being faster.

str4d:
[Force-pushed](https://github.com/zcash/halo2/compare/5e2aaa37f85338c953673cc1b594aaa6005d73cc..262cf782cc39befd1c141113fa071adb11c205a6) to address @daira's comments.

str4d:
[Rebased on `main`](https://github.com/zcash/halo2/compare/262cf782cc39befd1c141113fa071adb11c205a6..03d191ff44c065e79ec663b0e504fdf1756718af) to fix merge conflicts.

str4d:
[Force-pushed](https://github.com/zcash/halo2/compare/03d191ff44c065e79ec663b0e504fdf1756718af..5a93ca75439e59bab2d504663fde68e5e8436e3e) to adjust the dependency list and update the changelog.

daira (commented):
utACK. Needs 3 ACKs since it affects consensus.

daira:
Do we want to merge https://github.com/zcash/halo2_legacy_pdqsort/pull/3 and push a release 0.1.0 of `halo2_legacy_pdqsort` that has tests?

halo2_proofs/CHANGELOG.md
```diff
@@ -6,6 +6,24 @@ and this project adheres to Rust's notion of
...
+  it is not stable across targets or compiler versions. In particular, an edge
+  case within the sorting algorithm differed between 32-bit and 64-bit targets.
+  This meant that some circuits (like the [Orchard circuit]) would be layed out
+  differently, resulting in incompatible verifying keys. This release makes a
+  **breaking change** to the behaviour of `floor_planner::V1` to instead use a
+  stable sort.
+  - To retain compatibility with the Orchard circuit as deployed in [Zcash NU5],
+    a new `floor-planner-v1-legacy-pdqsort` feature flag has been added. When
+    enabled, `floor_planner::V1` instead pins its behaviour to the version of
+    `slice::sort_unstable_by_key` from Rust 1.56.1 on 64-bit targets.
```
daira, line 21: This is ambiguous; it could be interpreted as saying that the pinning is only done on 64-bit targets, which is not what we mean.
```suggestion
    `slice::sort_unstable_by_key` from Rust 1.56.1, always matching how that
    version behaved on 64-bit targets.
```

str4d:
[Force-pushed](https://github.com/zcash/halo2/compare/5a93ca75439e59bab2d504663fde68e5e8436e3e..8231d8ffab5257dd989bd2d13c66f94f9461d9f0) to address @daira's comment and use the published version of `halo2_legacy_pdqsort`.

nuttycom (approved):
utACK

daira (commented):
utACK

halo2_proofs/CHANGELOG.md
```diff
@@ -6,6 +6,25 @@ and this project adheres to Rust's notion of
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Breaking circuit changes
+- `halo2_proofs::circuit::floor_planner::V1` was relying internally on the Rust
+  standard library's [`slice::sort_unstable_by_key`]; while it is deterministic,
+  it is not stable across targets or compiler versions. In particular, an edge
+  case within the sorting algorithm differed between 32-bit and 64-bit targets.
+  This meant that some circuits (like the [Orchard circuit]) would be layed out
```
sellout, line 14: Nit:
```suggestion
  This meant that some circuits (like the [Orchard circuit]) would be laid out
```

halo2_proofs/src/circuit/floor_planner/v1/strategy.rs
```diff
@@ -211,7 +211,23 @@ pub fn slot_in_biggest_advice_first(
             .count();
         // Sort by advice area (since this has the most contention).
         advice_cols * shape.row_count()
-    });
+    };
+
+    // This used to incorrectly use `sort_unstable_by_key` with non-unique keys, which is
+    // nondeterministic (in particular, it differs between 32-bit and 64-bit platforms).
```
sellout, line 217: Inconsistent with the CHANGELOG description
```suggestion
    // This used to incorrectly use `sort_unstable_by_key` with non-unique keys, which is
    // deterministic but unstable (in particular, it differs between 32-bit and 64-bit platforms).
```

sellout (approved):
utACK

Some non-blocking doc-related nits.

halo2_proofs/src/circuit/floor_planner/v1/strategy.rs
daira: "Unstable" is ambiguous here. "Nondeterministic" is correct in the sense that it's dependent on something that is not an explicit input to the function (I've always thought of platform dependence as a form of nondeterminism for that reason).
```suggestion
    // This used to incorrectly use `sort_unstable_by_key` with non-unique keys, which gave
    // output that differed between 32-bit and 64-bit platforms, and potentially between Rust
    // versions.
```

daira (commented):
re-utACK. Only the changelog and comments have changed so this does not require 3 reACKs; one more will do.

daira:
@sellout said that @nuttycom's approval was intact; merging.

Issue #740, opened 2023-02-25
Add `Region::instance_value` method, or figure out how `RegionLayouter::instance_value` is made accessible


str4d:
#625 added the `RegionLayouter::instance_value` method, but AFAICT it is not accessible to anyone, because what users actually interact with is the `Region` struct. But it seems like this method was added in order to be used by https://github.com/zcash/halo2/pull/529#discussion_r916302260, so maybe it is already reachable in some other way? We need to investigate, and most likely add `Region::instance_value`.

PR #741, opened 2023-02-25
Create a `TracingFloorPlanner` for debugging purposes

### LINKED ISSUE #738: Create a `TracingFloorPlanner` for debugging purposes
The Halo 2 APIs provide a bunch of slots for annotations, namespaces etc. This is used in the dev tooling for debugging purposes. But it would also be useful to be able to just log everything that a given circuit passes through or uses.

One way to do this would be a `struct TracingFloorPlanner<P: FloorPlanner>`, which people could wrap around their floor planner of choice in their circuits. It would be a pass-through floor planner, except that it evaluates all annotations and represents them as `tracing` spans and events. Then we would be able to filter this information for whatever subsets of the circuit we care about.

str4d:
Closes #738.

### DIFF
--- a/halo2_proofs/src/circuit/value.rs
+++ b/halo2_proofs/src/circuit/value.rs
@@ -63,6 +63,11 @@ impl<V> Value<V> {
         }
     }
 
+    /// ONLY FOR INTERNAL CRATE USAGE; DO NOT EXPOSE!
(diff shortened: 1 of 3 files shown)


str4d:
Currently broken because of #740.

daira (approved):
utACK

PR #750, opened 2023-03-16
`error::Error::TableCellsNotAssigned`: better error for unassigned table cells

therealyingtong:


halo2_proofs/src/plonk/error.rs
```diff
@@ -36,6 +36,8 @@ pub enum Error {
     /// The instance sets up a copy constraint involving a column that has not been
     /// included in the permutation.
     ColumnNotInPermutation(Column<Any>),
+    /// A column in the table has unassigned cells.
+    TableCellsNotAssigned,
```
str4d, line 40: This should be added to the changelog as an addition.

halo2_proofs/src/circuit/floor_planner/v1.rs
```diff
@@ -321,7 +321,7 @@ impl<'p, 'a, F: Field, CS: Assignment<F> + 'a> AssignmentPass<'p, 'a, F, CS> {
                     _ => None,
                 }) {
                 Some(Some(len)) => len,
-                _ => return Err(Error::Synthesis), // TODO better error
+                _ => return Err(Error::TableCellsNotAssigned),
```
daira, line 324: This code is duplicated between layouter implementations that have a `default_and_assigned` field. If it were implemented as a helper function that takes `default_and_assigned` as input and returns a `Result<usize, Error>` for the length, then it could be deduplicated, and also that would make it worthwhile to spend more effort collecting error information.

I.e. something like
```rust
let first_unused = assigned_length(default_and_assigned)?;
```
where
```rust
fn assigned_length(
    default_and_assigned: &HashMap<TableColumn, (DefaultTableValue<F>, Vec<bool>)>
) -> Result<usize, Error> {
    let compute_lengths = || default_and_assigned
        .values()
        .map(|(_, assigned)| {
            if assigned.iter().all(|b| *b) {
                Some(assigned.len())
            } else {
                None
            }
        });
    match compute_lengths()
        .reduce(|acc, item| match (acc, item) {
            (Some(a), Some(b)) if a == b => Some(a),
            _ => None,
        }) {
        Some(Some(len)) => len,
        _ => {
            let error_info = ...; // based on the result of compute_lengths()
            Err(Error::TableCellsNotAssigned::new(error_info))
        }
    }
}
```

therealyingtong: I've tried out this approach in https://github.com/zcash/halo2/pull/755.

therealyingtong:
Closing in favour of https://github.com/zcash/halo2/pull/755

PR #752, opened 2023-03-20
`dev::cost`: Handle cost calculation for circuits without permutation

### LINKED ISSUE #748: halo2_proofs cost.rs bug
It gives `"thread panicked at 'attempt to subtract with overflow', halo2-d6fd4df1666d8b25/642924d/halo2_proofs/src/dev/cost.rs:366:54"` when there isn't enable_equality on any rows or any permutation checks (i.e. chunks is 0, so `3 * chunks - 1` is negative). Adding one enable equality solves the issue, but it should be possible to run this on circuits without permutations.

therealyingtong:
Closes #748


str4d (approved):
utACK

PR #758, opened 2023-03-21
Release 0.3.0 of halo2_proofs and halo2_gadgets

ebfull:


nuttycom (commented):
It seems like there are a lot of actionable lints being reported. Also, I tried a build of this branch with `--all-features` and it failed to build due to a MSRV conflict in the `image` crate, so an earlier version of that might need to be pinned.

daira (approved):
utACK modulo @nuttycom's comment.

halo2_proofs/CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ and this project adheres to Rust's notion of
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+
+## [0.3.0] - 2023-03-21
```
str4d, line 10: I forgot to add `TracingFloorPlanner` to this changelog in #741 so we should do so here.

halo2_gadgets/CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ and this project adheres to Rust's notion of
 [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+
+## [0.3.0] - 2023-03-21
```
str4d, line 10: Add `halo2_proofs 0.3` to the "Migrated to" list below.

str4d (approved):
utACK after changelog fixes.

str4d:
> It seems like there are a lot of actionable lints being reported.

Yep, they just need someone to spend some time resolving them. Not a blocker for the release.

> Also, I tried a build of this branch with `--all-features` and it failed to build due to a MSRV conflict in the `image` crate, so an earlier version of that might need to be pinned.

This is intended. The `image` crate comes in because of the `test-dev-graph` feature flag enabling `plotters/bitmap_backend`. This feature flag is not part of the public feature flag set, and not under MSRV. That aside, I'd consider even `dev-graph` to not be under MSRV, as it's a circuit development-time dependency, not a circuit runtime dependency, so it affects a much smaller set of downstream users. But for now, we do include it in MSRV (everything we do check under MSRV is in the `prepare` action).

str4d (approved):
utACK

PR #759, opened 2023-03-31
circuit: Derive PartialEq/Eq for RegionIndex, Cell, and Value.

parazyd:
Using this, inside chips/gadgets we can do something like the following:

```rust
    if x.value().copied().to_field() == Value::known(F::one()).to_field() {
        // do something
    }
```

### DIFF
--- a/halo2_proofs/src/circuit.rs
+++ b/halo2_proofs/src/circuit.rs
@@ -46,7 +46,7 @@ pub trait Chip<F: Field>: Sized {
 }
 
 /// Index of a region in a layouter
-#[derive(Clone, Copy, Debug)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub struct RegionIndex(usize);
 
 impl From<usize> for RegionIndex {
@@ -82,7 +82,7 @@ impl std::ops::Deref for RegionStart {
 }
 
 /// A pointer to a cell within a circuit.
-#[derive(Clone, Copy, Debug)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub struct Cell {
     /// Identifies the region in which this cell resides.
     region_index: RegionIndex,
--- a/halo2_proofs/src/circuit/value.rs
+++ b/halo2_proofs/src/circuit/value.rs
@@ -12,7 +12,7 @@ use crate::plonk::{Assigned, Error};
 ///   helps to ensure that unwitnessed values correctly propagate.
 /// - It provides pass-through implementations of common traits such as `Add` and `Mul`,
 ///   for improved usability.
-#[derive(Clone, Copy, Debug)]
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
 pub struct Value<V> {
     inner: Option<V>,
 }


therealyingtong (approved):

str4d:
> Using this, inside chips/gadgets we can do something like the following:
> 
> ```rust
>     if x.value().copied().to_field() == Value::known(F::one()).to_field() {
>         // do something
>     }
> ```

Under what conditions is this sound? If `// do something` includes performing any assignments, then in the simplest case the circuit will differ between keygen and proving, creating very hard to debug errors.

Before considering merging this, I want to see example code that would use it, so we can consider whether this is the best approach.

str4d (changes requested):
PR should include an example of how `impl PartialEq for Value` would be used.

parazyd:
@str4d Something like this function:

```rust
pub fn conditional_select(
    &self,
    layouter: &mut impl Layouter<F>,
    a: AssignedCell<F, F>,
    b: AssignedCell<F, F>,
    cond: AssignedCell<F, F>,
) -> Result<AssignedCell<F, F>, Error> {
    let config = self.config();
    let out = layouter.assign_region(
        || "conditional_select",
        |mut region: Region<'_, F>| {
            config.s_cs.enable(&mut region, 0)?;

            a.copy_advice(|| "copy a", &mut region, config.advices[0], 0)?;
            b.copy_advice(|| "copy b", &mut region, config.advices[1], 0)?;

            let cond = cond.copy_advice(|| "copy cond", &mut region, config.advices[3], 0)?;

            let selected =
                if cond.value().copied().to_field() == Value::known(F::one()).to_field() {
                    a.value().copied()
                } else {
                    b.value().copied()
                };

            let cell =
                region.assign_advice(|| "select result", config.advices[2], 0, || selected)?;
            Ok(cell)
        },
    )?;
    Ok(out)
}

```

parazyd:
I saw this code in https://github.com/young-rocks/rocks-smt

str4d:
This can be written using the current API as:
```rust
pub fn conditional_select(
    &self,
    layouter: &mut impl Layouter<F>,
    a: AssignedCell<F, F>,
    b: AssignedCell<F, F>,
    cond: AssignedCell<F, F>,
) -> Result<AssignedCell<F, F>, Error> {
    let config = self.config();
    let out = layouter.assign_region(
        || "conditional_select",
        |mut region: Region<'_, F>| {
            config.s_cs.enable(&mut region, 0)?;

            a.copy_advice(|| "copy a", &mut region, config.advices[0], 0)?;
            b.copy_advice(|| "copy b", &mut region, config.advices[1], 0)?;

            let cond = cond.copy_advice(|| "copy cond", &mut region, config.advices[3], 0)?;

            let selected = cond.value()
                .to_field()
                .zip(a.value())
                .zip(b.value())
                .map(|((cond, a), b)| if cond == F::one().into() {
                    a
                } else {
                    b
                });

            let cell =
                region.assign_advice(|| "select result", config.advices[2], 0, || selected)?;
            Ok(cell)
        },
    )?;
    Ok(out)
}
```

This also fixes a small bug in the code: if `cond = Value::known(F::one())` and `a = Value::known(a_val)` but `b = Value::unknown()`, then the original code would set `selected = Value::known(a_val)`, while the above code will correctly propagate the unknown value and set `selected = Value::unknown()` (because it's a bug for this function to be called at proving time without known values for all input `AssignedCell`s, as that would lead to non-deterministic proof failures depending on which inputs are or are not known).

This further indicates to me that `impl PartialEq for Value` is a misleading approach here. My next question is, what would be a better or more discoverable API here? `Value::map` is a functional programming approach, whereas the `impl PartialEq for Value` enables an imperative programming approach that maybe more downstream users want. Unfortunately the simple imperative approaches make it very easy to introduce bugs, but maybe there are other helper methods we could add to solve the specific issues people run into. For example, we already have a `CondSwap` chip in `halo2_gadgets` that we should turn into a more usable gadget, and we could then provide a `ConditionalSelect` gadget that downstream users can use instead of writing their own.

PR #760, opened 2023-04-01
dev: Add `halo2_proofs::dev::render_to_json` API

str4d:
This exposes the internally-collected data used by `CircuitLayout`, to enable alternative external visualisations.


str4d:
Here is the Orchard Action circuit rendered using the current version of this API:

[orchard-circuit.json.txt](https://github.com/zcash/halo2/files/11128601/orchard-circuit.json.txt)

PR #761, opened 2023-04-03
ConstraintSystem::set_minimum_degree: do not overwrite existing larger degree

therealyingtong:
`set_minimum_degree()` should never overwrite an existing larger degree.

### DIFF
--- a/halo2_proofs/src/plonk/circuit.rs
+++ b/halo2_proofs/src/plonk/circuit.rs
@@ -1188,7 +1188,9 @@ impl<F: Field> ConstraintSystem<F> {
     /// larger amount than actually needed. This can be used, for example, to
     /// force the permutation argument to involve more columns in the same set.
     pub fn set_minimum_degree(&mut self, degree: usize) {
-        self.minimum_degree = Some(degree);
+        self.minimum_degree = self
+            .minimum_degree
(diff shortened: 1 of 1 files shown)


str4d:
I am unsure why this is a bug. The minimum degree can be set to smaller than the circuit's required degree (e.g. less than the degree 3 required by the permutation argument), and it will in that case be ignored:
https://github.com/zcash/halo2/blob/eec65ead3bf5288162efa2bb2c23060db4258100/halo2_proofs/src/plonk/circuit.rs#L1421-L1431

The question here is rather, should it be possible to pick an initial candidate minimum degree, and then be allowed to change it arbitrarily later (or, after this PR, only ratchet it up)? I think yes, because I can imagine an optimisation process where the optimiser tries using a larger degree, and then lowers it as it tests other optimisations. This could also be done by completely re-running configuration, but in that case we may as well just make `ConstraintSystem::set_minimum_degree` callable only once. So from a builder pattern perspective, the current behaviour seems more correct and comprehensible.

As an aside, there could easily be a bug in one of the downstream forks of Halo 2, if they use the `minimum_degree` field for anything other than what we intended it for (a user-configurable floor on the `ConstraintSystem::degree` function). But that is outside the scope for this repo.

Issue #749, opened 2023-03-16
cost.rs for proving time?


Divide-By-0:
Is it possible to formalize the knowledge about the number of FFTs and MSMs for each type of gate/lookup/permutation check etc [as Ye explains](https://learn.0xparc.org/materials/halo2/learning-group-1/cost-model) into a cost.rs file for optimizing proving time, along with the breakdown of the # of FFTs, MSMs, and poly evaluations for each gate/lookup in my code so I can optimize that?

PR #753, opened 2023-03-20
`dev::cost`: Estimate prover work for a given circuit

therealyingtong:
Closes #749. Computes prover cost in terms of the number of:
- FFTs over the `2^k` domain,
- extended FFTs over `2^extended_k` domain, and
- MSMs of length `2^k`.

What other prover operations are significant? (e.g. polynomial evaluations)

PR #763, opened 2023-04-03
plonk::prover: Introduce counter feature for FFTs and MSMs

therealyingtong:
Closes https://github.com/zcash/halo2/issues/749.

Supersedes https://github.com/zcash/halo2/pull/753/files.


therealyingtong:
I realise that this may not be useful for people who want a proof estimate but DON'T want to run `create_proof`.

PR #769, opened 2023-04-20
Zsa functionalities 

PaulLaux:
The following modifications are required to support the ZSA circuit changes:
- Added a variable-base sign-scalar multiplication called `mul_sign`. The implementation is based on the `q_mul_fixed_short` gate.
- Added a function `CommitDomain::new_with_personalization` to initialize the `CommitDomain` from two different personalizations. This is required to be able to use an identical `R` for ZEC and ZSA note commitments but a different `Q`.
- Exposed the `HashDomain` in a `CommitDomain` publicly to enable the usage of Sinsemilla hash function from a `CommitDomain`.
- Added a function `commit_from_hash_point` to evaluate a Sinsemilla commit from a hash point.
- Added the ability to `hash` and `blind` independently.
- Added a multiplexer function `MUX(choice, left, right) = if {choice} right else left`
- Added a function `Point::new_from_constant` to create a Point from a constant.
- Optimized short range check on 4 and 5 bits to evaluate them with one lookup table instead of two
- Added a function `hash_to_point_with_private_init` to evaluate a Sinsemilla hash from a private initial point (instead of a public initial point)


PaulLaux:
Closed in favor of smaller iterations, starting from: https://github.com/zcash/halo2/pull/823

PR #773, opened 2023-05-01
Update rust-toolchain to 1.61.0

therealyingtong:
This is needed for the package `image v0.24.6`.

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,2 +1,2 @@
 [toolchain]
-channel = "1.60.0"
+channel = "1.61.0"


str4d:
The `image` dependency is only present because of the `test-dev-graph` feature flag, which is only present for `halo2` devs to run tests. So we need to ensure that CI is still checking against MSRV after this change. I think it is, but only because we're using the deprecated `actions-rs/toolchain`. In other repos I've migrated away from that by using `rust-toolchain.toml` to encode the MSRV, and this PR would prevent that.

str4d:
We will bump the MSRV when we cut the next breaking release of `halo2_proofs`, which will likely be when `ff` is updated.

PR #777, opened 2023-05-21
[book] Fix `p_poly` to match implementation; specify synthetic blinding factor `f` construction

therealyingtong:
`p_poly` is constructed as a linear combination of q_prime and the q_polys in steps 18 and 19 of the protocol description. Previously, the expression used in the protocol description did not match the implementation:

- https://github.com/zcash/halo2/blob/5678a506cbec60d593a21ff618d09bed48abf7f2/halo2_proofs/src/poly/multiopen/prover.rs#L105-L113
- https://github.com/zcash/halo2/blob/5678a506cbec60d593a21ff618d09bed48abf7f2/halo2_proofs/src/poly/multiopen/verifier.rs#L120-L129

PR #779, opened 2023-05-25
Patch protocol p poly

lassebramer:
I was unable to render the markdown locally, so there might be some errors...

The wording in step one might need some more explanation

Finally as Rasmus pointed out [here](https://github.com/zcash/halo2/pull/777?fbclid=IwAR3-iYgdTQVlz-_ePysX8TRTNMgMkLe6fM_VS2-JoNF6BoVSfmQL0tqoZCI#discussion_r1204307871) there might have been a typo in the last fix so i changed that aswell.


therealyingtong:
I've merged this into #777.

PR #788, opened 2023-07-30
halo2_proofs: Pin dev-dependencies to versions compatible with MSRV

str4d:


str4d:
Self-ACK; CI passes with this change, and the pinning only affects dev-dependencies so does not affect the public API.

PR #792, opened 2023-08-22
pow_vartime([n, 0, 0, 0] => pow_vartime([n]

zhiqiangxu:
`pow_vartime([n, 0, 0, 0] => pow_vartime([n]` always holds, so it seems there's no point in wasting resource computing `pow_vartime([n, 0, 0, 0] ` instead of `pow_vartime([n]`.

### DIFF
--- a/halo2_proofs/src/plonk/permutation/prover.rs
+++ b/halo2_proofs/src/plonk/permutation/prover.rs
@@ -128,7 +128,7 @@ impl Argument {
                     Any::Instance => instance,
                 };
                 parallelize(&mut modified_values, |modified_values, start| {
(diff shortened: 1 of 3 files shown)


str4d:
This is probably left over from when `pow_vartime` was on `FieldExt` and took a fixed-length array (IIRC). Generally +1 on the PR, but I'm on PTO and am not reviewing Zcash PRs currently.

zhiqiangxu:
> This is probably left over from when `pow_vartime` was on `FieldExt` and took a fixed-length array (IIRC). Generally +1 on the PR, but I'm on PTO and am not reviewing Zcash PRs currently.

No hurry:)

BTW, it seems `pow_vartime` and `pow` are almost the same, is the difference really that big or another left over to be unified?

PR #798, opened 2023-10-10
feat: update rust toolchain to 1.66.0

hunjixin:
update toolchain to resolve
```
error: package `rayon-core v1.12.0` cannot be built because it requires rustc 1.63 or newer, while the currently active rustc version is 1.60.0


error: package `constant_time_eq v0.3.0` cannot be built because it requires rustc 1.66.0 or newer, while the currently active rustc version is 1.63.0
```

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,2 +1,2 @@
 [toolchain]
-channel = "1.60.0"
+channel = "1.66.0"


str4d:
#805 adds a lockfile that enables us to maintain our current MSRV of 1.60. So there is no urgency to bump it to 1.66 (and we should instead look at what features bumping MSRV gains us, and what fraction of the userbase will become unable to use the crates).

PR #805, opened 2023-11-22
Add `Cargo.lock` to repository

str4d:
This is based on "whatever lockfile happened to last work for str4d" (with some updates), but going forward will be the lockfile we use for testing our MSRV. See https://blog.rust-lang.org/2023/08/29/committing-lockfiles for rationale on this change.

The extra packages added to `Cargo.lock` in 3cd1bed5a2b0707d473619a1ada03082be497581 are the ones needed when building/testing with `--all-features` after adding the `plotters/ttf` feature as a dependency of `test-dev-graph`, as resolved by `cargo +1.60`.

This PR also makes `rust-version` in subdirectories' `Cargo.toml` files consistent with the top-level `rust-toolchain.toml`, i.e. 1.60, and adds some generated/temporary files to `.gitignore`.


ebfull (approved):
ACK, didn't check the hashes but I think that's fine.

PR #831, opened 2024-12-13
Move Poseidon primitive into `halo2_poseidon`

str4d:


halo2_gadgets/src/poseidon.rs
```diff
@@ -180,7 +171,8 @@ impl<
             &mut self.state,
             Some(&self.mode),
         )?;
-        self.mode = Absorbing::init_with(value);
+        self.mode = Absorbing::init_empty();
+        self.mode.absorb(value).expect("state is not full");
```
nuttycom, line 175: Doesn't this suggest that https://github.com/zcash/halo2/pull/831/commits/0d883685d191e019915db41c7c2124f09e0f1464#diff-76142ee2fee95e1ba7ebdff8392dfb7442517333aaf489d5cf92d6dcebee37a4R190-R194 should be public?

halo2_poseidon/src/lib.rs
```diff
@@ -176,16 +188,82 @@ impl<F, const RATE: usize> SpongeMode for Squeezing<F, RATE> {}
...
                 .unwrap(),
         )
     }
 }
 
+impl<F, const RATE: usize> Absorbing<F, RATE> {
+    /// Attempts to absorb a value into the sponge state.
+    ///
+    /// Returns the value if it was not absorbed because the sponge is full.
+    pub fn absorb(&mut self, value: F) -> Result<(), F> {
```
nuttycom, line 212: Is `Result` the right type to return here? Or should it be a bespoke enum? It looks to me like the kind of thing that I would ordinarily use `Either` for rather than `Result`, but a bespoke enum might be better.

nuttycom (approved):
utACK 0d883685d191e019915db41c7c2124f09e0f1464 with two minor questions.

str4d:
[Rebased on `main`](https://github.com/zcash/halo2/compare/0d883685d191e019915db41c7c2124f09e0f1464..589385033e1e2971556c6e1f081d7520a17c2638) to fix CI.

nuttycom (approved):
re-utACK 589385033e1e2971556c6e1f081d7520a17c2638

PR #843, opened 2025-03-19
halo2_gadgets: depend on `rand` with default-features = false

daira:
The dependency on `rand` is now without default features, fixing use of `halo2_gadgets`  in `no_std` environments.


str4d:
`halo2_gadgets` is not a no-std crate. And IIRC we determined that the no-std issue that led to this PR being opened was actually a different issue in the dependency tree.

PR #847, opened 2025-07-21
Implement variable-length input hashing.

71104:


str4d (changes requested):
Changes look fine but I'd prefer to have test vectors.

Additionally, in order for this to be a patch release, it needs to be based on f066ace1f234d7fe1908851ed86b1801e0b1ffea (the `halo2_poseidon 0.1.0` release commit).

PR #850, opened 2025-08-25
docs: add Pallas-Vesta curves and Simplified SWU documentation

crStiv:
Fill in TODO sections in curves.md with:

- Pallas-Vesta curves: 2-cycle structure, field parameters, and curve equations
- Simplified SWU: hash-to-curve algorithm description and implementation steps

### DIFF
--- a/book/src/background/curves.md
+++ b/book/src/background/curves.md
@@ -271,7 +271,22 @@ over $E_q/\mathbb{F}_q,$ forming a 2-cycle with the first:
 
 ![](https://i.imgur.com/bNMyMRu.png)
 
-### TODO: Pallas-Vesta curves
+### Pallas-Vesta curves
+
+The Pallas and Vesta curves form a 2-cycle of elliptic curves designed specifically for Halo 2. They are defined over finite fields with highly 2-adic structure, meaning they have large multiplicative subgroups of order $2^S$ where $S = 32$.
+
+**Pallas curve ($E_p/\mathbb{F}_p$):**
+- Base field: $\mathbb{F}_p$ where $p = 2^{254} + t_p$ and $t_p = 45560315531419706090280762371685220353$
+- Scalar field: $\mathbb{F}_q$ where $q = 2^{254} + t_q$ and $t_q = 45560315531506369815346746415080538113$
+- Curve equation: $y^2 = x^3 + 5$
+
+**Vesta curve ($E_q/\mathbb{F}_q$):**
+- Base field: $\mathbb{F}_q$ (same as Pallas scalar field)
+- Scalar field: $\mathbb{F}_p$ (same as Pallas base field)
+- Curve equation: $y^2 = x^3 + 5$
+
+This 2-cycle structure enables efficient recursive proof composition, where proofs generated on one curve can be efficiently verified on the other curve. The highly 2-adic structure ($p-1 = T \cdot 2^{32}$ with $T$ odd) provides efficient FFT operations and enables a wide variety of circuit sizes.
+
 Reference: https://github.com/zcash/pasta
 
 ## Hashing to curves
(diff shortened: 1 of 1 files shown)


book/src/background/curves.md
```diff
@@ -271,7 +271,22 @@ over $E_q/\mathbb{F}_q,$ forming a 2-cycle with the first:
 
 ![](https://i.imgur.com/bNMyMRu.png)
 
-### TODO: Pallas-Vesta curves
+### Pallas-Vesta curves
+
+The Pallas and Vesta curves form a 2-cycle of elliptic curves designed specifically for Halo 2. They are defined over finite fields with highly 2-adic structure, meaning they have large multiplicative subgroups of order $2^S$ where $S = 32$.
```
daira:
```suggestion
The Pallas and Vesta curves form a 2-cycle of elliptic curves designed to allow efficient recursion in discrete-log-based proof systems such as Halo 2. They are defined over finite fields with highly 2-adic structure, meaning they have large multiplicative subgroups of order $2^S$, where in this case $S = 32$.
```

book/src/background/curves.md
```diff
@@ -271,7 +271,22 @@ over $E_q/\mathbb{F}_q,$ forming a 2-cycle with the first:
...
+
+**Pallas curve ($E_p/\mathbb{F}_p$):**
+- Base field: $\mathbb{F}_p$ where $p = 2^{254} + t_p$ and $t_p = 45560315531419706090280762371685220353$
+- Scalar field: $\mathbb{F}_q$ where $q = 2^{254} + t_q$ and $t_q = 45560315531506369815346746415080538113$
+- Curve equation: $y^2 = x^3 + 5$
```
daira, line 281: I checked that the values of $t_p$ and $t_q$ match https://github.com/zcash/pasta (and I know everything else does).

book/src/background/curves.md
```diff
@@ -290,7 +305,28 @@ framework used in the Internet Draft makes use of several functions:
 
 [cfrg-hash-to-curve]: https://datatracker.ietf.org/doc/draft-irtf-cfrg-hash-to-curve/?include_text=1
 
-### TODO: Simplified SWU
+### Simplified SWU
+
+The Simplified SWU (Shallue-van de Woestijne-Ulas) method is an efficient hash-to-curve algorithm that maps field elements to curve points. It is particularly well-suited for curves of the form $y^2 = x^3 + b$ where $b \neq 0$.
+
+For a field element $u \in \mathbb{F}_p$, the Simplified SWU algorithm works as follows:
```
daira: This matches the core of the algorithm in the paper, but it doesn't mention that for curves with j-invariant 0 like Pallas and Vesta, you need to do the hash-to-curve operation for an isogenous curve and then map to the target curve using the isogeny. There are additional complications as noted in my second suggestion below.

It probably isn't useful to spell out all the details here. Instead we should point to the [Sage reference implementation](htttps://github.com/zcash/pasta/blob/master/hashtocurve.sage) (which is algorithmically optimized, including some changes needed to support constant-time operation in a real implementation*, but well commented).

*Obviously Sage itself can't support constant-time operation.

```suggestion
For a field element $u \in \mathbb{F}_p$, the core of the Simplified SWU algorithm works as follows:
```

book/src/background/curves.md
```diff
@@ -290,7 +305,28 @@ framework used in the Internet Draft makes use of several functions:
...
+   - $x_1 = t_1^3 + b$
+   - $x_2 = t_2^3 + b$
+   - $x_3 = Z \cdot (t_1^2 \cdot t_2^2)$
+
+3. **Selection**: Choose the first $x_i$ that is a square in $\mathbb{F}_p$
+4. **Square root**: Compute $y = \sqrt{x_i^3 + b}$
+5. **Sign adjustment**: If $u \cdot y$ is negative, negate $y$
+
+The result is the curve point $(x_i, y)$.
+
```
daira, line 327:
```suggestion

In practice for curves with $j$-invariant $0$ such as Pallas and Vesta, this algorithm cannot be used as-is. Instead we map $u$ to an isogenous curve with a non-zero $j$-invariant ("iso-Pallas" or "iso-Vesta"), and then apply the isogeny to give a point on the target curve. Also, the full hash-to-curve operation involves pre-hashing the input with domain separation, and adding the results of two such mappings to the target curve (or equivalently, adding them on the isogenous curve and then applying the isogeny). For full details see the [Sage reference implementation](https://github.com/zcash/pasta/blob/master/hashtocurve.sage).

```

daira (dismissed):
ACK with suggestions.

daira:
Review needed from @ebfull or @str4d.

PR #851, opened 2025-09-02
fix: type declaration order, typo, and improve variable clarity

0xwitty:
just tidied up a few things:

* moved `TableError` above `Error` so Rust stops complaining about order.
* fixed a typo in the message: `enable_equalty` → `enable_equality`.
* renamed `(table, table_len)` → `(col2, col2_len)` in `TableError::UnevenColumnLengths` to make it clear we’re talking about a column, not the whole table.


str4d:
Rust doesn't require type declaration ordering like C-based languages.

The rename is also semantically incorrect.

No thanks, AI.

PR #859, opened 2025-12-04
CI: Build book with `mdbook 0.4`

str4d:
`mdbook 0.5` was recently released, but `mdbook-katex` doesn't yet support it.


nuttycom (commented):
post-hoc utACK

PR #868, opened 2025-12-05
perf(ecc): remove redundant y_check from short fixed-base mul gate

MozirDmitriy:
Removed the redundant y_check constraint from the short fixed-base mul gate in mul_fixed/short.rs. This constraint was implied by sign_check and negation_check, adding proving cost without increasing soundness. Updated tests to reflect the new constraint index (negation_check: 3 → 2). Behavior is unchanged; only the constraint set is smaller.


daira (changes requested):
These constraints are used in a production circuit and therefore cannot be changed without some kind of opt-in configuration. Also the saving is negligible.

PR #887, opened 2026-05-05
Add NonIdentityPoint::new_from_constant

ConstanceBeguier:
This PR adds `NonIdentityPoint::new_from_constant` to enable creating non-identity points that are properly pinned to constants, which is required by the Orchard ZSA circuit ([Orchard PR](https://github.com/QED-it/orchard/pull/246)).

More precisely, in the Orchard ZSA circuit, `q_init_zec` / `q_init_zsa` must be constrained to fixed constants, as they define the initial point `Q` of the Sinsemilla hash. If constructed via `NonIdentityPoint::new`, they remain unconstrained witnesses, allowing a prover to inject an arbitrary on-curve point and break commitment soundness.

Note that this is a breaking change for `halo2_gadgets` because it adds a method to a public trait (`EccInstructions`).


halo2_gadgets/src/test_circuits/circuit_data/vk_ecc_chip.rdata
```diff
@@ -9878,38 +9878,38 @@ PinnedVerificationKey {
     },
     fixed_commitments: [
         (0x05f5862cad2888855bc3c1843a9eff57b11b592d9eb0e13354256661387f5231, 0x32236b14df85bf5f532a930232cb23a5c56ef7d67aaeed8bcb8fc10ea132cbd6),
-        (0x0b70f6f89e88131bb5ab28571ca837a4160b51998c6ced04328991a5c3bdcae9, 0x2cee3ef6523c2d36e01981c4794271d53fb47e9c5480b982a2ff3a869144d648),
```
daira, line 9881: This changes because of the call to `super::chip::witness_point::tests::test_witness_non_id` in `MyEccCircuit::synthesize`. But we need to ensure circuit stability, so actually `MyEccCircuit` should have a `test_zsa_additions` flag that is set to false in order to check stability of the existing fixtures (both `_insecure` and `_fixed` after halo2_gadgets 0.5.0), and to true for the circuit variant that tests `non_constant_point_id` as well (there we only need the "fixed" variant).

daira (changes requested):
Blocking requests:
* rebase on top of `halo2_gadgets 0.5.0` when that merges;
* add a flag to `MyEccCircuit` so that the existing stability checks (`test_ecc_chip_*_against_stored_circuit`) continue to pass without changing the `*.bin,*.rdata` fixture files; and add new stability checks with new fixture files (only the `circuit_version = AnchoredBase` variant needed).

Otherwise this looks good, we'll merge it and release as `halo2_gadgets 0.5.1`.

ConstanceBeguier:
> Blocking requests:
> 
> * rebase on top of `halo2_gadgets 0.5.0` when that merges;
> * add a flag to `MyEccCircuit` so that the existing stability checks (`test_ecc_chip_*_against_stored_circuit`) continue to pass without changing the `*.bin,*.rdata` fixture files; and add new stability checks with new fixture files (only the `circuit_version = AnchoredBase` variant needed).
> 
> Otherwise this looks good, we'll merge it and release as `halo2_gadgets 0.5.1`.

All items have been addressed.

daira (dismissed):
ACK modulo the suggestion. Please add a CHANGELOG entry; this is technically a breaking change because it adds a method to a trait.

halo2_gadgets/src/ecc.rs
```diff
@@ -802,7 +831,7 @@ pub(crate) mod tests {
         type FloorPlanner = SimpleFloorPlanner;
 
         fn without_witnesses(&self) -> Self {
-            MyEccCircuit::with_version(false, self.circuit_version)
+            MyEccCircuit::with_version(false, false, self.circuit_version)
```
daira: Nit, fixes a latent testing bug:
```suggestion
            MyEccCircuit::with_version(false, self.test_zsa_additions, self.circuit_version)
```
Flags that change the circuit should be preserved by `without_witnesses`. (`test_errors` doesn't change the circuit; `circuit_version` does, and so `test_zsa_additions` should be treated like the latter.)

This didn't cause a test failure because keygen doesn't go through `without_witnesses`. But anything that did go through `without_witnesses` would be wrong for the test circuit.

(Found by Fable 5; I wouldn't have spotted it. Built locally and passes `cargo fmt --check`.)

ConstanceBeguier:
> ACK modulo the suggestion. Please add a CHANGELOG entry; this is technically a breaking change because it adds a method to a trait.

Applied the suggestion and updated CHANGELOG.

daira (approved):
utACK

daira (approved):
utACK

daira (approved):
utACK

daira (approved):
utACK

daira (approved):
utACK

PR #911, opened 2026-06-25
Clarify zero-window handling in decompose_word

daixihegu:
This PR clarifies the panic behavior of `halo2_gadgets::utilities::decompose_word` when called with a zero-bit window.

Previously, `window_num_bits = 0` reached the padding calculation and panicked due to division/remainder by zero. This change adds an explicit assertion before that calculation and updates the panic documentation to state that the window size must be in `1..=8`.

A regression test covers the zero-window case, and the `halo2_gadgets` changelog has been updated.

### DIFF
--- a/halo2_gadgets/CHANGELOG.md
+++ b/halo2_gadgets/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to Rust's notion of
 
 ## [Unreleased]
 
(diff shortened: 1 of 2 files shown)


daixihegu:
@ebfull @nuttycom Hi, Could you please review this PR at your convenience? Thank you very much.

PR #915, opened 2026-07-12
Update to Rust 1.88 / edition 2024, rebuild CI, add workspace lints

dannywillems:
Follows librustzcash, and mirrors [pasta_curves#90](https://github.com/zcash/pasta_curves/pull/90) and [orchard#532](https://github.com/zcash/orchard/pull/532). MSRV goes **1.60 -> 1.88** and all four crates move to edition 2024.

Also centralises `edition`/`rust-version`/`license`/`repository`/`categories` in `[workspace.package]` so they cannot drift between crates, and sets `resolver = "2"` (the virtual workspace was silently defaulting to resolver 1).

## Edition 2024

Most of the diff is the 2024 style edition. The migration itself needed one change: `NonIdentityPoint::constrain_equal` binds the converted point instead of leaving it a temporary of the tail expression, so its drop order relative to `other` no longer depends on the edition. Neither type has a side-effecting `Drop`, so this is a readability fix, not a behaviour fix.

The rest is warnings the 1.60 -> 1.88 jump surfaces, mostly from lints that did not exist on the old toolchain (`repeat_n`, `div_ceil`, `io::Error::other`, inlined format args, ...). **Two needed care rather than the mechanical fix:**

- `Domain::padding` **keeps** `repeat().take()`. `Self::Padding` is a public associated type naming the iterator concretely, so `repeat_n` (returning `iter::RepeatN`) cannot be used without changing the public API. Allowed with the reason recorded.
- Clippy's `redundant_closure_for_method_calls` fix in `circuit::value` and `poly` suggested a path through the **private** `plonk::assigned` module, which does not compile. Used the public `plonk::Assigned` re-export instead.

Separately, the `multicore` re-exports are aligned with their use: `ParallelIterator` and `IndexedParallelIterator` are only consumed by the batch verifier and only under `multicore`, but were re-exported unconditionally, so a build without that feature (which is exactly what CI runs) emitted unused-import and dead-code warnings.

## Workspace lints

The workspace had no lint configuration at all. Lints are now declared once in `[workspace.lints]` and inherited by all four crates, covering every target (lib, tests, benches, examples).

Deliberate exclusions, recorded next to the table:
- **`print_stdout` / `print_stderr` are NOT denied.** Printing is the job of `halo2_proofs::dev`, whose MockProver renders human-readable circuit failure diagnostics to stderr, and of the examples.
- The four existing `todo!()`s and two `unimplemented!()`s are deliberate, so they get an `allow` and a note; the deny then guards against *new* ones arriving unnoticed.

## CI

- The no-std build job targeted **`wasm32-wasi`, a target name removed in Rust 1.78** — it could not have worked on the new toolchain. Now `wasm32-wasip1`.
- Required jobs gated behind a single `required-checks` job to mark as required in branch protection; macOS/Windows, beta clippy and coverage stay advisory.
- Both lint workflows folded into ci.yml, replacing the archived `actions-rs/clippy-check` with `auguwu/clippy-action` (which lints-beta.yml had already moved to).
- New `build-latest` job, so the latest stable toolchain is exercised against the latest dependencies rather than only the pinned MSRV and the committed lockfile.
- Every action pinned by commit SHA, `permissions: {}` at the workflow level, minimum per job, `persist-credentials: false` on checkout; zizmor action updated.

## Known gap (pre-existing, left alone)

`cargo test --no-default-features` does **not** compile on `main` either: `tests/plonk_api.rs` imports `BatchVerifier` unconditionally, but it needs the `batch` feature. So this PR does not add a `--no-default-features` clippy job (pasta_curves and orchard got one). Worth a separate fix.

## Verified locally

On 1.88: `fmt --all --check`, `clippy -D warnings` under CI's exact feature set, under `--all-features`, and under default features; the full test suite (87 tests) with CI's feature set and with `--all-features`; and `mdbook test` on a clean target.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


dannywillems:
Closing as, after discussing with @nuttycom, we won't update to 1.88 until it is very necessary.

PR #918, opened 2026-07-21
Derive instance commitments from public inputs for Lean fixture

TalDerei:
Halo2 side PR for https://github.com/zcash/orchard/pull/538, and consumed in https://github.com/zcash/ironwood/pull/85.

Previously we supplied instance commitments derived from public inputs, but nothing on the Lean side verified that derivation, which made them glorified VK elements. The public instances weren't in Lean yet; now we supply them separately per action and check that each captured instance commitment is derived from them. What remains on the ironwood side is consuming that derived commitment in the Fiat–Shamir transcript and the opening queries so verification checks openings against the derived commitment rather than an opaque field. Concretely, this threads the public instances into `dump_vesta_lean_fixture` and emits the derivation check into Lean.

The public instances themselves (`capturedPublicInstances`), their derivation (`derivedInstanceCommitment`), and the captured commitments (`capturedInstanceCommitments`, the cross-check reference) all live as standalone defs/helpers in the fixture, separate from vk. vk mirrors the Rust `VerifyingKey` and touches none of them.

### DIFF
--- a/halo2_proofs/src/plonk/fingerprint/vesta_lean.rs
+++ b/halo2_proofs/src/plonk/fingerprint/vesta_lean.rs
@@ -10,12 +10,16 @@
 //! across the VK, proof string, transcript, and MSM.
 //!
 //! Trust boundary: the fixture attests that the assembled MSM matches, and evaluates to the
-//! identity, for *these* captured commitments and challenges. It does not re-derive the instance
-//! commitments from the public inputs (they enter as opaque points, like every other commitment),
-//! nor does it reproduce Halo2's Blake2b transcript or pinned-key serialization; those remain
-//! trusted from the Rust capture. Halo2's `MSM` also merges same-base terms and drops identity
-//! bases, where the Lean assembly deliberately does neither; such a capture is rejected at export
-//! (see [`VerifyingKey::dump_vesta_lean_fixture`]) rather than emitted.
+//! identity, for *these* captured commitments and challenges. The emitted `vk` mirrors halo2's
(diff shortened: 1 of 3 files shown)


halo2_proofs/src/plonk/fingerprint/vesta_lean.rs
```diff
@@ -333,6 +348,38 @@ impl VerifyingKey<EqAffine> {
...
+                let mut poly = column.to_vec();
+                poly.resize(params.n as usize, Fp::ZERO);
+                let poly = self.domain.lagrange_from_vec(poly);
+                let derived = params.commit_lagrange(&poly, Blind::default()).to_affine();
+                assert_eq!(
+                    derived,
+                    common_points[p * n_inst_cols + col],
+                    "supplied public instances do not reproduce the captured instance commitment \
+                     for proof {p}, column {col}; the instance-commitment derivation could not hold"
+                );
```
TalDerei, line 379: @ebfull you had asked something along the lines of like what if there's a verifier bug and we could create a proof and then find a new public input that produces the same public instance commitment, and would the proof catch this in some way?

well, `assert_eq!` demonstrates this. it ties instances to commitments up to **binding** (so if someone who could find a different instance with the same `commit_lagrange` would pass, but that's a DL-relation break). we'll exercise a similar binding argument on the Lean side.

ebfull (dismissed):

TalDerei:
good to merge on green CI in downstream https://github.com/zcash/ironwood/pull/85 which serves as a sanity check.

TalDerei:
my fork wasn't updated with the latest diff, which was causing downstream failures...sorry about that.

ebfull (approved):

PR #914, opened 2026-07-12
Add verifier fingerprint capture and Lean fixture export

TalDerei:
**Human Summary:**

Halo2 side PR for https://github.com/zcash/orchard/pull/531, and consumed in https://github.com/zcash/ironwood/pull/57. See the generated fixture in https://github.com/zcash/ironwood/blob/main/Zcash/Snark/Fixtures/MultiAction/Fixture.lean.

This feature gates (`unstable-verifier-fingerprint)` verifier tooling for downstream Lean fixtures. The Rust verifier can expose the assembled MSM fingerprint / FS challenges / ordered transcript events for comparing against the Lean version.

Concretely, 
  - Captures the verifier’s assembled MSM and Fiat–Shamir transcript,
  - Exports the VK, proof data, Vesta points, and full URS as Lean fixtures,
  - Generates Lean checks that the reconstructed MSM matches and evaluates to zero,

Open to organizational / feature gating preferences. Awaiting workflow approval.

----------

**AI Summary:**

This PR adds opt-in tooling that records one normal verifier run and exports it as a concrete Lean fixture. It captures the verifier's final MSM and transcript data; it does not create a new verification path or treat capture as proof acceptance.

## What Is Added

- `unstable-verifier-fingerprint` enables MSM and transcript capture together with the Vesta-specific Lean exporter.
- Capture uses the normal verifier and records proof inputs, common inputs, challenges, and transcript events.
- The exporter writes the verifying key, proof string, concrete Vesta points, full verifier URS, and captured MSM, while checking their counts and ordering.
- Generated fixtures check that Lean assembles the same MSM and that the concrete MSM evaluates to zero.
- The tooling is off by default; `SingleVerifier` remains the proof-acceptance API.

## Remaining Floor

This compares Lean against one Rust verifier run; it is not a soundness proof. The fixture still trusts Rust capture and export, and Lean does not independently derive Halo2's pinned-key serialization or Blake2b VK transcript representation.

PR #922, opened 2026-07-24
Split Lean VK CS data from fingerprint fixtures

mitschabaude:
## Summary

- replace the generated fixture's `import Zcash.Snark` with focused imports
- emit VK constraint-system data as a separate small Lean module
- make the proof-specific fixture reference the split gates, query layouts, permutation chunks, and lookup expressions
- use `SWPoint Vesta.curve` directly instead of importing the heavyweight `VestaG` alias

## Motivation

The generated `Fixture.lean` modules contain thousands of concrete points and an expensive native-decided MSM. Importing them for VK/circuit correspondence checks makes those checks unnecessarily slow, while the umbrella `Zcash.Snark` import also invalidates the fixtures after unrelated soundness changes.

`dump_vesta_lean_fixture` now returns `VestaLeanFixtureModules { fixture, vk_cs_data }` and accepts the Lean module name that `fixture` should import. Downstream generators should write both returned sources, for example `Fixture.lean` and `VkCsData.lean`.

This change was layered on top of zcash/halo2#914 and staged as [TalDerei/halo2#2](https://github.com/TalDerei/halo2/pull/2); with #914 merged, it now targets `main` directly (same three commits).

## Downstream

The fixtures checked into [zcash/ironwood](https://github.com/zcash/ironwood) (`Zcash/Snark/Fixtures/{SingleAction,MultiAction}/{Fixture,VkCsData}.lean`, currently on [zcash/ironwood#99](https://github.com/zcash/ironwood/pull/99)) are generated by this exporter, and ironwood's fixture-regeneration CI interim-pins `halo2_proofs` to this branch (merged with the `halo2_proofs-0.3.4` release for `[patch.crates-io]` version parity) plus an adapted orchard 0.15.3 capture harness; regeneration was verified byte-identical to all four checked-in modules. Once this PR lands in a `halo2_proofs` release and orchard's capture harness follows, those interim pins revert to upstream.

## Checks

- `cargo test -p halo2_proofs --features unstable-verifier-fingerprint --test fingerprint_export`
- `cargo test -p halo2_proofs --features unstable-verifier-fingerprint --test plonk_api --no-run`

🤖 Generated with [Claude Code](https://claude.com/claude-code)


TalDerei:
@mitschabaude this seems like just an organization thing, doesn't seem strictly necessary to cut another point release for? could we try to get the integration in https://github.com/zcash/ironwood/pull/99 to use the existing fixture that keeps everything in one exported file, unless it's insufficiently namespaced and just can't? or does this approach have any stronger guarantees i'm missing?

mitschabaude:
@TalDerei yeah it was just an efficiency thing (minimizing imports, so this slow thing don't has to rebuild as often), and a thing to do preliminary VK match tests before I had the full thing working. I'll move back 👍🏻

PR #924, opened 2026-07-31
Add a match-only Lean fixture exporter for non-accepting captures

ebfull:
Refactor dump_vesta_lean_fixture into a private mode-parameterized core (FixtureMode::{Honest, MatchOnly}) with two public wrappers. The honest exporter is unchanged in behavior and output. The new dump_vesta_lean_fixture_match_only emits fixtures for captures of non-accepting runs (random proof strings): it fails fast if the captured MSM *is* the identity, emits the fingerprint_matches theorem but neither eval theorem, and marks the generated header as a match-only capture. Everything else — coordinate ordering, on-curve validation, URS emission, instance-commitment re-derivation, schedule entries, and the slot-reconstruction and term-count guards — is shared with the honest path, so regeneration stays bit-reproducible across both modes.

Two new integration tests cover the match-only render (same decl list minus the eval theorems) and the inverted fail-fast (an accepting capture is refused).


halo2_proofs/src/plonk/fingerprint/vesta_lean.rs
```diff
@@ -251,6 +255,16 @@ fn render_transcript_capture(
...
 
+/// Which theorem family the exporter emits, and which identity fail-fast it enforces.
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+enum FixtureMode {
+    /// Accepting capture: the MSM must be the identity; the eval theorems are emitted.
+    Honest,
+    /// Non-accepting (random-input) capture: the MSM must *not* be the identity; only the match
+    /// theorem is emitted.
+    MatchOnly,
+}
```
TalDerei, line 276: in favor of parameterizing using an enum like this so these changes remain minimally invasive. agreement rather than acceptance makes sense here, and so we're proving agreement at captured samples and those samples are arbitrary proof strings.

halo2_proofs/src/plonk/fingerprint/vesta_lean.rs
TalDerei: referencing https://github.com/zcash/ironwood/pull/168#discussion_r3699483102, we should add `numInstanceColumns := {}` here (rather than embedding a custom default in Lean). this will also inform some other suggestions below that we should fold into this release.

github's suggestion feature doesn't seem to be working; replace:

```ts
"def shape : Shape := {{ k := {}, numProofs := {}, numAdviceColumns := {}, numLookups := {}, numPermutationSets := {}, numPermutationColumns := {}, numQuotientPieces := {}, numInstanceQueries := {}, numAdviceQueries := {}, numFixedQueries := {}, numPointSets := {} }}\n\n",
```
with

```ts
 "def shape : Shape := {{ k := {}, numProofs := {}, numAdviceColumns := {}, numLookups := {}, numPermutationSets := {}, numPermutationColumns := {}, numQuotientPieces := {}, numInstanceColumns := {}, numInstanceQueries := {}, numAdviceQueries := {}, numFixedQueries := {}, numPointSets := {} }}\n\n",
```

TalDerei: github's suggestion feature doesn't seem to be working; replace:

```ts
k, num_proofs, n_advice, n_lookups, n_perm_sets, n_perm_cols, n_quotient, n_inst_q, n_adv_q, n_fixed_q, n_point_sets,
```

with

```ts
k, num_proofs, n_advice, n_lookups, n_perm_sets, n_perm_cols, n_quotient, n_inst_cols, n_inst_q, n_adv_q, n_fixed_q, n_point_sets,
```

halo2_proofs/tests/fingerprint_export.rs
```diff
@@ -311,3 +313,115 @@ fn rejects_non_identity_capture() {
...
+        K,
+        &[&[&wrong_pubinputs[..]]],
+        &transcript,
+        &msm,
+    );
+    for expected in [
+        "-- Auto-generated by halo2 `dump_vesta_lean_fixture_match_only`. Do not edit by hand.",
+        "-- Match-only capture: the deployed verifier ran to completion on a random proof",
+        "namespace Halo2.Fixture.RenderMatchOnly",
+        "def shape : Shape",
```
TalDerei, line 354: github's suggestion feature doesn't seem to be working; replace:

```ts
"-- Match-only capture: the deployed verifier ran to completion on a random proof",
```

with

```ts
"-- Match-only capture: the deployed verifier ran to completion, but the captured",
```

(required by the header rewording in `vesta_lean.rs` — without it this test fails on a string the exporter no longer emits)

and replace:

```ts
"def shape : Shape",
```

with

```ts
"def shape : Shape",
"numInstanceColumns := 1",
"def capturedNumInstanceColumns : ℕ := shape.numInstanceColumns",
```

halo2_proofs/tests/fingerprint_export.rs
TalDerei, line 206: github's suggestion feature doesn't seem to be working; replace:

```ts
"def shape : Shape",
```

with

```ts
"def shape : Shape",
"numInstanceColumns := 1",
"def capturedNumInstanceColumns : ℕ := shape.numInstanceColumns",
```

halo2_proofs/src/plonk/fingerprint/vesta_lean.rs
TalDerei: github's suggestion feature doesn't seem to be working; replace:

```ts
        out.push_str(&format!(
            "def capturedNumInstanceColumns : ℕ := {}\n\n",
            n_inst_cols
        ));
```

with

```ts
        out.push_str("def capturedNumInstanceColumns : ℕ := shape.numInstanceColumns\n\n");
```

halo2_proofs/src/plonk/fingerprint/vesta_lean.rs
```diff
@@ -1133,24 +1208,38 @@ impl VerifyingKey<EqAffine> {
...
+            ),
+            FixtureMode::MatchOnly => {
+                out.push_str("-- Auto-generated by halo2 `dump_vesta_lean_fixture_match_only`. Do not edit by hand.\n");
+                out.push_str("-- Match-only capture: the deployed verifier ran to completion on a random proof\n");
+                out.push_str(
+                    "-- string and the captured MSM is deliberately NOT the identity, so no eval\n",
+                );
+                out.push_str(
+                    "-- theorem is emitted; this fixture witnesses coefficient agreement only.\n",
+                );
```
TalDerei, line 1273: i'd change this to say what the exporter enforces, and stop asserting what it can't check (ie. it verifies non-acceptance but doesn't know if it's random proof, honest proof with wrong public inputs, or anything else).

```ts
  FixtureMode::MatchOnly => {
      out.push_str("-- Auto-generated by halo2 `dump_vesta_lean_fixture_match_only`. Do not edit by hand.\n");
      out.push_str("-- Match-only capture: the deployed verifier ran to completion, but the captured\n");
      out.push_str("-- MSM is deliberately NOT the identity (a non-accepting run), so no eval theorem\n");
      out.push_str("-- is emitted; this fixture witnesses coefficient agreement only. What made the\n");
      out.push_str("-- run non-accepting is the caller's business and is not attested here.\n");
  }
```

TalDerei (commented):
left some observations on things we should fold into the release, and CI's failing.

TalDerei:
addressed feedback in https://github.com/ebfull/halo2/pull/2.

TalDerei:
afaict this Ci failure isn't being caused by anything on this branch. `ci.yml` seems to already be breaking on some other branches.

halo2_proofs/src/plonk/fingerprint/vesta_lean.rs
TalDerei, line 276: does this need to be non-exhaustive?

TalDerei:
> afaict this Ci failure isn't being caused by anything on this branch. `ci.yml` seems to already be breaking on some other branches.

working on another follow-up to fix build failures.

PR #927, opened 2026-08-07
Speed up IPA generator collapse with batched GLV

ValarDragon:
## Summary

Speeds up the IPA prover's generator-collapse step by batching the
multiplications that share one scalar.

Each IPA round computes the equivalent of

```text
g_lo[i] = g_lo[i] + challenge * g_hi[i]
```

for one challenge shared by the entire half-vector. The prover now delegates
that shared work to
`CurveExt::batch_mul_same_scalar_vartime`, adds the low-half points, and keeps
the existing final batch normalization. Pallas and Vesta use the batched GLV
implementation added by zcash/pasta_curves#97.

This also raises the workspace MSRV from Rust 1.60 to 1.63, as previously
agreed for this dependency update.

Depends on: https://github.com/zcash/pasta_curves/pull/97

## Security and correctness

The optimized multiplication is variable-time in the scalar. This call site
is appropriate because the scalar is a public Fiat-Shamir challenge and the
generator vector is public. No witness, blinding factor, spending key, or
other secret scalar enters this path.

The computation, transcript, proof encoding, and verifier are unchanged.
Focused Pallas and Vesta tests compare the optimized collapse with the prior
formula across identity points and scalar edge cases. Every benchmark backend
also creates and verifies a proof before measurement.

## Single-thread performance

These per-proof benchmarks deliberately use one worker so deployments can
parallelize independent proofs at a higher level. Results are means from an
ABBA run order on an otherwise idle arm64 machine.

| Workload | Native | Batched GLV | Latency reduction | Throughput |
| --- | ---: | ---: | ---: | ---: |
| Orchard proof, one action, protocol k = 11 | 3.925 s | 3.779 s | 3.73% | 1.039x |
| Synthetic Halo proof, k = 13 | 2.554 s | 1.804 s | 29.36% | 1.416x |
| k = 13 generator collapse only | 1.144 s | 395.4 ms | 65.44% | 2.893x |

The Orchard and synthetic rows are different circuits, so their absolute
times are not directly comparable. In the k = 13 run, generator collapse
accounts for 748.6 ms of the 749.7 ms end-to-end saving.

One k = 11 GLV run contained a severe high outlier outside generator collapse.
The ABBA median aggregate is 3.925 s to 3.769 s (3.98%, 1.042x), while the
directly affected collapse phase is stable at 240.7 ms to 83.9 ms (65.15%).

## Verifier impact

There is no direct verifier speedup in this PR. The new call is confined to
the IPA prover. Verification uses heterogeneous-scalar MSMs rather than a
vector multiplied by one shared scalar, so this particular batching hook does
not apply to either single or batch verification.

## API-surface audit

- No public Rust item, signature, or visibility changes are made in this
  repository.
- The `rust-version` of `halo2`, `halo2_proofs`, `halo2_gadgets`, and
  `halo2_poseidon` changes from 1.60 to 1.63.
- `halo2_proofs` enables the `pasta_curves/glv` dependency feature. Because
  `pasta_curves` is already publicly re-exported as `halo2_proofs::pasta`, this
  also makes `halo2_proofs::pasta::glv` reachable downstream.
- The minimum `pasta_curves` requirement changes from `0.5` to `0.5.2` in the
  draft. It must be updated to the release containing zcash/pasta_curves#97
  before merge.
- No `pub(crate)` item, Cargo feature, enum variant, type alias, constant, or
  constructor is added or changed.

## Validation

- Rust 1.63 workspace all-features check
- Rust 1.63 formatting check
- `halo2_proofs` all-features release tests and doctests
- focused Pallas and Vesta generator-collapse equivalence tests
- default-feature Clippy with warnings denied
- proof-generation preflights verified for the k = 11 and k = 13 benchmark
  workloads

## Stacked dependency note

This draft temporarily pins commit
`21898a77810260ae8f2ed4f27058315729e683bc` from the `valargroup/pasta_curves`
branch through `[patch.crates-io]`, so CI and reviewers can exercise the stack
before the prerequisite is published. This override must be removed and the
normal crates.io dependency updated after zcash/pasta_curves#97 is released;
the crate must not be released with the git patch.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -109,12 +109,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "349f9b6a179ed607305526ca489b34ad0a41aed5f7980fa90eb03160b69598fb"
 
 [[package]]
-name = "bit_field"
-version = "0.10.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "dc827186963e592360843fb5ba4b973e145841266c1357f7180c43526f2e5b61"
-
-[[package]]
 name = "bitflags"
 version = "1.3.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -591,22 +585,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "exr"
-version = "1.7.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d1e481eb11a482815d3e9d618db8c42a93207134662873809335a92327440c18"
-dependencies = [
- "bit_field",
- "flume",
- "half 2.2.1",
- "lebe",
- "miniz_oxide",
- "rayon-core",
- "smallvec",
- "zune-inflate",
-]
-
-[[package]]
 name = "fastrand"
 version = "1.9.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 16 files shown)


ValarDragon:
Superseded by [zakura-core/libraries#5](https://github.com/zakura-core/libraries/pull/5), main-line stack 2 of 6.

The replacement preserves the public-challenge security argument, generator-
collapse equivalence tests, benchmark context, and dependency on the migrated
same-scalar hook. It inherits the replacement repository's Rust 1.88 MSRV
instead of carrying the original upstream Rust 1.63 update. Closing this
upstream draft so the six-PR main line is reviewed and maintained together.

`valargroup/halo2@d49af5b44fb8825568bae3c6f8815cb4716efa9c`.

PR #928, opened 2026-08-08
Optimize small constants and subtraction in the polynomial evaluator

ValarDragon:
## Summary

This removes avoidable field work in the internal polynomial evaluator:

- Scales and direct constant products by `1`, `-1`, and `2` now use identity,
  negation, and doubling respectively instead of general field
  multiplication. Direct products are recognized with the constant on either
  side.
- `Ast::Add(lhs, Ast::Scale(rhs, -1))`, the representation produced by the
  subtraction operators, now subtracts `rhs` directly instead of first
  materializing and negating a complete intermediate vector.

The three changes are separate commits. They preserve the existing AST shapes
and evaluation interfaces; in particular, direct constant products are not
rewritten during AST construction. Values `3` and above retain the generic
multiplication path.

## Performance

In the single-threaded genuine one-action Orchard proving benchmark, avoiding
identity-scale multiplications reduced proving time by about 15.6%. Direct
subtraction removed a further 3.0% on the then-current stacked branch. The
combined effect is approximately 18% less proving time for this benchmark.

Against the two-commit PR head, the small-constant change reduced a stable
single-threaded bracket from 1,082.5 ms to 1,077.1 ms, or 0.50%. At Orchard's
16,384-point extended domain it avoids or replaces 901,120 general field
multiplications per proof: 262,144 identity products are eliminated and
638,976 multiplications by two become field doublings.

## Validation

- `cargo clippy -p halo2_proofs --lib -- -D warnings`
- Focused tests exercise both polynomial bases, constants on either product
  side, the `-1`/`1`/`2` paths, an arbitrary generic constant, and the
  subtraction AST path.

## API and dependencies

- No public or `pub(crate)` API changes.
- No function-signature or visibility changes.
- No feature or dependency changes.

### DIFF
--- a/halo2_proofs/CHANGELOG.md
+++ b/halo2_proofs/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to Rust's notion of
 
 ## [Unreleased]
 
+### Changed
+- The internal polynomial evaluator now avoids field multiplication for scales
+  by $1$ and $-1$, and evaluates subtraction without materializing a negated
+  polynomial.
+
 ## [0.3.5] - 2026-08-02
 ### Added
 - `halo2_proofs::plonk::VerifyingKey::dump_vesta_lean_fixture_match_only`
--- a/halo2_proofs/CHANGELOG.md
+++ b/halo2_proofs/CHANGELOG.md
@@ -8,9 +8,9 @@ and this project adheres to Rust's notion of
 ## [Unreleased]
 
 ### Changed
-- The internal polynomial evaluator now avoids field multiplication for scales
-  by $1$ and $-1$, and evaluates subtraction without materializing a negated
-  polynomial.
+- The internal polynomial evaluator now uses field identity, negation, or
+  doubling for multiplications by $1$, $-1$, or $2$, and evaluates subtraction
+  without materializing a negated polynomial.
 
 ## [0.3.5] - 2026-08-02
 ### Added
--- a/halo2_proofs/src/poly/evaluator.rs
+++ b/halo2_proofs/src/poly/evaluator.rs
@@ -162,6 +162,16 @@ impl<E, F: Field, B: Basis> Evaluator<E, F, B> {
                 ),
                 Ast::Add(a, b) => {
                     let mut lhs = recurse(a, ctx);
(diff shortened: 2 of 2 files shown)


halo2_proofs/CHANGELOG.md
```diff
@@ -7,6 +7,11 @@ and this project adheres to Rust's notion of
 
 ## [Unreleased]
 
+### Changed
+- The internal polynomial evaluator now avoids field multiplication for scales
+  by $1$ and $-1$, and evaluates subtraction without materializing a negated
+  polynomial.
```
ebfull: This doesn't affect the public API so it doesn't need to be written into the changelog.

halo2_proofs/src/poly/evaluator.rs
```diff
@@ -178,8 +188,14 @@ impl<E, F: Field, B: Basis> Evaluator<E, F, B> {
                 }
                 Ast::Scale(a, scalar) => {
                     let mut lhs = recurse(a, ctx);
-                    for lhs in lhs.iter_mut() {
-                        *lhs *= scalar;
+                    if *scalar == -F::ONE {
```
ebfull: I think we have a `NEG_ONE` constant _somewhere_, but I don't think it's accessible here. That's annoying because this code will probably negate this constant every time.

ebfull (dismissed):

ValarDragon:
Superseded by three independently reviewable replacements:

- [±1 cases, AST stack 1 of 3](https://github.com/zakura-core/libraries/pull/17)
- [Horner/subtraction, AST stack 2 of 3](https://github.com/zakura-core/libraries/pull/18)
- [doubling for constant 2, AST stack 3 of 3](https://github.com/zakura-core/libraries/pull/19)

Together they preserve the original three source commits, focused tests, and
Orchard measurements while separating the 15.6%, 3.0%, and 0.50% layers. The
third replacement intentionally retains the original commit's direct-
constant recognition for `±1` and `2`. The upstream feedback about the
internal-only changelog entry and repeated `-F::ONE` construction is carried
into the replacement review context. Closing this PR so maintenance continues
in a dedicated three-PR AST stack based directly on `main`.

`valargroup/halo2@934a008c4eade721764eecdeb7f73d44051f9b74`.

PR #929, opened 2026-08-08
Reduce verifier instance-commitment overhead

ValarDragon:
## Summary

This contains three independently reviewable commits:

1. Commit directly to the supplied Lagrange-prefix instance values instead of
   allocating and processing a domain-sized zero-padded vector.
2. Convert MSM scalars to canonical representations once per MSM instead of
   once per Pippenger window.
3. Batch-normalize instance commitments when a verification call contains at
   least four commitments; smaller calls retain individual normalization.

It also adds an Orchard-shaped Criterion benchmark at $k = 11$ with nine public
inputs per action and action counts 1, 2, 16, and 64.

## Performance

The following external integration measurements use `zcash/orchard` at
`2b531b9d`, an Apple M4 Max, Rust 1.85.1, and one Rayon thread. Proof creation
is outside the timed region. Each batch is drawn from 64 separately encoded
genuine one-action Orchard proofs with independently randomized proof
commitments. The values are medians of 15 measured `BatchVerifier::finalize`
runs after three warm-up runs:

| Proofs | Upstream main | Three-commit stack | Time reduction |
| ---: | ---: | ---: | ---: |
| 1 | 18.558 ms | 16.907 ms | 8.9% |
| 2 | 22.306 ms | 19.392 ms | 13.1% |
| 16 | 71.653 ms | 53.925 ms | 24.7% |
| 64 | 240.178 ms | 171.766 ms | 28.5% |

Only the first two commits affect this separate-proof workload. Each
`BatchVerifier` entry invokes `verify_proof` for one one-action proof, so it
contains one instance commitment and does not reach the four-commitment batch
normalization threshold.

The committed Criterion benchmark measures 1, 2, 16, and 64 action instances
inside one encoded proof. That workload exercises the third commit, but uses a
minimal Orchard-shaped circuit rather than the full Orchard circuit.

Applications can parallelize independent verification requests at a higher
layer.

## Validation

- `cargo test -p halo2_proofs --test plonk_api`
- `cargo check -p halo2_proofs --bench plonk_verifier`
- `cargo clippy -p halo2_proofs --all-targets --all-features -- -D warnings`
- Focused tests compare prefix and zero-padded commitments and exercise the
  batch-normalization threshold with a valid proof.

## API and dependencies

- No public or `pub(crate)` Rust API changes.
- No public function-signature or visibility changes.
- No dependency or feature changes.
- Adds the `plonk_verifier` benchmark target and its source file; Criterion was
  already a development dependency.

### DIFF
--- a/halo2_proofs/CHANGELOG.md
+++ b/halo2_proofs/CHANGELOG.md
@@ -11,6 +11,8 @@ and this project adheres to Rust's notion of
 - `halo2_proofs::plonk::verify_proof` now computes instance commitments
   directly from their supplied Lagrange coefficients, without zero-padding to
   the evaluation domain.
+- `halo2_proofs::plonk::verify_proof` now batch-normalizes instance commitments
+  when verifying at least four commitments.
 - `halo2_proofs::arithmetic::best_multiexp` now computes each scalar's
   canonical representation once per MSM instead of once per window.
 
--- a/halo2_proofs/CHANGELOG.md
+++ b/halo2_proofs/CHANGELOG.md
@@ -7,6 +7,11 @@ and this project adheres to Rust's notion of
 
 ## [Unreleased]
 
+### Changed
+- `halo2_proofs::plonk::verify_proof` now computes instance commitments
+  directly from their supplied Lagrange coefficients, without zero-padding to
+  the evaluation domain.
+
 ## [0.3.5] - 2026-08-02
 ### Added
 - `halo2_proofs::plonk::VerifyingKey::dump_vesta_lean_fixture_match_only`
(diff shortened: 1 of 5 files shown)


ValarDragon:
Superseded by [zakura-core/libraries#14](https://github.com/zakura-core/libraries/pull/14), main-line stack 4 of 6.

The replacement carries all three verifier commits, the Orchard-shaped
benchmark, focused correctness tests, and the single-thread batch-size
measurements into the maintained stack. Closing this upstream draft so review
and performance tracking continue there.

`valargroup/halo2@0f7a808231e487baa724517c3242bd79b276ffa8`.

PR #930, opened 2026-08-08
Avoid zero-padding prover instance commitments

ValarDragon:
## Summary

Public instance columns are prefix-populated and zero-padded to the evaluation
domain. The prover previously passed the entire padded polynomial to the
instance commitment MSM. This instead commits to the supplied prefix plus the
blinding generator; the omitted trailing coefficients are all zero.

The prover still constructs the same zero-padded polynomial for interpolation,
coset evaluation, and opening, so this changes only the amount of work used to
compute the commitment.

## Performance

On the single-threaded genuine one-action Orchard proving benchmark this saved
roughly 8 ms, or about 0.5%, on the then-current optimization stack.

## Validation

- `cargo clippy -p halo2_proofs --all-targets --all-features -- -D warnings`
- A focused equivalence test compares prefix commitments with the original
  zero-padded commitment for both an empty column and a column containing an
  internal zero.

## API and dependencies

- No public or `pub(crate)` API changes.
- No function-signature or visibility changes.
- No feature or dependency changes.

## Integration note

The [verifier optimization PR](https://github.com/zcash/halo2/pull/929) has the
same module-private helper. The second of these PRs to land should move that
helper to their common private parent module during rebase, avoiding permanent
duplication without making either PR depend on the other today.

### DIFF
--- a/halo2_proofs/CHANGELOG.md
+++ b/halo2_proofs/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to Rust's notion of
 
 ## [Unreleased]
 
+### Changed
+- `halo2_proofs::plonk::create_proof` now commits directly to the supplied
+  instance values instead of zero-padding them to the evaluation domain first.
+
 ## [0.3.5] - 2026-08-02
 ### Added
 - `halo2_proofs::plonk::VerifyingKey::dump_vesta_lean_fixture_match_only`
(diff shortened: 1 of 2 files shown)


ValarDragon:
Superseded by [zakura-core/libraries#15](https://github.com/zakura-core/libraries/pull/15), main-line stack 5 of 6.

The replacement preserves the sparse-prefix commitment optimization,
zero-padding equivalence test, and Orchard measurement. It shares the private
commitment helper with the preceding verifier layer. Closing this upstream
draft so the prover and verifier forms remain maintained together.

`valargroup/halo2@7b223457b6660f2999e86685268620dccccdb5ee`.

PR #931, opened 2026-08-08
Skip zero-only prefix stages in coefficient-to-extended FFTs

ValarDragon:
## Summary

When an `n`-coefficient polynomial is transformed over an extended domain of
size `N`, the remaining `N - n` inputs are structural zeros. After bit
reversal, the first `log2(N / n)` radix-2 stages only replicate each live
coefficient.

This change skips those zero-only stages, directly materializes the first
retained stage, and then resumes the existing butterfly schedule. It reduces
butterfly work from `N / 2 * log2(N)` to `N / 2 * log2(n)`. Twiddle generation
and materializing the `N` outputs remain `O(N)`.

This changes the coefficient-to-extended transform used during proving and
key generation. Inverse transforms and verification are unaffected.

## Orchard impact

Orchard transforms 2,048 coefficients over a 16,384-point extended domain, so
three of fourteen butterfly stages are zero-only. There are 23 such transforms
per proof. The change removes, per proof:

- 235,520 non-unity field multiplications;
- 1,130,496 field additions or subtractions.

A matched single-threaded real one-Action Orchard `k = 11` A/B/A benchmark on
an Apple M4 Max measured:

| Run | Time |
| --- | ---: |
| optimized | 897.86 ms |
| control | 911.13 ms |
| optimized | 896.07 ms |

The adjacent comparisons are 1.46% and 1.65% faster. With three Rayon workers,
a stable pair measured 612.73 ms optimized versus 614.14 ms control; the 0.23%
difference was within the benchmark's noise threshold. Benchmarks used Rust
1.85.1 and kept key generation and proof verification outside the timed loop.

## Memory and API

There is no persistent cache and no evaluation-domain layout or public API
change. During the transform, the implementation temporarily retains the
original `n` coefficients, `N` outputs, `N / 2` twiddles, and `N / n` scratch
field elements. For Orchard this adds approximately 64.3 KiB of transient
storage relative to the existing transform.

## Tests

- Exact differential coverage against `best_fft` for `k = 0..=8` and extension
  factors 1 through 16, including extension factors larger than `n`.
- Full `halo2_proofs` library tests on Rust 1.63 with three Rayon workers:
  33 passed.
- Focused differential test on Rust 1.63 with one Rayon worker.
- Focused differential test on Rust 1.63 with `--no-default-features`.
- `cargo +1.63.0 fmt --all -- --check` and `git diff --check`.

### DIFF
--- a/halo2_proofs/src/poly/domain.rs
+++ b/halo2_proofs/src/poly/domain.rs
@@ -3,6 +3,7 @@
 
 use crate::{
     arithmetic::{best_fft, parallelize},
+    multicore,
     plonk::Assigned,
 };
 
@@ -245,8 +246,7 @@ impl<F: WithSmallOrderMulGroup<3>> EvaluationDomain<F> {
         assert_eq!(a.values.len(), 1 << self.k);
 
         self.distribute_powers_zeta(&mut a.values, true);
-        a.values.resize(self.extended_len(), F::ZERO);
-        best_fft(&mut a.values, self.extended_omega, self.extended_k);
+        Self::fft_zero_padded(&mut a.values, self.extended_omega, self.k, self.extended_k);
 
         Polynomial {
             values: a.values,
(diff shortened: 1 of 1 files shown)


ValarDragon:
Superseded by [zakura-core/libraries#16](https://github.com/zakura-core/libraries/pull/16), main-line stack 6 of 6.

The replacement preserves the zero-padding derivation, differential tests,
memory/API notes, and single-thread Orchard measurements. The original
Codecov caveat (81.72% patch coverage, 17 changed lines uncovered) is recorded
for replacement review. Closing this upstream draft so the final layer is
maintained with the six-PR main line.

`valargroup/halo2@e28afc659af008c16ddd633b53fee8b1c643dbac`.

PR #933, opened 2026-08-26
Carry the consumed proof bytes in Lean fixture exports

TalDerei:
Related to https://github.com/zcash/orchard/pull/544 and consumed in https://github.com/zcash/ironwood/pull/215.

New `dump_vesta_lean_fixture_honest_with_proof_bytes` and `dump_vesta_lean_fixture_match_only_with_proof_bytes` fixture exports behind `unstable-verifier-fingerprint` that includes exact bytes consumed (proof commitments, polynomial evaluations, the multi-opening proof, and the IPA proof) by the verifier and reject mismatches. It reserializes the points/scalars using existing halo2 encoding, checks they match the original proof bytes, then exports those bytes as hex.


halo2_proofs/src/plonk/fingerprint/vesta_lean.rs
```diff
@@ -792,6 +902,41 @@ impl VerifyingKey<EqAffine> {
...
+        out.push_str("/-- The exact text `VerifyingKey::from_parts` hashed into `capturedVkTranscriptRepr`:\n");
+        out.push_str("the compact `Debug` rendering of the pinned key (`Halo2-Verify-Key` BLAKE2b over its\n");
+        out.push_str(
+            "little-endian `u64` byte length then its bytes, reduced modulo `p`). The exporter\n",
+        );
+        out.push_str("re-hashed it and checked the scalar above before emitting. -/\n");
+        out.push_str(&format!(
+            "def capturedPinnedKeyDescription : String :=\n  {:?}\n\n",
+            pinned
+        ));
```
TalDerei, line 939: `capturedPinnedKeyDescription` is the debug string representing `circuit_description_post_nu6_3` (a succinct representation of the raw 28K LOC dump). `Fixture.lean` now contains both:
- `capturedPinnedKeyDescription`: the exact hash preimage string
- `capturedVkTranscriptRepr`: the field element digest that's hashed into the transcript (which we can now derive in Lean instead of trusting the fixture)

ebfull (approved):

daira (approved):
utACK

PR #935, opened 2026-08-31
halo2 evaluation micro-opts

dannywillems:
See commits.

### DIFF
--- a/halo2_proofs/src/poly/domain.rs
+++ b/halo2_proofs/src/poly/domain.rs
@@ -3,6 +3,7 @@
 
 use crate::{
     arithmetic::{best_fft, parallelize},
+    multicore,
     plonk::Assigned,
 };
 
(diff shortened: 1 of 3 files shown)


halo2_proofs/src/poly/domain.rs
```diff
@@ -382,6 +382,84 @@ impl<F: WithSmallOrderMulGroup<3>> EvaluationDomain<F> {
         });
     }
 
+    fn fft_zero_padded(coefficients: &mut Vec<F>, omega: F, log_n: u32, extended_log_n: u32) {
```
dannywillems, line 385: Reviewer note: this should not be in this structure, as it doesn't depend on it. It should be extracted in a utils.

halo2_proofs/src/poly/domain.rs
```diff
@@ -497,6 +653,34 @@ pub struct PinnedEvaluationDomain<'a, F: Field> {
     omega: &'a F,
 }
 
+#[test]
+fn test_zero_padded_fft_matches_best_fft() {
+    use crate::{arithmetic::best_fft, pasta::pallas::Scalar};
```
dannywillems, line 658: This `use` statement should be moved up.

PR #936, opened 2026-09-02
Use MSRV 1.88

dannywillems:


halo2_gadgets/src/ecc/chip/mul.rs
```diff
@@ -421,6 +421,7 @@ impl<F: Field> Deref for Z<F> {
 // https://p.z.cash/halo2-0.1:ecc-var-mul-witness-scalar?partial
 #[allow(clippy::assign_op_pattern)]
 #[allow(clippy::ptr_offset_with_cast)]
+#[allow(clippy::manual_div_ceil)]
```
dannywillems, line 424: This should be removed and the lint should be addressed.

dannywillems: Removing it seems to be giving:

```
= note: this warning originates in the macro `$crate::construct_uint`
        which comes from the expansion of the macro `construct_uint`
```

Leaving it like this for now.

dannywillems (commented):
Self utACK 3bafd6bdacdfe61d452ba35e47d78d362858c7c7

dannywillems (commented):
self-utACK 3d57b0f48123e590108b1d501157543e18d78ff9

nuttycom (approved):
utACK

PR #943, opened 2026-09-15
Update ff, group, rand and other packages

dannywillems:


halo2/CHANGELOG.md
```diff
@@ -8,6 +8,8 @@ and this project adheres to Rust's notion of
 ## [Unreleased]
 ### Changed
 - The minimum supported Rust version is now 1.88.
+- The re-exported `halo2_proofs` API migrated to `ff 0.14`, `group 0.14` and
```
str4d: This is completely false. There is nothing re-exported from the `halo2` crate; it is completely empty.

halo2_proofs/CHANGELOG.md
```diff
@@ -25,6 +25,20 @@ and this project adheres to Rust's notion of
...
+  internal committing APIs it drives) is now bound by `rand_core::Rng` instead
+  of `rand_core::RngCore`, which `rand_core 0.10` deprecates in favour of `Rng`.
+  `Rng` is the infallible half of the new fallible RNG interface, so a fallible
+  source such as `getrandom::SysRng` must be wrapped in `rand_core::UnwrapErr`
+  (which panics on failure, matching the old `rand_core::OsRng` behaviour).
+- `halo2_proofs::plonk::BatchVerifier::finalize` now draws its internal
+  randomness from `UnwrapErr(rand::rngs::SysRng)` rather than
+  `rand_core::OsRng`, which no longer exists. `SysRng` is likewise stateless, so
+  the property this relies on (an RNG that does not clone its internal state
+  when shared between threads) is preserved.
```
str4d: This is a no-op (SysRng is just renamed OsRng), and so shouldn't be in the changelog.

halo2_proofs/Cargo.toml
```diff
@@ -49,11 +49,12 @@ harness = false
...
 [dependencies]
 backtrace = { version = "0.3", optional = true }
-ff = "0.13"
-group = "0.13"
+ff = "0.14"
+group = "0.14"
 indexmap = "1"
 pasta_curves = "0.5"
-rand_core = { version = "0.6", default-features = false }
+rand = { version = "0.10", default-features = false, features = ["sys_rng"], optional = true }
```
str4d: Hmm, I'd prefer we just depend on `rand_core` and `getrandom` instead of `rand` if possible. I see that we do depend on `rand` in `pasta_curves` though, so I won't block on this unless I can determine that we can also just depend on `rand_core` in `pasta_curves`.

halo2_proofs/Cargo.toml
```diff
@@ -70,15 +71,10 @@ assert_matches = "1.5"
...
-
-# Indirect dev-dependencies that we pin to preserve MSRV in CI checks.
-dashmap = ">=5, <5.5.0" # dashmap 5.5.0 has MSRV 1.64
-image = ">=0.24, <0.24.5" # image 0.24.5 has MSRV 1.61
-tempfile = ">=3, <3.7.0" # tempfile 3.7.0 has MSRV 1.63
+rand = "0.10"
 
 [target.'cfg(all(target_arch = "wasm32", target_os = "unknown"))'.dev-dependencies]
-getrandom = { version = "0.2", features = ["js"] }
+getrandom = { version = "0.4", features = ["wasm_js"] }
```
str4d: CI is failing on this target, saying that this feature needs to be enabled. So something isn't working with the cfg gate here.

str4d (changes requested):
Reviewed 3716502f0752dbb0921e7f31edb1865e2e3e83de.

str4d:
[Force-pushed](https://github.com/zcash/halo2/compare/3716502f0752dbb0921e7f31edb1865e2e3e83de..5e2fb5723d2843d224a6ce075952be73bc25fd54) to address my review comments.

str4d (approved):
ACK 5e2fb5723d2843d224a6ce075952be73bc25fd54. I confirmed that the pinned commits exist in the upstream main branches; we can bump to the final versions once they are published.

Issue #578, opened 2022-05-09
Extend `EccChip` to support Vesta


str4d:
`halo2_gadgets::ecc` provides general-purpose ECC gadgets that are suitable for any curve. However, `EccChip` (the only current chip implementation of `EccInstructions`) is specialised to the Pallas curve. We will need to extend it to support Vesta in order to use it for recursion.

therealyingtong:
Related to #454

PR #945, opened 2026-09-20
book: fix slips on the gadget design pages and define the Pasta primes

daira:
Fixes on the gadget design pages and the background pages, found while checking them against `halo2_gadgets` and against a formalization of the gadgets. The constraints in the code are not changed. Degrees are counted as `Expression::degree` does: a selector and a fixed-column query each count one.

## Where `p` and `q` are defined

The gadget pages use `p` and `q` for the Pallas and Vesta base field moduli without defining them. The Fields background page gains a "The Pasta primes" section with the values and the cycle, and the Curves page links to it. The Curves page's discussion of cycles now starts with Halo 2 over Vesta, matching Orchard's usage; its hashing-to-curves section cites RFC 9380 and Simplified SWU and links to the Ironwood book's Group-Hash Indifferentiability page; and the implementation Fields page's link to the Pasta blog post is repaired.

## ECC addition

- The argument for the `(0, 0)` encoding named the field as `F_q`; the curve is Pallas, so its base field is `F_p`. (The claim holds in both Pasta base fields: 5 is a non-square and −5 a non-cube modulo both `p` and `q`.)
- In the analysis of constraint 3, the disjunct `x_q = 0` was written `x_p = 0` a second time.
- Two case justifications cited the wrong hypothesis: (4) in the `(0, 0) + (x, y)` case, and (5) in the doubling case.
- The incomplete-addition note said only that `P ⸭ (-P)` is unsatisfiable. It now also says that for `P ⸭ P` both constraints reduce to `0 = 0` and leave the output unconstrained, and that the cases involving `𝒪` are not implemented, which is why it is neither sound nor complete for arbitrary inputs.

## ECC chip page

The non-square and non-cube claims are about `F_p`, not `F_q`. The stale claim that Vesta support is coming soon is replaced by a pointer to #578.

## Decomposition

In the short range check the third cell holds `2^{-n}`, not `2^{K-n}`, and the bitshift gate is `q_bitshift · (α · 2^K · 2^{-n} − α')`, of degree 3 rather than 2. The running-sum index range and a typo are fixed, and the page is wrapped.

## Fixed-base scalar multiplication

The final complete addition adds the last window to the accumulated sum, not to `M[83][k_83]`. The interpolation constraint has degree 9, not 8. The window table is stated for general `W`, so it no longer hard-codes 83 and 84. A set brace, a column index in the layout table, `t_P` for `t_p`, and two wording slips are fixed.

## Variable-base scalar multiplication

The gradient-2 constraint has degree 4, not 3, in both incomplete gates. The complete-round and LSB gate tables get their degrees. `2[T]_y` meant a coordinate of `[2]T`, the loop index runs from 254 down to 4, and the overflow-check link is now relative.

## Sinsemilla

The message is `kn` bits, the running sum starts at `m` rather than an undefined `α`, and the closed form for the running sum summed the wrong words. The `y` constraint has degree 6, not 5, and the gated lookup argument has degree 7, not 6. The note on exceptional cases now says what the security argument shows and that the specification defines the result to be `⊥`.

## MerkleCRH

`b_{0..=10}` denoted `b_0`.

🤖 Claude Fable 5.1


daira (commented):
Self-ACK

